PR #1, opened 2026-04-22
feat: add Docker support with Dockerfile and .dockerignore

dalessandro07:
Amazing project.
Just adding Dockerfile to facilitate deployment on Coolify and other VPSs.


standin000:
want this Docker feature, will it be in the future?

PR #2, opened 2026-04-22
refactor: optimize Dockerfile for multi-stage builds and production readiness

dalessandro07:
Amazing project.
Just adding Dockerfile to facilitate deployment on Coolify and other VPSs.

update.

Dockerfile now tested and working on Coolify.


tashfeenahmed:
Thanks for this! The Dockerfile + .dockerignore parts look useful, but the PR also adds a full login/session layer which conflicts with the project's stated single-user local-first design (see README). Could you open a narrower PR with just the Dockerfile, .dockerignore, and the related .env.example / README.md deployment notes? I'd like to keep the auth discussion separate.

PR #10, opened 2026-04-26
feat: Add DeepSeek API support

minhduckd5:
Adds support for DeepSeek API. Includes DeepSeek V3 and R1 via OpenAI compat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 # FreeLLMAPI
 
-**One OpenAI-compatible endpoint. Fourteen free LLM providers. ~1.3B+ tokens per month.**
+**One OpenAI-compatible endpoint. Fifteen free LLM providers. ~1.3B+ tokens per month.**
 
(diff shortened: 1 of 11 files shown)


tashfeenahmed:
Thanks for the PR @minhduckd5 — appreciate you taking the time.

Closing this with a polite no: DeepSeek's API requires payment (the platform credits aren't fungible to a no-card free tier, and the catalog standard is \"genuinely free, no payment method on file\"). DeepSeek models *are* reachable through the catalog via OpenRouter routes (e.g. `openrouter/deepseek/deepseek-chat-v3:free` and similar) which keeps the same usability without breaking the no-card rule.

If a true permanent-free DeepSeek-direct endpoint ever opens up, very happy to revisit. The PR also has merge conflicts now after a number of intervening changes, so a fresh PR would be cleaner if/when that day comes.

Thanks again for contributing.

PR #33, opened 2026-05-07
feat(router): exhaust all keys for a model before falling back

deadc:
This PR optimizes the routing logic to ensure that the system makes full use of all available quotas for high-priority models before falling back to lower-tier models.

### Improvements
- **Model-First Prioritization:** The router now attempts every healthy key associated with a model before giving up and moving to the next model in the fallback chain.
- **Improved Resilience:** Users with multiple keys for the same platform (e.g., Google AI Studio) will now see their most intelligent models utilized more consistently, even if individual keys hit their specific rate limits.
- **Minor Optimization:** Moved model-level limit calculations outside the key iteration loop.

This ensures that if you have multiple keys with different quotas, the system will prioritize the better model as long as at least one key has remaining quota.


tashfeenahmed:
Superseded by #42 (merged as 7c... — see latest main). Couldn't rebase this branch directly: it included your already-merged Gemini commits from #32 plus a revert of them, so a rebase onto current main produced conflicts on `google.test.ts`. Cherry-picked the two routing commits onto a fresh branch, plus one fix-up for a bug in the new test (the second test made an unguarded `routeRequest(100)` call before adjusting its mock, which threw "All models exhausted" and crashed). Authorship preserved on both routing commits. Thanks for the PR @deadc.

Issue #35, opened 2026-05-11
Security: Multiple vulnerabilities in API authentication, input validation, and error handling


jtbrennan-git:
## Summary

This issue documents several security vulnerabilities found during a code review. I've prepared a fix branch (`security/hardening`) that addresses all of them. Details below.

---

## 1. Authentication Bypass via IP Spoofing (HIGH)

**File:** `server/src/routes/proxy.ts:175`

The `/v1/chat/completions` endpoint skips API key validation for requests from `127.0.0.1` / `::1`:

```js
const isLocal = req.ip === '127.0.0.1' || req.ip === '::1' || req.ip === '::ffff:127.0.0.1';
if (!isLocal) { /* check key */ }
```

`req.ip` is derived from `X-Forwarded-For` when behind a reverse proxy (nginx, which the README recommends). Any client can set `X-Forwarded-For: 127.0.0.1` and bypass auth entirely.

**Fix:** Remove the localhost bypass. Require the unified API key on every request. If local-trust is needed, configure Express's `trust proxy` setting explicitly and document it.

## 2. No Authentication on Admin API Routes (HIGH)

**File:** `server/src/app.ts:31-36`

All `/api/*` routes (keys CRUD, settings, fallback chain, analytics, health) have no authentication:

- `GET /api/keys` — lists all provider key metadata
- `GET /settings/api-key` — returns the unified API key in plaintext
- `PUT /api/fallback` — reorders the entire routing chain
- `POST /api/keys` — adds new provider keys

Anyone who can reach the server can reconfigure everything.

**Fix:** Apply auth middleware to all `/api/*` routes.

## 3. SQL Injection via String Interpolation (MEDIUM)

**File:** `server/src/routes/analytics.ts:30,69,101,132,168`

The `getTimeFilter()` function returns raw SQL fragments interpolated via template literals:

```js
WHERE created_at >= ${since}
```

The `range` query parameter flows directly into SQL. While the current implementation only returns hardcoded strings for known values, the pattern is fragile and the `default` case is an unvalidated fallback.

**Fix:** Compute the cutoff timestamp in JavaScript and pass it as a parameterized query (`?` placeholder).

## 4. Open CORS (MEDIUM)

**File:** `server/src/app.ts:27`

`cors()` with no options defaults to `Access-Control-Allow-Origin: *`. Any browser-based JavaScript on any origin can make requests to all API endpoints.

[...]

PR #36, opened 2026-05-11
Security hardening: auth on all routes, SQL injection fixes, rate limiting, CORS, error sanitization

jtbrennan-git:
## Summary

Fixes multiple security vulnerabilities documented in #35:

1. **Auth bypass (HIGH)** — Removed localhost IP trust on `/v1/chat/completions`. All requests now require the unified API key.
2. **Unauthenticated admin routes (HIGH)** — All `/api/*` endpoints now require authentication.
3. **SQL injection (MEDIUM)** — Replaced template-literal SQL interpolation in `analytics.ts` with parameterized queries.
4. **Open CORS (MEDIUM)** — Changed from wildcard to same-origin.
5. **Timing side-channel (LOW)** — API key comparison now uses `crypto.timingSafeEqual()`.
6. **No rate limiting (MEDIUM)** — Added per-IP sliding window (120 req/min).
7. **Error message leakage (LOW)** — Sanitized provider errors returned to clients.

## Changes

- New `server/src/middleware/auth.ts` — unified API key validation + rate limiting
- `server/src/app.ts` — auth middleware on all routes, restrictive CORS
- `server/src/routes/proxy.ts` — removed localhost bypass, sanitized error messages
- `server/src/routes/analytics.ts` — parameterized queries instead of SQL interpolation
- `server/src/routes/fallback.ts` — map-based sort preset whitelist
- New `server/src/__tests__/helpers.ts` — shared test helper with auth headers
- Updated all test files to include authentication

## Testing

- All 90 existing tests pass
- Build succeeds (`npm run build`)
- Manual verification confirms:
  - No auth → 401
  - Wrong key → 401
  - Correct key → routes work
  - Admin endpoints without key → 401
  - Rate limiting activates at 120 req/min

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1216,448 +1216,6 @@
         "get-tsconfig": "^4.7.0"
       }
     },
-    "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.28.0",
-      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.28.0.tgz",
-      "integrity": "sha512-lhRUCeuOyJQURhTxl4WkpFTjIsbDayJHih5kZC1giwE+MhIzAb7mEsQMqMf18rHLsrb5qI1tafG20mLxEWcWlA==",
-      "cpu": [
-        "ppc64"
-      ],
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "aix"
-      ],
-      "peer": true,
-      "engines": {
-        "node": ">=18"
-      }
-    },
-    "node_modules/@esbuild/android-arm": {
-      "version": "0.28.0",
-      "resolved": "https://registry.npmjs.org/@esbuild/android-arm/-/android-arm-0.28.0.tgz",
-      "integrity": "sha512-wqh0ByljabXLKHeWXYLqoJ5jKC4XBaw6Hk08OfMrCRd2nP2ZQ5eleDZC41XHyCNgktBGYMbqnrJKq/K/lzPMSQ==",
-      "cpu": [
-        "arm"
-      ],
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "android"
-      ],
-      "peer": true,
-      "engines": {
-        "node": ">=18"
-      }
-    },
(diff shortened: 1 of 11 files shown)


tashfeenahmed:
Thanks for the thorough review and the fix branch — really appreciated the structured writeup in #35.

I landed the safe subset in #40 (merged as 8415820), with credit to you in README contributors:

- ✅ Parameterized analytics SQL (all 5 endpoints)
- ✅ Sort-preset whitelist map in fallback.ts
- ✅ `crypto.timingSafeEqual` for the unified API key compare in proxy.ts
- ✅ Mid-stream error sanitization (full message logged internally, generic to client)

What I deferred from this PR and why:

- **Auth required on all `/api/*` and removing the `/v1` localhost bypass** — the React dashboard's `apiFetch` doesn't send Authorization today, and there's no key-prompt UI / localStorage flow. Landing the auth middleware here would 401 every dashboard page on first load (bootstrap chicken-and-egg: dashboard discovers its own key via `GET /api/settings/api-key`). This needs to be paired with client work. Happy to take a follow-up PR that does both together if you're up for it, otherwise I'll get to it later.
- **`/v1` localhost bypass** — on closer inspection, the XFF spoof in #35 issue #1 isn't actually exploitable in the current code: Express's default is `trust proxy = false`, so `req.ip` is the real TCP socket peer, not from `X-Forwarded-For`. The bypass is safe today. Added a comment in proxy.ts flagging that this assumption MUST be re-checked if anyone ever enables `trust proxy`.
- **Per-IP rate limit** — coupled to the auth-everywhere change to be most effective; deferring with that.
- **CORS tightening** — `cors({origin: true, credentials: true})` reflects any Origin anyway, so the practical effect once auth exists is minimal. Deferring to the same follow-up.
- **502 non-retryable error sanitization** — I kept the upstream provider message in the 502 path. For a single-user self-hosted setup, that detail is the most useful debug signal in a 502; sanitizing it makes "why did this fail" harder to answer. The mid-stream sanitization is in because mid-stream errors are already in a weird partial state.

Closing this PR in favor of #40. Thanks again — the review was high-signal even where I ended up disagreeing on a detail.

PR #53, opened 2026-05-21
feat(router): treat model "auto" as fallback-strategy routing

### LINKED ISSUE #48: Opencode config
Thanks for this! I've got it working on playground. But cannot get it to work in opencode.
It was giving me errors to remove from models in the config
but removing models has now stopped it appearing!

This is the start of my current opencode config

thanks!
{
  "$schema": "https://opencode.ai/config.json",
  "provider": {
    "freellmapi": {
      "npm": "@ai-sdk/openai-compatible",
      "name": "FreeLLMAPI (Unified)",
      "options": {
        "baseURL": "http://localhost:3001/v1",
        "apiKey": "{env:freellmapi-THISISTHEKEYHERE}"}
    },

also tried removing the env:

what do I put in for models?
thanks!

supertorpe:
## Summary
Adds support for `"model": "auto"` in `/v1/chat/completions`, allowing clients like OpenCode that always send a fixed `model` field to opt into fallback-strategy routing instead of being pinned to a specific model.
## Motivation
Closes #48. OpenCode (and similar OpenAI-compatible clients) always send a `model` field with every request, making it impossible to use FreeLLMAPI's auto-routing. The only workaround was to explicitly list individual models in the client's config, which defeated the purpose of the fallback chain. When users tried removing the `model` field entirely, the client would error out.
## Change
**`server/src/routes/proxy.ts:266`** — One-line change: widen `if (requestedModel)` to `if (requestedModel && requestedModel !== 'auto')`. When `model` is `"auto"`, the request behaves identically to omitting the `model` field:
1. Sticky session lookup is attempted first (preserves model continuity across conversation turns)
2. Falls through to `routeRequest()` which iterates the fallback chain (respecting dynamic penalties, rate limits, budgets, and key availability)
**`server/src/__tests__/integration/full-flow.test.ts`** — New Step 13 verifies that `model: "auto"` is accepted (not rejected as unknown/disabled) and routes through the normal fallback mechanism.
## Testing
- All 99 tests pass (14 test files)
- New integration test confirms `model: "auto"` produces a non-400 response

### DIFF
--- a/server/src/__tests__/integration/full-flow.test.ts
+++ b/server/src/__tests__/integration/full-flow.test.ts
@@ -174,4 +174,12 @@ describe('Full Integration Flow', () => {
     expect(body.error.code).toBe('model_not_found');
     expect(body.error.message).toContain('is disabled');
   });
+
+  it('Step 13: Explicit model "auto" is accepted and routes via fallback strategy', async () => {
+    const { status } = await req(app, 'POST', '/v1/chat/completions', {
+      model: 'auto',
+      messages: [{ role: 'user', content: 'hi' }],
+    });
+    expect(status).not.toBe(400);
+  });
 });
--- a/server/src/routes/proxy.ts
+++ b/server/src/routes/proxy.ts
@@ -263,7 +263,7 @@ proxyRouter.post('/chat/completions', async (req: Request, res: Response) => {
   // different model would be surprising to OpenAI-compatible clients.
   // Sticky-session is the fallback when no `model` field was sent at all.
   let preferredModel: number | undefined;
-  if (requestedModel) {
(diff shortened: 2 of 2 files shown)


tashfeenahmed:
Thank you for raising this @supertorpe 🙏 — your one-line fix was exactly the right read of the problem. While reviewing your PR, #57 landed with a slightly larger scope (also exposing `auto` in `GET /v1/models` so OpenAI-compatible UIs can list it), which closed the loop for clients like Hermes / OpenCode that auto-populate model dropdowns.

The combined fix is now in `main` via #62 (`21215502`). Closing this in favor of the superset PR — credit to you for catching the original docs-vs-code mismatch.

PR #56, opened 2026-05-21
Improve frontend UI

moyaqoob:
Things I did here
   -redesign the chat interface and improve the chatting. 
   -the chatbar always have a focus on the cursor, previously it didnt
   - add a fallback page incase if a model fails to respond. it will give a good error. 

image
image


tashfeenahmed:
Thanks @moyaqoob. Closing because the branch doesn't build — the existing UI components and routing wiring need to stay aligned with the server-side keys/health/fallback contracts, and several of the changes here break that.

If you'd like to take another pass: the V11/V12/V13/V14 catalog work changed a lot of the Platform / FallbackPage code paths, so I'd recommend starting from current main and proposing smaller individual UI polish PRs rather than a big rewrite.

PR #57, opened 2026-05-22
feat(proxy): add `auto` as a real /models option (#50)

### LINKED ISSUE #50: Add `auto` as an actual /models option
Some agents like Hermes require sending some model-id to the OpenAI endpoint.

Currently if I want to point Hermes at Freellmapi, I need to write some shim that discards the model-id so that freellmapi doesn't see the field. Adding a hardcoded model option like `auto` or `freellmapi-auto` would make this much nicer.

mybropro:
Closes #50.

## Problem

Some OpenAI-compatible clients (e.g. Hermes, opencode) require a non-empty `model` field on every request. freellmapi's whole point is to pick the model itself, so today you have to write a shim that strips the `model` field before it reaches the proxy — otherwise you get a `400 model_not_found`. `auto` is mentioned in the README but never actually worked, and doesn't appear in `/v1/models`.

## Change

Adds a virtual `auto` model that means "let the router decide". Identical behavior to omitting the `model` field entirely.

- **`GET /v1/models`** now lists `auto` as the first entry (`owned_by: "freellmapi"`), ahead of the real catalog.
- **`POST /v1/chat/completions`** treats `model: "auto"` as auto-route via the existing sticky-session logic, instead of returning a 400. This is intercepted before catalog validation, so both the streaming and non-streaming paths share the same `preferredModel` resolution.
- Unknown models still return `400 model_not_found`; the error message now points users at `auto`.

## Tests

New `proxy-auto-model.test.ts` covers:
- `auto` listed first in `/v1/models`
- `model: "auto"` auto-routes (200, not 400)
- unknown models still rejected with `model_not_found`

Full server suite: passing.

### DIFF
--- a/server/src/__tests__/routes/proxy-auto-model.test.ts
+++ b/server/src/__tests__/routes/proxy-auto-model.test.ts
@@ -96,40 +96,6 @@ describe('Virtual "auto" model', () => {
     expect(body.choices[0].message.content).toBe('routed via auto');
   });
 
-  it('accepts the freellmapi-auto alias too', async () => {
-    const origFetch = global.fetch;
-
-    vi.spyOn(global, 'fetch').mockImplementation(async (url, init) => {
-      const urlStr = typeof url === 'string' ? url : url.toString();
-      if (urlStr.includes('api.groq.com/openai/v1/chat/completions')) {
-        return {
-          ok: true,
-          json: () => Promise.resolve({
-            id: 'chatcmpl-auto-alias',
-            object: 'chat.completion',
-            created: 123,
-            model: 'openai/gpt-oss-120b',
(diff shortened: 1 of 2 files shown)


tashfeenahmed:
Thank you for this @mybropro 🙏 — the `/v1/models` exposure was the missing piece (#53 only handled the routing side). I cherry-picked your implementation onto current `main` (post-#55 added a Bearer requirement on `/v1/chat/completions`, so your test file needed `Authorization` headers — small adaptation, no logic change).

Merged in #62 (`21215502`), and you've been added to the contributors grid in the README.

Closing this PR in favor of the cherry-pick. Thanks again!

PR #69, opened 2026-05-22
fix(schema): accept array content in messages (OpenAI multimodal format)

jhash:
Fixes a validation error when clients send array-format message content, which is valid per the OpenAI spec but was rejected by the Zod schema with `"Expected string, received array"`.

This was motivated by getting the Pi coding agent working locally — it sends `content: [{type:"text",text:"..."}]` rather than `content: "..."`.

**Changes:**
- `shared/types.ts`: widen `ChatMessage.content` to `string | ContentBlock[] | null`
- `proxy.ts`: update all four message schemas (`system`, `user`, `assistant`, `tool`) to accept both formats
- `cloudflare.ts`, `cohere.ts`: flatten `ContentBlock[]` → string before forwarding (upstream APIs expect `string | null`)

Both formats still accepted; existing string-content clients unaffected.

---

*Apologies in advance — these changes were made via AI-assisted vibe coding (Claude Code). I've reviewed the output and tests pass, but flagging it in case anything looks off.*

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -7,7 +7,8 @@
     "dev": "vite",
     "build": "tsc -b && vite build",
     "lint": "eslint .",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "test": ""
   },
   "dependencies": {
     "@base-ui/react": "^1.3.0",
--- a/server/src/providers/cloudflare.ts
+++ b/server/src/providers/cloudflare.ts
@@ -22,10 +22,14 @@ export class CloudflareProvider extends BaseProvider {
 
   // Cloudflare's OpenAI-compat endpoint rejects `content: null` on assistant
   // messages that carry tool_calls, even though the OpenAI spec allows it.
+  // Also flatten ContentBlock[] arrays to plain strings since Cloudflare expects string | null.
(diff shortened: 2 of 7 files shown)


tashfeenahmed:
Thanks for this @jhash 🙏 — accepting OpenAI's multimodal array envelope was clearly the right call. I cherry-picked your work onto current main with a few polish changes:

- Extracted a shared `contentToString()` / `flattenMessageContent()` helper in `server/src/lib/content.ts` instead of duplicating the flatten logic across providers.
- Audited every provider: turns out the 12 platforms behind `OpenAICompatProvider` (Groq, Cerebras, SambaNova, NVIDIA, Mistral, OpenRouter, GitHub, Z.ai, Ollama, Kilo, Pollinations, LLM7) all accept arrays natively, so only Cohere, Cloudflare, and Google need flattening — saved a lot of churn.
- Google now extracts text into Gemini's \`parts:[{text}]\` shape (was silently dropping arrays before).
- Added 14 tests (8 unit + 6 integration covering schema acceptance, image_url dropped, full 200 round-trip, regression guards).
- Dropped the unrelated \`"test": ""\` script entry in \`client/package.json\`.

Merged in #75 (\`f2ece37\`), and you've been added to the contributors grid in the README. Thanks again!

PR #70, opened 2026-05-22
fix(google): Remove unsupported additionalProperties

jhash:
Google's API rejects the `additionalProperties` keyword in JSON schemas, causing 502 errors. This PR removes `additionalProperties` from tool parameter schemas before sending them to Google's API.

Builds off of: https://github.com/tashfeenahmed/freellmapi/pull/69

Looks like it might be duplicated by: https://github.com/tashfeenahmed/freellmapi/pull/65

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -7,7 +7,8 @@
     "dev": "vite",
     "build": "tsc -b && vite build",
     "lint": "eslint .",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "test": ""
   },
   "dependencies": {
     "@base-ui/react": "^1.3.0",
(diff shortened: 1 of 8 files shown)


tashfeenahmed:
Thanks @jhash 🙏 same problem space landed via #65 (now in main), which uses a recursive walker with a deny-list set rather than hardcoded nested deletes. additionalProperties isn't in that set yet, though - happy to take a one-line follow up that adds it (and credit you for catching it). Closing this in favor of #65; ping me on a small PR for the deny list addition?

jhash:
@tashfeenahmed new PR here: https://github.com/tashfeenahmed/freellmapi/pull/84

Thank you!

PR #71, opened 2026-05-22
use sqlite-napi for better compatibility with another enviroments like bun

nglmercer:
Better SQLite doesn't work in environments like BUN; it was replaced by an alternative made in Rust.
https://www.npmjs.com/package/sqlite-napi
https://github.com/nglmercer/sqlite-napi

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -18,6 +18,7 @@
     "drizzle-orm": "^0.44.2",
     "express": "^5.1.0",
     "helmet": "^8.1.0",
+    "sqlite-napi": "^1.3.0",
     "zod": "^3.24.4"
   },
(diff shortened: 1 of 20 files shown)


tashfeenahmed:
Thanks for the contribution @nglmercer 🙏 appreciate the effort here. Closing this one though: Bun isn't a supported runtime for FreeLLMAPI, and better-sqlite3 is battle tested in our stack with an active release just this month. If Bun support ever lands on the roadmap, the canonical path is Bun's built-in bun:sqlite (3-6x faster than better-sqlite3 per Bun's own docs, no native module install), rather than a separate native binding. Thanks again for thinking about cross runtime compat!

PR #81, opened 2026-05-24
feat(keys): add per-provider enable/disable toggle on KeysPage

Sumit4codes:
Adds a Switch toggle per provider group on the KeysPage that enables or disables all API keys for that provider at once.

**Backend**: new PATCH /api/keys/platform/:platform endpoint toggles all keys for a given platform. Validates platform against the allowlist.

**Frontend**: per-provider Switch toggle in the provider group header. Calls the new endpoint and invalidates relevant query caches.

The router already respects api_keys.enabled, so no routing changes were needed - disabled keys are automatically skipped.


tashfeenahmed:
Per-provider toggle already landed via #80 (Sumit included both commits in that PR). Closing as superseded.

PR #89, opened 2026-05-25
Add Gemma 4 models and fix Google schema additionalProperties validation

m1nuzz:
This PR adds the new Google Gemma 4 models (26B and 31B IT) to the catalog and fixes a validation issue where Gemini would reject 'additionalProperties' in JSON schemas.

### DIFF
--- a/server/src/db/index.ts
+++ b/server/src/db/index.ts
@@ -49,6 +49,7 @@ export function initDb(dbPath?: string): Database.Database {
   migrateModelsV12(db);
   migrateModelsV13(db);
   migrateModelsV14(db);
+  migrateModelsV15(db);
   ensureUnifiedKey(db);
 
   console.log(`Database initialized at ${resolvedPath}`);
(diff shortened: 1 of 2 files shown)


tashfeenahmed:
Thanks @m1nuzz. Two concerns before this can land:

1. **The schema denylist additions are too aggressive.** Adding `anyOf`, `oneOf`, `allOf` (and `const`) to `GEMINI_UNSUPPORTED_SCHEMA_KEYS` will silently strip those keywords from valid tool schemas — but Gemini *does* support them (see https://ai.google.dev/api/caching#Schema). Removing them will downgrade well-formed tool calls. Please drop those four entries.

   `additionalProperties` is already handled by #84 which landed earlier today.

2. **The Gemma 4 model IDs need probe verification against the real `generativelanguage.googleapis.com` endpoint.** `gemma-4-26b-a4b-it` and `gemma-4-31b-it` are Hugging Face / Cloudflare-style identifiers; Google AI Studio's `/v1beta/models` listing uses different IDs (`models/gemma-2-...`, `models/gemma-3-...`). Can you paste the output of:
   ```
   curl -s 'https://generativelanguage.googleapis.com/v1beta/models?key=YOUR_KEY' | jq '.models[].name' | grep gemma
   ```
   so we can confirm the exact IDs Google AI Studio exposes them as on the free tier? V13 already added `@cf/google/gemma-4-26b-a4b-it` on Cloudflare; we want to know the Google-side IDs are real before duplicating.

Happy to merge once those two issues are addressed.

tashfeenahmed:
Thank you, @m1nuzz 🙏 — Gemma 4 (`gemma-4-31b-it` and `gemma-4-26b-a4b-it`) shipped in #181 via the Google AI Studio free tier. The Google JSON-Schema `additionalProperties` fix you included also already landed in main via #84 and #105, so both halves of this PR are covered. Credited on the Contributors wall in #183. Closing as done — thanks for the contribution!

PR #99, opened 2026-05-27
chore: update package-lock.json for Node.js v26

clubpassmarbella:
## Summary

- Regenerated `package-lock.json` after running `npm install` on Node.js v26.0.0
- Removes redundant top-level `esbuild` v0.28.0 peer dependency entry
- Strips `"peer": true` flags from `@esbuild/*` platform-specific optional binaries (resolved correctly without the flag on Node v26)

## Why

The existing lockfile was generated on an older Node version. On Node v26, npm resolves peer dependencies differently and produces a cleaner lockfile without the duplicate esbuild peer entries at the top level. No runtime behavior changes.

## Test plan

- [ ] `npm install` completes without errors on Node v26
- [ ] `npm run dev` starts the dev server correctly
- [ ] `npm run build` produces a valid production build


thedavidweng:
The current LTS version is **Node.js 24** (not 26).

tashfeenahmed:
Closing: CI runs Node 20 and 22, and a lockfile regenerated on Node 26 with peer flags stripped risks breaking those builds. The engines range is >=20 <25, so a Node 26 lockfile is ahead of what we support.

Issue #97, opened 2026-05-26
400 Bad Request / 502 Error when using Tools with Gemini (additionalProperties not supported)


AlexCoursi:
**Describe the bug**
When using an OpenAI-compatible client (like OpenClaw) that sends tool definitions using the standard JSON Schema, the proxy returns a 502 Bad Gateway (or 400 Bad Request).

The underlying cause is that the Google Gemini API is stricter than the OpenAI standard and rejects any payload containing the additionalProperties keyword in tool definitions.

**To Reproduce**

Use a client that follows the OpenAI tool-calling specification.
Send a request with tools defined in the payload.
The request includes the additionalProperties field in the JSON Schema of the tool parameters.
The Gemini API returns a 400 error, which the proxy reports as a 502.

**Expected behavior**
The proxy should sanitize the tool definitions by stripping out incompatible JSON Schema keywords (such as additionalProperties) before forwarding the request to the Google Gemini API, ensuring compatibility between the OpenAI standard and Gemini's requirements.

**Actual behavior**
The request fails with the following error:
Invalid JSON payload received. Unknown name "additionalProperties" at 'tools[0].function_declarations[0].parameters.properties...'

**Logs/Error Message**

`502 Provider error (Gemini): Google API error 400: Invalid JSON payload received. `

AlexCoursi:
Same problem with groq

tashfeenahmed:
Fixed on main, Gemini tool schemas are now sanitized before forwarding. sanitizeForGemini() strips additionalProperties (and other keys Gemini's Schema proto rejects) from tool parameters, landed in #65 and #84. Broader keyword coverage (const, oneOf, examples, etc.) is in flight via #105. Closing as resolved... but please reopen if you still hit it on latest.

Thanks for the report!

PR #105, opened 2026-05-27
fix(google): strip every JSON Schema key not in Google's Schema proto (examples/const/readOnly/writeOnly/uniqueItems/not/allOf/oneOf/...)

Tushar49:
## Problem

When clients send JSON Schema with keys like `examples`, `const`, `readOnly`, `writeOnly`, `uniqueItems`, or any of the composition keywords (`not`, `allOf`, `oneOf`, `prefixItems`, `contains`, `propertyNames`, `multipleOf`, `dependencies`, `deprecated`) in tool/function parameters, the Google provider forwards them to the Gemini API, which rejects the request with:

```
400 Invalid JSON payload received. Unknown name "<key>" at 'tools[0].function_declarations[0].parameters...': Cannot find field.
```

This affects any client that emits richer JSON Schema (e.g. OpenCode, LangChain, MCP tool schemas, Zod-generated schemas), and breaks tool calling on every Gemini and Gemma model -- including Gemma 4 26B A4B IT and Gemma 4 31B IT, which natively support function calling via AI Studio.

Refs #97.

## Fix

`server/src/providers/google.ts` -- expand `GEMINI_UNSUPPORTED_SCHEMA_KEYS` to cover **every** JSON Schema keyword that is NOT in Google's [`Schema` proto](https://ai.google.dev/api/caching#Schema). Google's supported set is: `type, format, title, description, nullable, default, enum, maxItems, minItems, properties, required, minProperties, maxProperties, minLength, maxLength, pattern, example, anyOf, propertyOrdering, items, minimum, maximum`. Everything else must be stripped.

## Testing

- **10 unit tests** in `server/src/__tests__/providers/google-schema.test.ts`, all passing.
- **Live verification against Google AI Studio** with a rich schema payload (uniqueItems + pattern + minLength + maxLength + examples + default + additionalProperties) -- all three models that were 400-ing returned proper `tool_calls`:
  - `gemma-4-26b-a4b-it` -> 200, `tool_calls: [{name: "test_tool", arguments: {...}}]`
  - `gemini-3-flash-preview` -> 200
  - `gemini-3.5-flash` -> 200

The `uniqueItems` rejection was confirmed via a verbatim Google API response captured during a live freellmapi session:

```
"Invalid JSON payload received. Unknown name \"uniqueItems\" at
 'tools[0].function_declarations[0].parameters.properties[2].value':
 Cannot find field."
```

## Why this matters for Gemma users

Gemma 4 IT models on AI Studio have **exceptionally generous quotas** -- far higher than the Gemini family:

| Model              | RPM | TPM       | RPD   |
|--------------------|-----|-----------|-------|
| gemma-4-26b-a4b-it | 15  | Unlimited | 1,500 |
| gemma-4-31b-it     | 15  | Unlimited | 1,500 |
| gemini-2.5-flash   | 5   | 250K      | 20    |

Unlimited TPM + 1,500 RPD makes Gemma 4 IT one of the best free tool-calling backends available to any developer routing through freellmapi -- but only if the schema sanitizer does not break tools first. This patch unblocks that.

## Commits in this PR

1. `482ee11` -- fix(google): strip examples/const/readOnly/writeOnly from Gemini schema sanitization
2. `5ba759d` -- test(google): cover examples/const/readOnly/writeOnly stripping
3. `32ae8ba` -- fix(google): strip uniqueItems + composition keys (not/allOf/oneOf/prefixItems/contains/propertyNames/multipleOf/dependencies/deprecated)
4. `21c01aa` -- test(google): cover uniqueItems and composition-keyword stripping

## Checklist

- [x] Bug fix (non-breaking)
- [x] 10 unit tests added and passing
- [x] Manually verified against live Gemini + Gemma endpoints
- [x] No changes to public API or config shape
- [x] Strip list cross-checked against Google's official Schema proto field list

### DIFF
--- a/server/src/__tests__/providers/google-schema.test.ts
+++ b/server/src/__tests__/providers/google-schema.test.ts
@@ -149,4 +149,63 @@ describe('sanitizeForGemini', () => {
       required: ['command'],
     });
   });
+
+  it('strips uniqueItems (Gemini rejects arrays with uniqueItems with 400)', () => {
+    // Confirmed via live Google AI Studio call:
+    // "Invalid JSON payload received. Unknown name 'uniqueItems' at
+    //  'tools[0].function_declarations[0].parameters.properties[2].value':
+    //  Cannot find field."
+    const input = {
+      type: 'object',
+      properties: {
+        files: {
+          type: 'array',
+          items: { type: 'string' },
+          uniqueItems: true,
+          minItems: 0,
+          maxItems: 10,
+        },
+      },
+    };
+    expect(sanitizeForGemini(input)).toEqual({
+      type: 'object',
+      properties: {
+        files: {
+          type: 'array',
+          items: { type: 'string' },
+          minItems: 0,
+          maxItems: 10,
+        },
+      },
+    });
+  });
+
+  it('strips composition / structural keywords absent from Google Schema proto', () => {
+    // not / allOf / oneOf / prefixItems / contains / propertyNames / multipleOf /
+    // dependencies / deprecated are all absent from Google's Schema proto
+    // (https://ai.google.dev/api/caching#Schema). Only anyOf is supported.
+    const input = {
+      type: 'object',
+      not: { type: 'null' },
+      allOf: [{ type: 'object' }],
+      oneOf: [{ type: 'string' }, { type: 'number' }],
+      dependencies: { a: ['b'] },
+      deprecated: true,
+      properties: {
+        tuple: { type: 'array', prefixItems: [{ type: 'string' }] },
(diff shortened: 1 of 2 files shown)


Tushar49:
Friendly nudge @tashfeenahmed — this one is still green and self-contained: strips JSON Schema keys (examples / const / readOnly / writeOnly / uniqueItems / not / allOf / oneOf / ...) that Google's Schema proto rejects, so tool-calls on Gemini stop failing for any client that emits a richer JSON Schema. Happy to rebase if there's been drift, or address any review feedback. Thanks!

PR #112, opened 2026-05-28
docs: add OpenCode integration guide

johnnyhliang:
## Summary

- Adds an **OpenCode** section to the README showing how to configure FreeLLMAPI as a custom provider
- Covers global and per-project config file locations
- Recommends `auto` as the default model so it works regardless of which providers are configured
- Includes a note that the model catalog changes and points readers to the Fallback Chain page / `/v1/models` for the live list

## Test plan

- [ ] Docs-only change — no code modified, existing test suite passes (`npm test`: 143/143)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Aggregate the free tiers from Google, Groq, Cerebras, SambaNova, NVIDIA, Mistral
 - [Not yet supported](#not-yet-supported)
 - [Quick start](#quick-start)
 - [Using the API](#using-the-api)
+- [OpenCode](#opencode)
 - [Screenshots](#screenshots)
 - [How it works](#how-it-works)
 - [Limitations](#limitations)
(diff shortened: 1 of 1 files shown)


tashfeenahmed:
Still useful, but the README churned a lot so this needs a rebase. Suggestion: make it docs/integrations/opencode.md with a one-line README link instead; the README is oversized already and we're moving integration guides out of it.

tashfeenahmed:
Thanks for writing this up, and sorry it sat open for so long.

Since you opened it, the CLI grew a `setup-opencode` command that writes this exact config file for you, and the per-client recipes moved into [docs/clients.md](https://github.com/tashfeenahmed/freellmapi/blob/main/docs/clients.md) so the README could stay short. Between those two, the section here is covered, and that is the only reason I am closing it.

I appreciate you taking the time to document it properly, examples and all. If you ever spot something missing or wrong in the OpenCode row of docs/clients.md, a PR there would be very welcome.

johnnyhliang:
Awesome! Yeah I kind of forgot about this but it's great you guys implemented the generator.

PR #44, opened 2026-05-19
Add Docker and GHCR publishing support

thedavidweng:
## Summary
- add a multi-stage Dockerfile that builds the server and dashboard into a non-root production image
- add Docker Compose with persistent SQLite storage and health checks
- add a GHCR workflow that builds PRs and publishes images from main/tags
- document Docker Compose as the recommended install path with operations and curl examples
- align the root npm test script with the current CI server test suite

Closes #45.

## Verification
- npm test
- npm run build
- docker build -t freellmapi:local .
- docker run --rm -d --name freellmapi-smoke -e ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000 freellmapi:local
- docker inspect --format '{{.State.Status}} {{.State.Health.Status}}' freellmapi-smoke -> running healthy
- docker compose version -> Docker Compose version 5.1.3
- docker compose config
- ruby YAML parse for docker-compose.yml and .github/workflows/docker.yml

PR #116, opened 2026-05-28
Add Docker setup and compose config (Readme and DockerIgnore)

m-shalll:
Add multi-stage Dockerfile, docker-compose.yml and docker/README.md to enable containerized builds and local development. Include .dockerignore and update .gitignore to exclude local envs and tooling files. 
The Dockerfile builds the app, prunes dev deps, and produces a minimal runtime image; docker-compose provides a convenient service definition mapping port 3001 and mounting persistent data and .env.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+.git
+.github
+node_modules
+**/node_modules
+dist
+**/dist
+repo-assets
+coverage
+.env
+.env.*
+Dockerfile
+docker-compose.yml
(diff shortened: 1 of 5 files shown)


tashfeenahmed:
Thanks for this. Closing as superseded by #129, which adopted @thedavidweng's #44 (with multi-arch and a localhost-bind default added on top). Really appreciate the effort, the side-by-side helped us land on the most complete approach.

Issue #117, opened 2026-05-29
Generic Open AI compatible provider


punk1234567:
Hi Team,

First, thank you for building the tool. 

I would like to request support for a generic OpenAI-compatible provider. While the platform already supports many providers, there are an increasing number of services and self-hosted solutions that expose OpenAI-compatible APIs. 

It would be great if users could add a custom provider by specifying:

* Provider name
* Base URL
* API key
* Optional model ID for validation

I believe this would make FreeLLMAPI significantly more flexible and future-proof.

Thank you for considering the request.

Best regards,
Sam

chongjiazhen:
+1 — llama.cpp's `llama-server`, LM Studio, and vLLM all speak OpenAI `/v1` natively, so a generic base_url + optional key provider picks them up too. I run a local llama-server fleet I'd slot in behind the cloud tiers.

tashfeenahmed:
Shipped on `main` in #145. 🎉

freellmapi now has a generic **custom OpenAI-compatible provider** — point it at any OpenAI-compatible endpoint (llama.cpp `llama-server`, LM Studio, vLLM, a local Ollama, or a remote gateway) and route to its models. The API key is optional (most local servers don't need one).

**Dashboard:** Keys page → "Add a custom OpenAI-compatible model" (base URL + model + optional display name/key). Add one row per model; they share the one endpoint.

**API:**
```bash
curl -X POST http://localhost:3001/api/keys/custom \
  -H 'Content-Type: application/json' \
  -d '{"baseUrl":"http://127.0.0.1:11434/v1","model":"qwen3:4b","displayName":"Local Qwen3 4B"}'
```

The model joins the fallback chain like any other (no rate limits; sorts last in the intelligence preset).

🙏 Big thanks to @Aldo-f (#121) and @xristi1 (#104) — their local-provider PRs prototyped the per-key `base_url` approach this generalizes.

**Scope note:** this supports one shared custom endpoint (one base URL, many models), which covers @chongjiazhen's local-fleet case. Multiple *distinct* custom endpoints, and model auto-discovery via the endpoint's `/v1/models` (which #121 prototyped), are natural follow-ups. Closing as implemented.

PR #121, opened 2026-05-29
feat: add local Ollama provider with sync endpoint and tests

Aldo-f:
## Summary

Adds support for a local Ollama provider (`ollama-local`) with an on-demand sync endpoint that imports models from a local Ollama instance into the database.

### New files
- **`server/src/routes/sync.ts`** — `GET /api/sync` endpoint that:
  - Requires `provider=ollama-local` query parameter (returns 400 if missing/invalid)
  - Fetches model list from the local Ollama Tags API
  - Filters out cloud stubs (`:cloud` suffix, `ollama:` prefix)
  - Inserts/ignores models into the DB
  - Creates fallback config entries for newly inserted models
  - Returns `{ success, imported_count, models }` — tracks only **newly inserted** models via `INSERT OR IGNORE` + `result.changes`
- **`server/src/__tests__/routes/sync.test.ts`** — full test suite:
  - Uses a mock HTTP server to avoid external dependency on local Ollama
  - Tests: missing provider (400), no providers configured (404), successful import (200), invalid provider (400), idempotency (second call returns 0 imported)

### Modified files
- **`server/src/routes/sync.ts`** — updated to add provider validation, improve response with accurate `imported_count` tracking
- **`shared/types.ts`** — adds `ollama-local` to platform types
- **`server/src/services/router.ts`** — routes `ollama-local` keys correctly
- Various provider/routing files to support the new platform

## Commits

@xristi1:
1. `2e1d442` — Add local Ollama support
2. `11e97bb` — Fix fallback when local Ollama is unavailable

@Aldo-f:
3. `00c4bd3` — feat: add ollama-local provider with baseUrl and sync endpoint
4. `b4c1425` — feat(sync): update sync endpoint + test(sync): add tests

## Testing

All 156 server tests pass. Sync tests are self-contained (mock Ollama server), no external dependencies.


tashfeenahmed:
Thank you for this, @Aldo-f — genuinely solid work (clean `base_url`-on-`api_keys` design, the `/api/sync` discovery endpoint, and a real test suite incl. Playwright e2e).

I've landed a **generic custom OpenAI-compatible provider** in #145 that generalizes the per-key `base_url` approach you prototyped here into a provider-agnostic `custom` platform — so it covers local Ollama *and* llama.cpp / LM Studio / vLLM / remote gateways through one path (see #117). Because it supersedes the `ollama-local`-specific plumbing, I'm closing this PR — but it directly informed the design and you're credited in the #145 commit and the #117 thread.

Two ideas from this PR I'd love as follow-ups on top of the new endpoint: **model auto-discovery** (your `/api/sync` querying `/v1/models`) and **multiple distinct custom endpoints**. If you're up for a smaller PR adding sync against the new `custom` provider, that'd be a great next step. 🙏

Aldo-f:
The `api/sync` could be a real asset. Also instead database updates we could make use of endpoints with filters on (probably) free models. I was thinking about a way that we could sort models by a real 'intelligence'.


Next would be a 'sorting on success rate'.

Other ideas:
- sorting with sliders (60% intelligence, 40% speed, ...)
- and a real check if models are valid (a small "Say 'Hi'" request to sort models by real values (speed, success rate, ...)
- Multiple keys (one key for each order of (active) models
- Move models to top (as i have access to a very long list)
- [Toast](https://www.npmjs.com/package/react-toastify)/[sonner](https://ui.shadcn.com/docs/components/radix/sonner) for when we hit some buttons that would trigger actions
- A "Are you sure" (confirmation) when you click "Regenerate"
- A upload keys by file (.env and .json)

PR #127, opened 2026-05-29
fix(api): emit UTC-marked ISO timestamps so the dashboard localizes correctly

mybropro:
## Problem

Dashboard timestamps (Recent Errors on the Analytics page, key `created`/`last checked` on Health/Keys) display in the wrong time — shifted by the viewer's UTC offset rather than shown in their local time.

**Root cause:** `created_at`/`last_checked_at` are stored via SQLite `datetime('now')` (`server/src/db/index.ts`), which produces a naive, space-separated UTC string like `2026-05-29 04:17:11` — no `T`, no `Z`. The API returns it raw, and the client does `new Date(value)`. Browsers (V8) parse that non-standard space-separated form as **local** time, so a UTC value gets relabeled as local with no conversion, and every timestamp is off by the viewer's offset.

```
stored "2026-05-29 04:17:11"  (UTC)
  new Date(space-form)  -> Fri May 29 2026 04:17:11 GMT-0400   (parsed as LOCAL — wrong)
  toLocaleTimeString    -> 04:17 AM                            (should be 12:17 AM EDT)
```

## Fix

Tag these timestamps as ISO-8601 UTC (`2026-05-29T04:17:11Z`) at the API boundary so `new Date(...)` parses them correctly and the browser localizes. Adds a small `toIsoUtc()` helper (`server/src/lib/time.ts`) applied to the timestamp fields returned by the analytics, keys, and health routes. Storage stays UTC (correct) — this only marks the zone on the way out. Strings already carrying a `T`/`Z` are passed through unchanged.

## Test plan

- `npm run build` (server) passes.
- Verified on a live instance: `/api/analytics/errors` now returns `createdAt` like `2026-05-29T18:14:49Z`, and the dashboard's Recent Errors / key timestamps render in the viewer's local timezone.

### DIFF
--- a/server/src/routes/analytics.ts
+++ b/server/src/routes/analytics.ts
@@ -219,8 +219,12 @@ analyticsRouter.get('/errors', (req: Request, res: Response) => {
   const since = getSinceTimestamp(range);
   const db = getDb();
 
+  // created_at is stored as a naive UTC string ("YYYY-MM-DD HH:MM:SS"); emit it
+  // as ISO-8601 with a 'Z' so the dashboard's new Date(...) parses it as UTC and
+  // localizes it (the space-separated form is otherwise parsed as local time).
   const rows = db.prepare(`
-    SELECT id, platform, model_id, error, latency_ms, created_at
+    SELECT id, platform, model_id, error, latency_ms,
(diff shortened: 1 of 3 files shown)


mybropro:
Closing — handling this differently on our side for now. Thanks!

PR #138, opened 2026-05-31
feat(keys): per-key time-based disable with auto re-enable

EkmalRey:
## Summary

Adds the ability to disable individual API keys for a configurable duration (1h, 5h, 24h, 7 days, or forever) with automatic re-enable on expiry. Previously, only providers could be be toggled on/off permanently, this adds time-based control and a toggle per key so you can temporarily pause a key (e.g. during rate limit cooldown) without forgetting to re-enable it.

### Database (`server/src/db/index.ts`)
- Non-destructive migration adds `disabled_until TEXT` column to `api_keys`
- Uses `PRAGMA table_info` check to avoid duplicate columns on re-run

### Backend (`server/src/routes/keys.ts`)
- `PATCH /api/keys/:id` now accepts optional `disableDuration` alongside `enabled` and `label`
- Supported values: `1h`, `5h`, `24h`, `7d`, `forever`

**Examples:**

Disable a key for 1 hour:
```json
PATCH /api/keys/42
{ "enabled": false, "disableDuration": "1h" }
```
Response:
```json
{ "success": true, "enabled": false }
```

Disable a key forever (until manually re-enabled):
```json
PATCH /api/keys/42
{ "enabled": false, "disableDuration": "forever" }
```

Re-enable a key (clears any timer):
```json
PATCH /api/keys/42
{ "enabled": true }
```

Edit label (existing feature, unchanged):
```json
PATCH /api/keys/42
{ "label": "my updated label" }
```

`disableDuration` is validated — invalid values return `400`:
```json
{ "error": { "message": "disableDuration must be one of: 1h, 5h, 24h, 7d, forever" } }
```

### Router (`server/src/services/router.ts`)
- Key selection query now filters out temporarily disabled keys:
  ```sql
  AND (disabled_until IS NULL OR disabled_until < datetime('now'))
  ```

### Health Service (`server/src/services/health.ts`)
- On each health check cycle (every 5 min), re-enables keys whose `disabled_until` has passed:
  ```sql
  UPDATE api_keys SET enabled = 1, disabled_until = NULL
  WHERE enabled = 0 AND disabled_until IS NOT NULL AND disabled_until < datetime('now')
  ```
- Logs how many keys were re-enabled

### Frontend (`client/src/pages/KeysPage.tsx`)
- Each key row now has its own toggle Switch (previously only per-platform toggles existed)
- Toggling OFF opens a duration picker with options: 1h, 5h, 24h, 7d, Until I turn it on
- Temporarily disabled keys show remaining time in amber: `disabled · 4h 32m`
- Permanently disabled keys show `disabled` in muted text
- Toggling back ON immediately re-enables the key (clears `disabled_until`)

### Shared Types (`shared/types.ts`)
- `ApiKey` interface updated with `disabledUntil: string | null`

## Testing
- All 148 tests pass (20 test files)
- Server build (`tsc`) clean
- Client build (`tsc -b && vite build`) clean
- Rebased on latest `main` — no conflicts


tashfeenahmed:
Thank you for the contribution, @EkmalRey — the per-key timed toggle with auto re-enable is nicely built, and the non-destructive migration is the right approach.

That said, the core motivation here (pausing a key during a rate-limit cooldown so you don't forget to re-enable it) is already handled automatically by the router/ratelimit layer, and arguably more precisely:

- **Penalty / priority demotion** (`router.ts`): every upstream 429 adds a penalty that sinks the model in the fallback order, capped and then decaying ~1 step every 2 minutes — and a success actively reduces it. So a rate-limited model floats back up on its own.
- **Scoped cooldown** (`ratelimit.ts`): a 429 benches that specific **model + key** combo (not the whole key). Transient RPM/TPM hits get a short cooldown; it only escalates toward a longer quarantine (up to 24h) when the daily quota is genuinely exhausted. Cooldowns are persisted in `rate_limit_cooldowns` and auto-expire — no manual cleanup.

Because the rate-limit case is already covered automatically (per model+key, with escalation and decay), a manual whole-key timed disable adds a DB column plus router/health surface area for a fairly narrow benefit. So I'm going to close this one for now.

Really appreciate the effort and the clean implementation — please don't let this discourage you from sending more PRs. 🙏

PR #160, opened 2026-06-01
feat(google): support OpenAI-compatible reasoning_effort for Gemini thinking models

Tushar49:
Adds OpenAI-compatible `reasoning_effort` parameter for the Google provider, mapping to Gemini's `thinkingConfig` so callers can opt into Gemini's thinking budget without leaving the OpenAI request shape.

## What this PR does

- **Schema**: `shared/types.ts` + `server/src/routes/proxy.ts` accept `reasoning_effort: "minimal" | "low" | "medium" | "high"` on `ChatCompletionRequest`. Omitting it preserves existing behavior exactly (no `thinkingConfig` is sent).
- **Mapping** (`server/src/providers/google.ts`):
  - **Gemini 3.x** → `thinkingConfig.thinkingLevel` (`MINIMAL` / `LOW` / `MEDIUM` / `HIGH`). `gemini-3.1-pro` does not accept `MINIMAL` → falls back to `LOW`.
  - **Other Gemini models** (2.x, etc.) → `thinkingConfig.thinkingBudget` integer. Canonical mapping: `minimal` → 512 (Pro: 128), `low` → 1024, `medium` → 8192, `high` → 24576. Clamped per model: `gemini-2.5-pro` min 128 / max 24576, Flash 0–24576, Flash-Lite 512–24576.
  - Always sets `includeThoughts: true` when `reasoning_effort` is provided.
- **Response parsing**: Gemini returns `candidates[0].content.parts[]` where parts may carry `thought: true`. Non-thought parts go to `message.content` (unchanged); thought parts are aggregated into `message.reasoning_content` (OpenAI extension used by opencode and other clients). `usageMetadata.thoughtsTokenCount` → `usage.completion_tokens_details.reasoning_tokens`.
- **Streaming**: thought-part deltas emit as `delta.reasoning_content`, normal text deltas as `delta.content`.

## Why

OpenAI clients (opencode, Continue, Cursor, etc.) already use `reasoning_effort` to drive reasoning models. Today they have no way to dial Gemini's thinking budget through freellmapi without bypassing the proxy. Google's own [OpenAI-compatible endpoint](https://ai.google.dev/gemini-api/docs/openai) maps the same field to the same `thinkingConfig` — this PR brings freellmapi to parity.

## Verification

From `server/`:

- `bun run build` (tsc): clean
- `bun run test src/__tests__/providers/google-reasoning.test.ts`: **8 / 8 pass** (1 file)
- `bun run test src/__tests__/providers/google.test.ts src/__tests__/providers/google-schema.test.ts`: **19 / 19 pass** (no regressions to existing Google tests / PR #105 tests)

New tests cover: effort → budget mapping per model class, Gemini 3.x `thinkingLevel` path, `gemini-3.1-pro` MINIMAL → LOW fallback, `includeThoughts` on/off semantics, thought-part parsing into `reasoning_content`, `reasoning_tokens` in usage, and streaming `delta.reasoning_content`.

## Style / scope

- Matches the four-commit conventional-commit shape of #105 (`feat(api):` → `feat(google):` → `feat(proxy):` → `test(google):`).
- No new dependencies. No request-schema removals or renames.
- When `reasoning_effort` is omitted, `generationConfig.thinkingConfig` is not added to the outgoing request, so existing callers see byte-for-byte the same Gemini payload.

Happy to iterate on naming, default budgets, or split into smaller PRs if preferred.


tashfeenahmed:
This is the most valuable stale PR in the queue: reasoning_effort mapped to Gemini thinkingConfig is a real parity gap we want to close. Two asks: (1) rebase onto current main (google.ts gained per-call timeout plumbing in #461), and (2) keep the diff surgical rather than rewriting google.ts and proxy.ts wholesale; the smaller the diff, the faster it merges. Please also re-verify the model id list is current. If you don't have time, say so and we'll pick it up ourselves with credit for your groundwork.

tashfeenahmed:
Thanks for this, and sorry for the long silence on it.

Reasoning control for the Gemini models landed in #602 last week. It maps `reasoning_effort` onto `thinkingBudget` in much the same way you proposed here, so the behaviour is now in main and this PR is duplicated by it.

That is on me for leaving your PR sitting while the same ground got covered elsewhere, and I am sorry for the wasted effort. You identified the right gap and the approach you took was sound. Thanks for putting the work in.

PR #170, opened 2026-06-02
fix: local display of SQLite UTC time in Analytics and Keys pages

Tazrif-Raim:
## Description of change

Adds shared timestamp utilities for handling SQLite UTC datetime strings on the client.

SQLite stores timestamps like `YYYY-MM-DD HH:MM:SS`, which do not include a timezone marker. Passing these values directly to `new Date(...)` can make the browser treat them as local time instead of UTC.

This adds:

- `sqliteUtcToIso` to convert SQLite UTC datetime text into ISO UTC format.
- `formatSqliteUtcToLocalTime` to format SQLite UTC datetime text as the client's local time for display.

These utilities are now used in the existing timestamp displays and can be reused anywhere client-side SQLite UTC timestamps need to be shown as local time.

## Updated usage

Replaced direct usage of:

`new Date(value).toLocaleTimeString(...)`

with:

`formatSqliteUtcToLocalTime(value)`

in:

- `AnalyticsPage.tsx`
- `KeysPage.tsx`

## Screenshots

Previous:
image

fix:
image

### DIFF
--- a/client/src/lib/utils.ts
+++ b/client/src/lib/utils.ts
@@ -4,3 +4,27 @@ import { twMerge } from "tailwind-merge"
 export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs))
 }
+
+export function sqliteUtcToIso(value: string | null | undefined): string | null | undefined {
+  if (!value) return value;
+
+  return value.includes('T') ? value : `${value.replace(' ', 'T')}Z`;
+}
+
+// Converts SQLite UTC datetime text to the client's local time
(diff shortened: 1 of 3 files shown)


tashfeenahmed:
Thanks @Tazrif-Raim — shipped in #176 (v0.2.5) and live. The `sqliteUtcToIso` / `formatSqliteUtcToLocalTime` helpers now back both the Analytics and Keys timestamp displays. Credited you as co-author. 🙏

PR #188, opened 2026-06-03
feat(embeddings): replace EMBEDDINGS_GOOGLE_KEY env-var with provider-abstraction key store

nalepy:
## Problem

The current `/v1/embeddings` implementation (#173) hard-codes Google as the only provider and requires a separate `EMBEDDINGS_GOOGLE_KEY` environment variable. It lives inline in `proxy.ts`, has no tests, and can't fall back when the Google key is missing or rate-limited.

## Solution

Replace it with the same provider-abstraction pattern used everywhere else in the codebase.

## Changes

- **`routes/embeddings.ts`** (new) — dedicated router: zod validation, auto-routing (Google → Mistral fallback), encrypted key-store auth, per-key cooldown, request analytics
- **`providers/base.ts`** — adds optional `supportsEmbeddings` + `embeddings()` to `BaseProvider`
- **`providers/google.ts`** — implements `embedContent` (single input) and `batchEmbedContents` (array input)
- **`providers/openai-compat.ts`** — generic `embeddings()` method; wired to Mistral via `embeddingModels` constructor option
- **`providers/index.ts`** — Mistral registration gains `embeddingModels: ['mistral-embed']`
- **`shared/types.ts`** — adds `EmbeddingsRequest`, `EmbeddingObject`, `EmbeddingsResponse`
- **`db/index.ts`** — seeds `text-embedding-004`, `text-multilingual-embedding-002`, `mistral-embed` model rows
- **`routes/proxy.ts`** — removes the old inline handler and `EMBEDDINGS_GOOGLE_KEY` dependency
- **`app.ts`** — mounts `embeddingsRouter` after `responsesRouter`

## Migration

Remove `EMBEDDINGS_GOOGLE_KEY` from your `.env`. Add your Google (or Mistral) key through the Keys page — it will be picked up automatically.

## Test plan

- [ ] `npm run test -w server` — all tests pass (9 new embedding tests)
- [ ] Single string input normalised to array, returns correct OpenAI `{ object: "list", data: [...] }` shape
- [ ] Array input routes to `batchEmbedContents`
- [ ] Auto-routing: prefers Google, falls back to Mistral when no Google key is configured
- [ ] Returns `503 service_unavailable` when no keys are configured
- [ ] `text-multilingual-embedding-002` routes correctly to Google


server/src/routes/embeddings.ts
```diff
@@ -0,0 +1,106 @@
...
+    candidates.push({ platform: 'google', modelId: 'text-embedding-004' });
+    candidates.push({ platform: 'mistral', modelId: 'mistral-embed' });
+  } else if (m.startsWith('text-embedding') || m.startsWith('text-multilingual')) {
+    candidates.push({ platform: 'google', modelId: m });
+  } else if (m === 'mistral-embed') {
+    candidates.push({ platform: 'mistral', modelId: m });
+  } else {
+    candidates.push({ platform: 'google', modelId: 'text-embedding-004' });
+    candidates.push({ platform: 'mistral', modelId: 'mistral-embed' });
+  }
```
Tazrif-Raim, line 68: Vectors from different embedding models are not compatible, so `/embeddings` should require an explicit `model` instead of falling back to `auto`. I'd suggest to fail the request if `model` is not present in the request.

Tazrif-Raim (changes requested):

tashfeenahmed:
Thanks for putting this together, the direction was exactly right. This got superseded by #195, which landed a dedicated embeddings router with the same key-store approach plus family-based routing and in-family provider failover. EMBEDDINGS_GOOGLE_KEY is gone from main now. Appreciate the push on this, it shaped where the design ended up.

PR #198, opened 2026-06-04
fix: stop fallback chain from 502-ing on a dead first route

hodlmybeer69-bit:
## Problem

The proxy returned `no working model` for every auto-routed request, causing 502s on all traffic.

## Root Cause

1. **Dead Cerebras route**: `qwen-3-235b-a22b-instruct-2507` (created by V11's rename) returns 404 `model_not_found`. It was the first key-having route in the chain.
2. **Non-retryable 404**: `isRetryableError` treated 404 as fatal — the dead route short-circuited to a 502 without falling through to working google/groq routes.
3. **Restricted Groq key**: An account-restricted key returned HTTP 400 with the same short-circuit effect, blocking 2 of 3 Groq keys.

### `server/src/routes/proxy.ts`
- `404` / `model_not_found` / `unavailable_model` → **retryable** (skip dead route, fall through)
- `organization has been restricted` → **retryable** (skip banned key, fall through)
- Pinned-model requests still pre-validated against catalog (fast 400 for genuine typos)

### `server/src/db/index.ts`
- Add idempotent `migrateModelsV12`:
  - Disable dead cerebras id
  - Pin `groq/llama-3.3-70b-versatile` to priority 0 (verified working)

## Verification

- Plain + tool-call requests return 200 via groq/llama-3.3-70b
- All 98 server tests pass
- Proxy has been running live with this fix (PM2, zero 502s)


tashfeenahmed:
Thanks for the solid diagnosis — both root causes you identified were real, and both are now fixed on main independently: 404s are retryable since 8aea92c (#76), and the dead Cerebras route (plus 6 other dead models) was pruned in the V21 catalog migration (#184). This branch is based on a much older main, so its diff now removes recent migrations and tables (V13–V22, rate-limit/users tables), which makes it unmergeable as-is. Closing as superseded — the fixes you proposed are live. Appreciate the investigation!

PR #199, opened 2026-06-04
chore: bump better-sqlite3 to ^12.10.0 + sync lockfile

hodlmybeer69-bit:
## Summary

- Bumps `better-sqlite3` from `^12.4.1` → `^12.10.0` in `server/package.json`
- Syncs `package-lock.json` to match what is actually installed

## Why

The proxy was already running on 12.10.0 (auto-upgraded at `npm install`) while `package.json` still declared `^12.4.1`. This PR aligns the version declaration with the installed reality so the lockfile is not stale and future `npm ci` installs match the running binary.


tashfeenahmed:
Thanks for the bump! Unfortunately this branch is based on a much older main, so beyond the lockfile it now reverts recent server changes (840 lines of db migrations, 293 lines of proxy fixes) — that makes it unmergeable as-is. The better-sqlite3 bump itself is a good idea, so we're landing it freshly on current main. Appreciated!

PR #216, opened 2026-06-05
feat(custom-models): implement virtual platform namespace architectur…

phoenixikkifullstack:
implement virtual platform namespace architectur

- Implements 1:1 key-model binding for custom models using 'custom:ID' namespace
- Resolves issue where multiple custom models overwrite each other
- Adds data migration fission to split legacy custom keys
- Aggregates custom models in analytics SQL for clean charts
- Fixes PATCH /platform/custom toggle applying to virtual namespaces
- Refactors cascade deletion to only scope to the deleted key's namespace
- Restores vitest config to prevent parallel sqlite DB locking
- Restores google.test.ts mock state isolation

### DIFF
--- a/client/src/pages/KeysPage.tsx
+++ b/client/src/pages/KeysPage.tsx
@@ -203,7 +203,7 @@ function CustomProviderSection() {
       <h2 className="text-sm font-medium mb-1">Add a custom OpenAI-compatible model</h2>
       <p className="text-xs text-muted-foreground mb-3">
         Point at any OpenAI-compatible endpoint — llama.cpp, LM Studio, vLLM, a local Ollama, or a remote
-        gateway. Add each model you want routed; they all share the one endpoint. The API key is optional
+        gateway. Add each model you want routed; models sharing the same Base URL will group under the same endpoint below. The API key is optional
         (most local servers don't need one).
       </p>
       <form onSubmit={submit} className="flex flex-wrap items-end gap-3 rounded-3xl border p-4 bg-card">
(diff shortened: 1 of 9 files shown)


tashfeenahmed:
Thank you for this — you correctly identified the overwrite bug (#212) and the 1:1 key-model binding is exactly the right idea. Apologies for the awkward timing: a fix for the same issue landed in #218 about an hour after you opened this, before we'd seen it.

#218 takes a less invasive route to the same guarantee: models bind to their endpoint via a `models.key_id` column instead of namespacing the platform string, so analytics aggregation, platform toggles, the shared `Platform` type, and the client UI all stay untouched, and legacy installs get backfilled on migration. The virtual-namespace approach is clever but its blast radius (analytics SQL, PATCH /platform routing, client changes) is what tipped the decision.

Two things from your PR may still have standalone value — if you've seen real test flakiness, a focused PR for the vitest sqlite locking config and the google.test.ts mock isolation would be very welcome. Thanks again for the solid work here.

phoenixikkifullstack:
Thank you for the detailed review and for explaining the approach taken in #218! I completely agree that introducing a models.key_id column is a much less invasive and cleaner solution than the virtual-namespace approach.

I actually ran into this overwrite bug (#212) during my own daily usage, which motivated me to dive into the codebase and try to fix it. It's great to see it resolved so elegantly in the main branch.

I'd be more than happy to extract the test stability improvements (the vitest sqlite locking config and the google.test.ts mock isolation) into a separate, focused PR as you suggested.

I really enjoy using this project and will definitely continue to use it and contribute where I can in the future. Thanks again for your time and guidance!

PR #223, opened 2026-06-05
fix(test): resolve vitest cli error and mock leakage

phoenixikkifullstack:
- Remove invalid `--poolOptions` vitest CLI flag which causes `CACError: Unknown option` in Vitest 3+
- Use supported `--fileParallelism=false` flag to safely run sequential sqlite tests
- Add `vi.clearAllMocks()` to google.test.ts to prevent spy leakage from previously executed sequential test files

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -7,7 +7,7 @@
     "dev": "tsx watch src/index.ts",
     "build": "tsc",
     "start": "node dist/index.js",
-    "test": "vitest run --pool=forks --poolOptions.forks.singleFork=true --no-file-parallelism",
+    "test": "vitest run --pool=forks --fileParallelism=false",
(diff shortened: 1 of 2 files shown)


tashfeenahmed:
Merged, thanks! Verified on the actual installed vitest (3.2.4 via the ^3.1.3 range): the old dot-notation flags still parse there, so the CACError seems to come from newer vitest or a different shell, but the new form is cleaner either way and the isolation win is real. Dropping singleFork means each test file now runs in its own fork, which is exactly the cross-file spy leakage your google.test.ts fix addresses, fixed at the root. Full suite passes 383/383. Heads up that the suite takes ~17s now instead of ~3s due to per-file forks, a fair trade for hermetic tests.

Issue #225, opened 2026-06-05
Feature Request: Support Outbound HTTP/SOCKS Proxy for Provider Requests


homayounmohseni:
Hi,

It would be useful to support configuring an outbound proxy for requests sent to LLM providers.

This would help users who:

* Need to route traffic through a local proxy/VPN
* Are behind network restrictions
* Want more control over outbound connections

One possible approach would be to support standard proxy environment variables such as:

* HTTP_PROXY
* HTTPS_PROXY

or a dedicated application setting.

Ideally, HTTP, HTTPS, and SOCKS5 proxies could be supported.

Thanks!

tashfeenahmed:
Done — outbound HTTP/SOCKS proxy support shipped in #286 (rebased from @Slyker's #231). Configure via the Keys dashboard or the `PROXY_URL` env var; per-platform bypass and a global toggle are included. Off by default.

PR #231, opened 2026-06-05
feat(proxy): outbound SOCKS5/HTTP proxy with per-platform bypass

Slyker:
## What

Adds outbound proxy support so all LLM provider requests can be routed through a SOCKS5 or HTTP/HTTPS proxy. Configurable via the Keys dashboard or PROXY_URL environment variable.

## Why

Users behind restrictive networks or needing IP rotation (e.g. Webshare rotating proxies) need all outbound LLM calls to go through a proxy. Previously this wasn't possible — undici's experimental SOCKS5 support fails on common configurations (port 80 proxies).

## Changes

- **server/src/lib/proxy.ts** (new) — `proxyFetch()` drop-in replacement for `fetch()` that routes through the configured proxy. SOCKS5 via `socks-proxy-agent` (same stack as curl), HTTP/HTTPS via `undici.ProxyAgent`. Supports streaming (ReadableStream), AbortSignal, and configurable bypass.

- **server/src/providers/base.ts** — `fetchWithTimeout` calls `proxyFetch(url, init, this.platform)`. Single choke point for all 18 chat providers — no per-provider changes needed.

- **server/src/providers/google.ts** — Image inline fetch uses `proxyFetch` (standalone function).

- **server/src/services/embeddings.ts** — All `fetch()` → `proxyFetch()` (global toggle only, no per-platform bypass — embeddings use their own family routing).

- **server/src/routes/settings.ts** — New GET/PUT /api/settings/proxy for proxyUrl, enabled, and bypassPlatforms (CSV).

- **server/src/index.ts** — Loads proxy_url, proxy_enabled, proxy_bypass from DB on startup.

- **server/src/routes/proxy.ts** — Added `'fetch failed'` to isRetryableError so proxy transport errors trigger model failover instead of 502.

- **client/src/pages/KeysPage.tsx** — New Outbound proxy section: URL field, on/off toggle (preserves URL), per-provider bypass checkboxes.

- **server/package.json** — Added `socks-proxy-agent` dependency.

## Architecture

- **Single injection point:** `BaseProvider.fetchWithTimeout` is the choke point — all providers (Google, Cohere, Cloudflare, OpenAICompat...) use it. No per-provider changes.
- **SOCKS5 bypasses undici:** Node.js undici ProxyAgent fails on common SOCKS5 setups. We use socks-proxy-agent + native http/https instead — same stack as curl — with a proper Response shim for SSE streaming.
- **Embeddings: global toggle only.** They use a separate family routing system and don't participate in per-platform bypass (a chat-provider concern).

## How to test

Via dashboard (Keys tab): paste your proxy URL, click Save, toggle on/off.

Via env var (takes priority): `PROXY_URL="socks5://user:pass@p.webshare.io:80/" npm run dev`

## Checks

- ✅ TypeScript: 0 errors (server)
- ✅ Tests: 345/345 pass

### DIFF
--- a/client/src/pages/KeysPage.tsx
+++ b/client/src/pages/KeysPage.tsx
@@ -8,7 +8,7 @@ import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@
 import { Switch } from '@/components/ui/switch'
 import { PageHeader } from '@/components/page-header'
 import type { ApiKey, Platform } from '../../../shared/types'
-import { Pencil, ExternalLink } from 'lucide-react'
+import { Pencil, ExternalLink, Globe } from 'lucide-react'
 import { formatSqliteUtcToLocalTime } from '@/lib/utils'
 
 // Small "Get API key" external link shown next to a provider (#137).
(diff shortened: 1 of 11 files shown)


tashfeenahmed:
Thanks for this, the core architecture is right. One choke point in `BaseProvider.fetchWithTimeout` plus per-platform bypass is exactly how this should work. I want to land it, but a few things first:

**1. Split the catalog migrations out.** V23/V24 in `db/index.ts` (new OpenCode models, full intelligence re-rank) are unrelated to proxying, and V24 changes routing order for every existing user. Separate PR please, happy to review it on its own.

**2. Unify the transport on undici dispatchers.** There are two stacks here: a hand-rolled `http.request` + Response shim for SOCKS, and an npm undici 8 ProxyAgent passed into Node's built-in fetch for HTTP. Both are fragile: mixing a dispatcher from npm undici 8 with the global fetch (a different undici major) is a known compat risk around streaming, the SOCKS shim breaks on `Headers` instances and reimplements fetch by hand, and undici 8 needs Node >= 22.19 while this project supports Node 20 (Pi installs included).

Simpler fix: route every proxied request through `undici.fetch` with a dispatcher. [`fetch-socks`](https://github.com/Kaciras/fetch-socks) wraps the same `socks` lib you're using but exposes it as an undici dispatcher, so SOCKS and HTTP become one code path and the whole `socksFetch` shim can be deleted. Pin `undici@^7` for Node 20. Direct requests keep the global fetch, so nothing changes when no proxy is set.

**3. Accept `socks5h://` and `socks4a://`.** The validator and scheme check reject them, but remote DNS through the proxy is the main use case: the primary audience is behind DNS-poisoned networks and their clients (Clash etc.) expect socks5h semantics. fetch-socks does remote DNS by default, so after point 2 this is just a validator fix.

**4. Always bypass local targets.** `localhost`, `127.0.0.0/8`, RFC1918, `*.local` should never be proxied regardless of settings, or the `custom` provider (LM Studio, local Ollama) silently breaks when a proxy is enabled. Also worth seeding `zhipu` into the default bypass, since bigmodel.cn is faster direct for the users who need this most.

**5. Standard env vars.** #225 asked for `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY`. Suggested precedence: `PROXY_URL` > `HTTPS_PROXY`/`ALL_PROXY` > dashboard setting, honoring `NO_PROXY`. When an env var wins, the dashboard should say "managed by environment variable" rather than letting Save silently no-op.

**6. Fail fast when the proxy itself is down.** Treating every `fetch failed` as retryable means a dead proxy burns the full 20-attempt fallback chain per request. Classify "cannot reach the proxy" separately and return one clear error.

Smaller: the lockfile has unrelated `"peer": true` churn (please regenerate with the repo's npm), the explicit `{}` args in embeddings are unneeded, and this needs tests. Bypass logic and URL normalization are pure functions, and an in-process CONNECT proxy + `socksv5` server in the suite would let CI verify a streaming completion through both proxy types offline.

Nice-to-have, not blocking: a Test button that fetches through the proxy and shows latency plus exit-IP country. Gemini is region-locked server-side, so a working proxy with a HK exit still fails for Gemini, and surfacing that would prevent confused bug reports.

This rework deletes more code than it adds. If you'd rather not take it on, no problem, I'll build on your branch with credit to you and @homayounmohseni. Otherwise I'll re-review quickly.

tashfeenahmed:
Thank you @Slyker! Merged in #286 — I rebased this onto current main (it had drifted into conflicts + a tsc break), kept your design intact (single `proxyFetch` chokepoint in `BaseProvider.fetchWithTimeout`, so every provider is covered automatically), fixed one post-rebase client type error, and added a routing test. You're credited in the README contributors. 🙏

PR #239, opened 2026-06-06
fix: multiply monthly token budget by key count per platform

Slyker:
## What

Multiplies each model's monthly token budget by the number of enabled API keys for its platform in the `/api/fallback/token-usage` endpoint.

**Before**: a provider with 3 API keys showed the same budget as 1 key.
**After**: budget = per-key quota × key count, correctly reflecting total available capacity.

## Why

Each API key carries its own independent free-tier quota. The dashboard's stacked bar should reflect aggregate available budget, not per-key budget.

## Note

Optimal usage requires a proxy setup — some providers rate-limit by API key + IP, so multiple keys from the same IP may share a single rate-limit pool rather than being truly independent. The multiplication is correct for most providers (Google, OpenRouter, etc.) but the effective multiplier depends on deployment topology.

## Changes

- `server/src/routes/fallback.ts`: `/token-usage` now counts enabled keys per platform via `GROUP BY` and multiplies `parseBudget()` by key count


Slyker:
OUPSIE

PR #241, opened 2026-06-06
feat: rich tooltips on token budget bar segments

Slyker:
## What

Adds rich tooltips to the monthly token budget bar segments. Hovering any colored segment shows:

- **DisplayName**
- **platform**
- **X remaining** tokens (hidden if 0)
- **Y quota** (hidden if 0)

The gray "used" segment shows total consumed tokens.

## Why

The native `title` attribute doesn't render on 10px-high bar segments. The Tooltip component renders via `createPortal` to `document.body`, bypassing `overflow-hidden` clipping. The `className='h-full flex'` override replaces the default `inline-flex`, giving the wrapper span proper block-level sizing so it fills the parent div and mouse events fire correctly.

## Changes

- `client/src/pages/FallbackPage.tsx`: TokenUsageBar segments now wrapped in `<Tooltip className='h-full flex'>`

### DIFF
--- a/client/src/pages/FallbackPage.tsx
+++ b/client/src/pages/FallbackPage.tsx
@@ -274,21 +274,26 @@ function TokenUsageBar({ data }: { data: TokenUsageData }) {
 
       <div className="flex h-2.5 rounded-full overflow-hidden bg-muted">
         {modelsWithWidth.map((m, i) => (
-          <div
-            key={i}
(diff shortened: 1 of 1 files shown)


Slyker:
Should we only show active models in this section ?

tashfeenahmed:
Closing: this targets FallbackPage, which is deprecated on main (/fallback now redirects to /models/chat). The token usage bar lives on in other views, so if you want to port rich tooltips there, happy to review that.

PR #243, opened 2026-06-07
refactor(db): move migrations out of `initDb()`

LoneRifle:
## Problem
The `db` module is growing, due to active curation of free models, resulting in an increase in the number of migrations. This makes the module unwieldy, and any changes to the module itself in a fork may not be cleanly submitted back to the original repo in a pull request, due to the likelihood of merge conflicts.

### Impact
This may thus deter those who fork this repo to make changes (eg, migrating from SQLite to Cloudflare D1 or postgres) to contribute any improvements to the module back here.

## Solution
`initDb()` actually performs _two_ distinct tasks: 
- initialising the database client at runtime
- performing a series of schema migrations 

Move the latter to its own module, cleanly separating the two.


tashfeenahmed:
Merged, thank you! This was the right call: db/index.ts is the most-touched file in the repo right now and the split will save everyone merge pain.

Two things I fixed while resolving against main (the V23 free-tier audit landed after you branched):

1. Regenerated the split from current main so the V23 migration is included in migrations.ts.
2. Changed the import to `./migrations.js`. With `moduleResolution: bundler` tsc compiles the extensionless specifier fine, but the emitted ESM fails at runtime under `node dist/index.js` with ERR_MODULE_NOT_FOUND. Also added the missing `crypto` import that came along with `ensureUnifiedKey`.

Verified end to end: typecheck, 378/378 tests, production build, compiled server boots, and a fresh DB built from this branch is byte-identical to one built from main.

PR #244, opened 2026-06-07
docs: add Chinese README translation

Robs87:
## Problem

The project currently only has an English README, which creates a barrier for Chinese-speaking users.

## Solution

- Add `README_CN.md` with a complete Chinese translation
- Add language toggle links at the top of both `README.md` and `README_CN.md`
- Preserve all Markdown formatting, code blocks, links, and images
- Keep technical terms (API, SDK, endpoint, token, etc.) untranslated

## Why this approach

1. Large Chinese-speaking user base — translating the README significantly lowers the adoption barrier
2. No changes to existing format — doesn't affect other contributors
3. Language toggle links allow easy switching between English and Chinese

## Testing

- Verified all Markdown links navigate correctly
- Confirmed code block formatting is intact
- Checked table rendering is normal

---

Happy to maintain the Chinese translation as the upstream README evolves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+**[English](./README.md)** | **[中文](./README_CN.md)**
+
 <div align="center">
 
 # FreeLLMAPI
--- a/README_CN.md
+++ b/README_CN.md
@@ -1,0 +1,494 @@
+**[English](./README.md)** | **中文**
+
+<div align="center">
+
+# FreeLLMAPI
+
+**一个 OpenAI 兼容的 endpoint。十六个免费 LLM 供应商。每月约 17 亿 token。**
+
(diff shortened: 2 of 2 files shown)


tashfeenahmed:
I'd like to take this with two changes: (1) rebase onto current main, and (2) add a short banner at the top of README_CN.md noting the translation may lag the English README, with a link to it. Hand-maintained translations drift as the English file churns, so that expectation needs to be set inline. Once those are in, happy to merge.

tashfeenahmed:
Thanks for this, and sorry it sat so long. Superseded by #689, which is now merged.

Your PR still merged cleanly, which turned out to be the misleading part. `README.md` has taken 75 commits since 7 June, and diffing it against the version you translated shows 254 lines added and 334 removed out of about 588. More than half the page had changed underneath the translation. Merging it would have shipped a Chinese README describing a product that no longer matched, to exactly the readers least able to spot the difference. That is a worse outcome than having no translation, so rather than ask you to redo two months of drift, I re-translated against current `main` and built the structure your PR was missing.

What landed:

- `docs/i18n/<locale>/` mirroring the English tree, so a translated file's original is found by deleting the prefix from its path, and adding a language is copying a directory
- zh-CN covering the root README and the docs index in full, with the five long guides marked as English in a status table rather than shipped half-translated
- Terminology from `docs/translating.md`, the glossary that came out of the zh-CN dashboard review in #669, so the README and the app now say the same things

On that last point, your instinct to leave technical terms in English was reasonable in June, but the dashboard has since settled on `提供方` for provider and `词元` for model tokens against `令牌` for credentials. The README follows that now so the two do not contradict each other.

You are credited in the footer of the translated page and you were already in the Contributors list. If you want to keep maintaining zh-CN, the guides in `docs/i18n/README.md#status` are unclaimed and `install.md` is the one most people would read next. Happy to review.

PR #261, opened 2026-06-08
feat(routing): honor upstream Retry-After when benching a key

danscMax:
## What
When a provider returns 429/503 with a `Retry-After` header (delta-seconds or an HTTP-date), the router now honors it when setting that key's cooldown, instead of always using its own heuristic.

## How
- `parseRetryAfterMs()` + `providerHttpError()` in `providers/base.ts` capture the header (and status) and attach `retryAfterMs` to the thrown error; every adapter (`openai-compat`, `cohere`, `cloudflare`, `google`) now throws via that helper.
- `getCooldownDurationForLimit()` treats `Retry-After` as a **floor**: never shorter than the existing heuristic (so a daily quarantine isn't undercut), but extended — capped at a day — when the provider explicitly asks to wait longer. Strict improvement: it never causes an earlier retry.

## Tests
Unit tests for the parser (delta-seconds, HTTP-date, invalid) and the cooldown floor. Full server suite passes.

### DIFF
--- a/server/src/__tests__/services/ratelimit.test.ts
+++ b/server/src/__tests__/services/ratelimit.test.ts
@@ -12,6 +12,7 @@ import {
   providerDailyRequestCount,
   getProviderDailyRequestCap,
 } from '../../services/ratelimit.js';
+import { parseRetryAfterMs } from '../../providers/base.js';
 
 function removeDbFile(dbPath: string) {
   for (const suffix of ['', '-shm', '-wal']) {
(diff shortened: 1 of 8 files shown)


tashfeenahmed:
Integrated into #263 (cherry-picked, authorship preserved). Retry-After-as-a-floor behavior and the parser tests carried over intact. Thanks @danscMax — credited in the README. Closing in favor of the batch PR.

PR #262, opened 2026-06-08
feat: add Chinese language support (i18n)

3215:
- Add i18next + react-i18next for internationalization
- Create en.json and zh.json translation files
- Add language switcher in navbar (EN/ZH toggle)
- Translate all UI text across 8 pages: App, auth-gate, Keys, Playground, Fallback, Analytics, Embeddings, models-tabs
- Language preference persisted in localStorage


tashfeenahmed:
Superseded by #314 (a zero-dependency Context approach across five languages). Thanks for exploring this — credited you in the README Contributors. 🙏

PR #279, opened 2026-06-10
feat(proxy): add optional context handoff on model switch

immanuelsavio:
<html><head></head><body><h2>Context Handoff on Model Switch</h2><h3>The problem I was trying to solve</h3><p>I run long coding sessions through FreeLLMAPI - multi-turn conversations where an agent is working on a real task.</p><p>When a provider like Groq hits a rate limit mid-session and the router silently falls back to another model, such as Gemini or Cerebras, the fallback itself works, but the new model can lose task orientation.</p><p>In practice, that can look like the model re-introducing itself, asking questions that were already answered a few turns ago, or re-explaining things the user already told the previous model. The conversation technically continues, but it can feel like the task partially restarted.</p><p>What I wanted was a lightweight way for context to travel with the session, for the new model to know:</p><blockquote><p>“You are taking over an ongoing task. Continue from here.”</p></blockquote><p>A full memory layer with persistent storage, embeddings, cross-session recall, and retrieval is out of scope for a small self-hosted proxy. So this PR adds the smallest possible thing that still improves continuity.</p><hr><h3>What this PR adds</h3><p>This PR adds an optional, lightweight <strong>context handoff</strong> for auto-routed chat completion sessions.</p><p>When the router switches to a different model mid-session, FreeLLMAPI injects one compact system message into the outbound request, telling the new model that it is continuing an existing conversation.</p><p>That is all.</p><p>Example injected message:</p><pre><code class="language-text">FreeLLMAPI context handoff:
You are taking over an ongoing coding-agent conversation from another model (groq:llama-3.3-70b → google:gemini-flash-2.0).
Continue the user's task using the conversation context already provided in this request.
Do not restart the task, re-ask already answered setup questions, or discard prior tool results.
Respect the user's latest message as the highest-priority instruction.

Recent session summary:
User: can you refactor the auth middleware to use JWT...
Assistant: Sure, I'll start by reading the current middleware...
[...]
</code></pre><p>The feature is disabled by default.</p><p>Enable it with:</p><pre><code class="language-env">FREELLMAPI_CONTEXT_HANDOFF=on_model_switch
</code></pre><hr><h3>How it works</h3><p>Each auto-routed session gets an in-memory context slot, keyed by session ID or a SHA-1 hash of the first user message.</p><p>On every incoming request, FreeLLMAPI trims and stores up to 12 recent turns, capped at 500 characters per message.</p><p>After a successful response, it records which model handled that session.</p><p>On the next request, if the router selects a different model, FreeLLMAPI injects one handoff message after any existing system messages.</p><p>After the new model succeeds, it becomes the current model for that session, so no more handoff messages are injected until the model changes again.</p><p>No handoff is injected if the client pinned a specific model. This only applies to auto-routed sessions.</p><hr><h3>Storage behavior</h3><p>Storage is strictly in-memory:</p><ul><li><p>3-hour TTL per session</p></li><li><p>Hard cap at 500 session entries with LRU eviction</p></li><li><p>No disk writes</p></li><li><p>No database</p></li><li><p>No content logging</p></li><li><p>Nothing exposed through the dashboard or API</p></li></ul><hr><h3>Fresh-conversation guard</h3><p>Some agent harnesses reuse the same session ID for new tasks.</p><p>To avoid injecting stale handoff context into a brand-new conversation, this PR includes a fresh-conversation guard: if a session ID is reused but the first payload has no assistant turns, the stored model state is reset.</p><p>That prevents the handoff from firing at the start of a new task on a recycled session ID.</p><hr><h3>Routing overhead vs actual token overhead</h3><p>There are two different kinds of overhead here.</p><p>When the feature is enabled, the routing token estimate is padded by <code inline="">HANDOFF_MAX_TOKENS</code> on every auto-routed request. This is intentional routing headroom, not actual provider tokens. It means models that are already close to their context-window limit may be bypassed slightly more aggressively.</p><p>Actual tokens sent to the LLM only increase on turns where a handoff is injected, meaning when a model switch is detected for that session.</p><p>So the tradeoff is:</p><ul><li><p><strong>Every auto-routed request while enabled:</strong> conservative routing-side headroom padding</p></li><li><p><strong>Only model-switch requests:</strong> actual injected handoff message sent to the provider</p></li></ul><p>This avoids selecting a model that barely fits the original request but would fail once the handoff message is added.</p><hr><h3>Limitations</h3><p>This is not persistent memory.</p><p>The handoff message is a lossy orientation snapshot, not the full conversation. Long tool outputs and code blocks are truncated at 500 characters per message.</p><p>The new model should still receive the conversation history from the client in the request. The injected summary is only meant to orient the fallback model, not replace the actual message history.</p><p>This only applies to <code inline="">/v1/chat/completions</code> for now. Agents using the Responses API (<code inline="">/v1/responses</code>) do not get handoffs yet.</p><p>There is no cross-session or cross-restart memory. The in-memory store is cleared when the server restarts.</p><p>The routing token estimate is padded by roughly the handoff budget while the feature is enabled, so very tight context-window candidates may be skipped more often. Actual provider-token usage only increases when a handoff is injected.</p><hr><h3>Files changed</h3>
File | Change
-- | --
server/src/services/context-handoff.ts | New service for session context storage, handoff injection logic, TTL cleanup, and LRU eviction
server/src/routes/proxy.ts | Small integration to wire handoff into auto-routed chat completions
server/src/__tests__/services/context-handoff.test.ts | Unit tests for injection, TTL, truncation, reset behavior, pinned-model behavior, and disabled mode
.env.example | Documents FREELLMAPI_CONTEXT_HANDOFF
README.md | Adds Context Handoff section

<hr><h3>Test result</h3><p>All 422 tests pass.</p></body></html>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,6 +20,11 @@ PORT=3001
 REQUEST_ANALYTICS_RETENTION_DAYS=90
 REQUEST_ANALYTICS_MAX_ROWS=100000
 
+# Context handoff on model switch. When enabled, FreeLLMAPI injects a compact
+# system message into the outbound request whenever a session switches from one
+# model to another (e.g. after a fallback). Off by default.
+# FREELLMAPI_CONTEXT_HANDOFF=on_model_switch
+
 # Comma-separated extra origins allowed to call the API from a browser.
 # localhost:5173, 127.0.0.1:5173, [::1]:5173 are allowed by default for the
 # Vite dev dashboard. Only set this if you serve the dashboard from a
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Aggregate the free tiers from Google, Groq, Cerebras, NVIDIA, Mistral, OpenRoute
 - [Using the API](#using-the-api)
 - [Screenshots](#screenshots)
 - [How it works](#how-it-works)
+- [Context Handoff](#context-handoff)
 - [Limitations](#limitations)
 - [Contributing](#contributing)
 - [Terms of Service review](#terms-of-service-review)
@@ -87,6 +88,7 @@ Plus a **custom** provider — point at any OpenAI-compatible endpoint (llama.cp
 - **Health checks** — Periodic probes mark keys as `healthy`, `rate_limited`, `invalid`, or `error` so the router skips dead ones automatically.
 - **Admin dashboard** — React + Vite UI to manage keys, reorder the fallback chain, inspect analytics, and run prompts in a playground. Dark mode included.
 - **Analytics** — Per-request logging with latency, token counts, success rate, and per-provider breakdowns.
+- **Context handoff on model switch** — Optional. When a session falls over to a different model, injects one compact system message so the new model knows it is continuing an existing task. Disabled by default; enable with `FREELLMAPI_CONTEXT_HANDOFF=on_model_switch`. See [Context Handoff](#context-handoff).
 - **Runs anywhere Node 20+ runs** — Windows, macOS, Linux servers, or a small ARM SBC (Raspberry Pi included). ~40 MB RSS at idle behind PM2 / systemd / whatever supervisor you prefer.
 
 ## Not yet supported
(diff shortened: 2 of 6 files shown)


immanuelsavio:
Hey @tashfeenahmed this was a feature I wanted for my use but thought it could be helpful for others as well. Let me know if this is good to merge

PR #312, opened 2026-06-13
feat(client): add Simplified Chinese dashboard localization

XJPeng12:
## Summary

- Add `i18next` + `react-i18next` to the client with English and Simplified Chinese locale files.
- Add a navbar language toggle (EN / 中文) that persists the user's choice in `localStorage`.
- Translate dashboard UI strings across auth, models, keys, playground, analytics, and premium pages.
- Auto-select `zh-CN` when the browser language is Chinese on first visit.

## Motivation

The dashboard was English-only. This makes the admin UI usable for Chinese-speaking users without relying on browser translation.

## Notes

- Dynamic content from the API (model names, provider labels, error messages) is unchanged.
- No server-side changes.

## Test plan

- [ ] Open the dashboard and switch language via the EN/中文 toggle in the navbar.
- [ ] Verify login/setup, keys, models (chat + embeddings), playground, analytics, and premium pages in both locales.
- [ ] Refresh the page and confirm the selected language is remembered.
- [ ] Run `npm run build -w client` and confirm the production build succeeds.


XJPeng12:
#314

PR #319, opened 2026-06-15
fix(i18n): improve zh-CN translation quality

3215:
- Improved translation accuracy and naturalness
- Fixed punctuation consistency (using full-width punctuation)
- Optimized technical term translation
- Simplified complex sentences
- Updated playground input placeholder for clarity


tashfeenahmed:
Thanks for this, @3215! The zh-CN wording improvements are genuinely nice. A few things need addressing before this can merge.

**Please scope this down to zh-CN only.** The PR is titled as a translation fix, but it also includes some unrelated changes that need to come out:

1. **`server/src/providers/google.ts`** changes the API base from `/v1beta` to `/v1beta/openai`. This one is breaking. The rest of that file builds native Gemini requests (`:generateContent`, `:streamGenerateContent?alt=sse`, `?key=`, with `contents`/`generationConfig`/`systemInstruction` bodies). Adding `/openai` makes those URLs invalid, so it breaks the Google provider entirely. It looks like it may have slipped in from a rebase. Could you revert this line? If you do want to move Google to the OpenAI-compatible endpoint, that is a separate change that also needs the request/response format rewritten, so it should be its own PR.

2. **`package-lock.json`** and **`desktop/package-lock.json`** have unrelated lockfile churn (the `"peer": true` removals and a version bump). Please revert both so the diff stays focused on the translation.

**On the translation itself:**

3. `savingsHint` lost its meaning. The English source says the number "projects your pace from the last `{span}` of data to a full 30 days," and the new version drops both the `{span}` variable and that projection note. That caveat is the only thing separating `savingsHint` (a projected estimate) from `savingsHintExact` (the real total) right below it, so please keep it. Something like restoring the "based on `{span}` of data, projected to 30 days" clause.

4. The PR mentions fixing punctuation consistency, but only `smartestBlurb` was converted to full-width punctuation. Its siblings (`balancedBlurb`, `fastestBlurb`, `reliableBlurb`, `manualBlurb`, `customBlurb`) still use half-width commas, so right now they are inconsistent with each other. Could you either convert all of them to full-width or leave them as-is for a follow-up?

Everything else (`toolsTitle`, `addCustomDescription`, `inputPlaceholder`, `upsellDescription`, the `smartestBlurb` rewording) looks great. Once it is a clean zh-CN-only diff with the `savingsHint` fix and CI passing, happy to merge. Thanks again for contributing!

tashfeenahmed:
Carried your good translation improvements forward into a clean, translation-only PR: #322. It keeps your wording changes (with credit to you) and restores the `{span}` projection clause in `savingsHint`. Thanks again @3215!

PR #321, opened 2026-06-15
Codex/fix key persistence restart

johan-droid:


tashfeenahmed:
Reviewed this against current `main` while processing the key/provider PR batch.

The restart-safe persistence part has already been superseded by the current `FREEAPI_DB_PATH`/SQLite persistence and backup flow on `main`, including docs and tests around persistent DB paths/key state. This branch is also much broader than the PR title now: it adds a remote Postgres/Neon secret mirror, package changes, app-shell/key-page refactors, security header changes, and several server route changes. It conflicts heavily with current `main` across the client, DB, keys/settings/health routes, package lock, and docs.

I am not merging it as-is. If the remote secret mirror is still wanted, it should be split into a fresh, focused PR against current `main`; the key-persistence bug itself appears already handled.

tashfeenahmed:
Thanks for this, and sorry it sat for a while. I went through the branch commit by commit today.

The core problem this PR set out to fix (keys becoming unusable after a server restart) is now solved on current main, which grew equivalent mechanisms while this branch aged:

- The auto-generated encryption key is persisted in the DB at first boot (`server/src/lib/crypto.ts`), so stored provider keys stay decryptable across restarts even when no `ENCRYPTION_KEY` is set.
- `FREEAPI_DB_PATH` covers the persistent-SQLite-path-from-env case (this PR's `DB_PATH`).
- `FREEAPI_ENV_PATH` covers explicit .env locations for embedders.

I verified this on main today: fresh DB, added a key via the dashboard API, restarted the server, and the key both lists and decrypts correctly.

That said, there is genuinely good material in here beyond the persistence fix, and I would love to see it come back as focused PRs against current main:

1. **remote-secrets backup service** — this is the interesting one for ephemeral-disk hosts (Render/Fly free tiers). It overlaps directly with #399; a standalone PR referencing that issue would be very welcome.
2. **Security hardening** (log redaction, SQLite runtime settings/permissions, response headers) — each of these would merge quickly as its own small PR.
3. **Mobile-first dashboard polish** — same.

Closing this one since the base fix is superseded and the branch mixes 14 commits that are hard to review as a unit. Your diagnosis of the restart problem was on point and I'll credit you in the README contributors section. Please do resubmit the pieces above!

Issue #308, opened 2026-06-13
[Docs] Add Android/Termux installation guide


PSYCH0patch:
Hi @tashfeenahmed! I successfully got FreeLLMAPI running on Android via Termux after working around a few platform-specific issues. I'm sharing my complete setup guide here in case it helps other mobile users, and in case you'd like to officially support Android. Happy to convert this into a PR if you're interested!

# 📱 FreeLLMAPI on Android (Termux) - Installation Guide

> **Issue Type:** Installation Guide / Platform Support  
> **Platform:** Android (via Termux)  
> **Tested on:** Android 10+, ARM64, 6GB RAM  
> **Node Version:** v20.x / v22.x (LTS)

This guide documents the complete process to install and run **FreeLLMAPI** on Android devices using Termux, including all workarounds for issues encountered during installation.

---

## 🐛 Issues Encountered & Workarounds

| # | Issue | Root Cause | Workaround |
|---|-------|------------|------------|
| 1 | `better-sqlite3` build fails with `Undefined variable android_ndk_path` | `node-gyp` lacks Android NDK config by default | Export `GYP_DEFINES="android_ndk_path=''"` |
| 2 | Native compilation errors | Node.js v26 (non-LTS) incompatibility | Use `nodejs-lts` package instead |
| 3 | npm scripts blocked | New npm `allowScripts` policy | Run `npm approve-scripts <package>` |
| 4 | Dashboard inaccessible from LAN/VPN | Vite & server bind to `localhost` only | Set `HOST=0.0.0.0` in `.env` and edit `vite.config.ts` |
| 5 | Server killed when screen off | Android battery optimization | Use `termux-wake-lock` |

---

## 📋 Prerequisites

- **Termux** installed from [F-Droid](https://f-droid.org/packages/com.termux/) ⚠️ *(Play Store version is outdated)*
- **Android 7.0+** with **ARM64** architecture
- **Minimum 4GB RAM** (6GB+ recommended)
- **~2GB free storage**
- Stable internet connection

---

### Step 1: Update Termux & Install Base Packages

```bash
pkg update -y && pkg upgrade -y && pkg install -y git python clang make build-essential binutils pkg-config nodejs-lts termux-api nano
```

### Step 2: Verify Node.js Version (Must be LTS v20.x or v22.x)

```bash
node -v && npm -v
```

> ⚠️ If you see `v26.x.x` or higher, run: `pkg uninstall nodejs -y && pkg install nodejs-lts -y`

[...]

PR #328, opened 2026-06-16
feat: image generation via Pollinations

PSYCH0patch:
Adds `POST /v1/images/generations` as the first image-generation endpoint, routed to the Pollinations anonymous tier (keyless, no card). Closes the "Image generation (`/v1/images/*`)" gap in the **Not yet supported** list.

## What this PR adds

- **`PollinationsImageProvider`** — new image-only adapter, platform `'pollinations-image'` (separate from the existing text `'pollinations'` so the chat catalog stays untouched).
- **`POST /v1/images/generations`** — OpenAI-compatible route, same auth gate (`extractApiToken` + `timingSafeStringEqual`) as `/v1/chat/completions`, Zod-validated request.
- **`BaseProvider.generateImage()` + `supportsImages()`** — opt-in extension points. Default throw / `false`, so every existing provider is unchanged.
- **`shared/types.ts`** — `ImageGenerationRequest` / `ImageGenerationResponse` mirroring the OpenAI wire format.
- **17 unit tests** matching the `cloudflare.test.ts` pattern (`vi.spyOn(global, 'fetch')`).
- **README** — bullet under Features, full **## Image Generation** section with curl example and upstream notes, removed from **Not yet supported**.

## Verified upstream behavior (live-probed 2026-06-16)

Every claim in the code and docs is from a real probe, not recalled:

- `GET https://image.pollinations.ai/models` returns `["sana"]` — only one model on the anonymous tier, so that's the honest default.
- Sequential requests: **5/5 returned 200 OK** at 1s intervals.
- Concurrent requests: **2/3 returned HTTP 402** "Payment Required" — anonymous tier is 1-concurrent per IP.
- Response header `x-auth-status: unauthenticated` confirms the request was served by the free path.
- `x-model-used: sana` confirms what actually served the request. Requests for `model=flux` return 200 but with no `x-model-used: flux` — likely silent fallback to sana, so we don't pretend flux works.

## Design decisions worth flagging for review

- **`n>1` is rejected with 400**, not looped server-side. The 1-concurrent limit would make fan-out unreliable (the second image would 402). OpenAI clients can loop multiple single-image calls themselves with their own pacing.
- **URL response_format is lazy** — it builds a deterministic URL locally and returns it without calling Pollinations. Saves the 1-concurrent slot for callers who only want the URL anyway. `b64_json` is eager (downloads + base64-encodes, 60s timeout for `sana` cold starts).
- **Upstream 402 is preserved on the HTTP response**, but mapped to OpenAI's `rate_limit_error` type at the route layer (`images.ts`) so existing OpenAI SDK retry logic handles it. The provider's error message says "concurrent request limit reached" so the cause is clear.
- **`validateKey()` probes `/models`**, not a generated image — keeps the periodic health check from burning a real concurrent slot.

## Scope

Intentionally tight per CONTRIBUTING:

- One provider (Pollinations image), one endpoint (`/v1/images/generations`), backend only.
- No dashboard UI changes, no DB catalog seeding (Pollinations has one model, hardcoded default is enough).
- Cloudflare image / HuggingFace image / model-based routing land in future PRs.

## Manual test

`curl -X POST http://localhost:3001/v1/images/generations -H "Authorization: Bearer $FREELLMAPI_KEY" -H "Content-Type: application/json" -d '{"prompt":"a red panda coding on a laptop, watercolor","size":"1024x1024"}'`

Tested locally:
- Default model resolves to `sana`
- `n=2` returns 400 with clear message
- `n=1` explicit returns 200
- `response_format: "b64_json"` returns JPEG bytes base64-encoded
- Wrong API key returns 401
- Missing prompt returns 400

## Test suite

- New: **17/17 pass** (`pollinations-image.test.ts`)
- Existing: **487/487 pass** — no regression
- The 2 pre-existing failures in `db/idempotency.test.ts` and `services/ratelimit.test.ts` are environment-specific (Termux has no `/tmp`). Confirmed identical failure on `main` — not caused by this PR. They pass on standard Linux/macOS CI.

## AI-assisted disclosure

This PR is AI-assisted. Per CONTRIBUTING, no special label needed — the bar is the same as any other PR. Every factual claim is probe-verified, every line in the diff is self-reviewed, tests exercise the real behavior (not a mock of the wrong shape).

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Aggregate the free tiers from Google, Groq, Cerebras, NVIDIA, Mistral, OpenRoute
 - [Desktop app](#desktop-app)
 - [Premium (live catalog)](#premium-live-catalog)
 - [Using the API](#using-the-api)
+- [Image Generation](#image-generation)
 - [Screenshots](#screenshots)
 - [How it works](#how-it-works)
 - [Context Handoff](#context-handoff)
@@ -87,6 +88,7 @@ Plus a **custom** provider — point at any OpenAI-compatible endpoint (llama.cp
 - **Responses API** — `POST /v1/responses` (the wire format current Codex CLI versions require) is implemented as a translating shim over the same router, with full streaming events and tool calls.
 - **Streaming and non-streaming** — Server-Sent Events for `stream: true`, JSON response otherwise. Every provider adapter implements both.
 - **Tool calling** — OpenAI-style `tools` / `tool_choice` requests are passed through, and assistant `tool_calls` + `tool` role follow-up messages round-trip across providers.
+- **Image generation** — `POST /v1/images/generations` via Pollinations (keyless, anonymous tier). Returns `url` or `b64_json` per the OpenAI spec. `n>1` is rejected — see [Image Generation](#image-generation).
 - **Embeddings** — `/v1/embeddings` with family-based routing: failover only ever happens between providers serving the *same* model (vectors from different models are incompatible), never across models. See [Embeddings](#embeddings).
 - **Automatic fallover** — If the chosen provider returns a 429, 5xx, or times out, the router skips it, puts the key on a short cooldown, and retries on the next model in your fallback chain (up to 20 attempts).
 - **Per-key rate tracking** — RPM, RPD, TPM, and TPD counters per `(platform, model, key)` so the router always picks a key that's under its caps.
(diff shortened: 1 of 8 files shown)


tashfeenahmed:
Closing: image generation landed on main via the media pipeline shortly after this was opened, sorry for the overlap. The probe-verified approach here was clean work. Your Android/Termux guide offer in #308 is still very welcome.

PR #342, opened 2026-06-19
quota observability

johan-droid:


tashfeenahmed:
Thanks for this! Closing in favor of #344, which has the same quota observability feature in a much tighter, focused diff. This one had picked up a lot of unrelated changes from the fork. Let's land it via #344.

PR #376, opened 2026-06-25
feat: add Alibaba DashScope provider with dynamic model sync

jeqele:
Fetch chat models from DashScope /v1/models and register them in the Default routing profile. Pin the dev API base URL to 127.0.0.1 and update Keys/API docs so external SDK clients match the Playground proxy.

### DIFF
--- a/client/src/components/api-usage.tsx
+++ b/client/src/components/api-usage.tsx
@@ -1,14 +1,7 @@
 import { CopyButton } from '@/components/copy-button'
 import { useI18n } from '@/i18n'
 
-// The /v1 base URL for ready-to-run snippets, derived the same way as the chat
-// model page + Keys page: the dev server port in DEV, the page origin in a
-// packaged/hosted build.
-export function apiBaseUrl(): string {
-  return import.meta.env.DEV
-    ? `http://${window.location.hostname}:${__SERVER_PORT__}/v1`
-    : `${window.location.origin}/v1`
-}
+export { apiBaseUrl, apiOrigin } from '@/lib/api-base-url'
 
 // A copy-able "ways to use the API" code block, matching the chat detail page's
 // snippet card so every modality's detail page looks the same.
--- a/client/src/i18n/locales/en.json
+++ b/client/src/i18n/locales/en.json
@@ -176,6 +176,7 @@
     "endpointEmbeddingsHint": "model: \"auto\" or a family from the Embeddings tab",
     "endpointMessages": "Messages",
     "endpointMessagesHint": "Anthropic-compatible (Claude)",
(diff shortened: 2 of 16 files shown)


tashfeenahmed:
The DashScope provider itself is welcome. Two things changed since this was opened: model data now comes from the hosted catalog (no DB-migration seeding of model rows), and provider registration moved around a bit. Could you re-cut with just the provider and key plumbing, letting the catalog own the model rows? The dynamic /v1/models sync idea is interesting on its own; let's discuss that separately in an issue.

tashfeenahmed:
Thanks for this, and sorry it sat so long. The engineering here is solid: the dynamic `/v1/models` sync with backfill into the Default profile is a genuinely good pattern, and the tests are real tests. I'm going to decline it, but not for the reason you might expect, so let me be specific.

**It isn't a duplication problem.** We do already carry `modelscope` (魔搭社区), and it would be easy to assume that covers this. It doesn't. ModelScope is the model hub at `api-inference.modelscope.cn`; you're adding Alibaba Cloud Model Studio at `dashscope-intl.aliyuncs.com`. Different products, different auth, different endpoint. Your addition is genuinely new, and it reaches users ModelScope can't, since ModelScope only works after binding to an Alibaba Cloud China-site account with real-name verification.

**The blocker is the free tier.** This catalog only lists tiers that keep working — the router models everything as a recurring monthly budget and renders it in the dashboard's token-budget bar. Model Studio's free allowance is a one-time new-user quota, valid 30 to 90 days from activation, after which [Alibaba's own docs](https://www.alibabacloud.com/help/en/model-studio/new-free-quota) say you're charged for continued use. There's no monthly renewal and no always-free model on the API side. The separate free developer API tier was retired earlier this year.

That matters more than it might sound. If this shipped, everyone who added a Model Studio key would see it counted as ongoing free capacity, and then start getting billing errors a quarter later with no signal about why. A catalog entry that quietly stops being free is the same class of bug as a wrong rate limit: it ships to everyone at once.

Worth saying plainly — you opened this in late June, and the tier had changed shortly before that. There was no way to tell from the console.

**What I'd like to keep.** `provider-model-sync.ts` and the per-key **Sync models** button aren't Alibaba-specific. Any provider with a catalog that moves could use them, and right now we hardcode model lists. If you're up for it, that's a PR I'd take on its own, and it's a better piece of work than the provider registration around it. Two notes if you do: the `apiBaseUrl` / 127.0.0.1 change is unrelated and should go separately, and `main` has moved under this branch — `KeysPage.tsx` is now split into `components/keys/*`, and `startHealthChecker` / `startCatalogSync` take a `scheduler` argument.

If you'd rather stay on Alibaba, the higher-value thread is #581. ModelScope's 2000 requests/day is genuinely recurring, so it's the Alibaba product that actually fits this catalog. It just needs per-model testing before rows can land.

Thanks again, and sorry to send you away with a no on your first PR here.

PR #383, opened 2026-06-26
Add DGrid Free Models Router provider

ethanclarkeio:
## Summary

Adds DGrid as an OpenAI-compatible provider and seeds the `dgridai/free` Free Models Router as a recurring-free model.

The DGrid Free Models Router provides a single free smart-routing endpoint over DGrid's current free model pool, with conservative default limits of 10 requests/minute and 100 requests/day. The README also documents the optional $5 lifetime top-up quota and briefly notes DGrid's broader 200+ model catalog.

## Changes

- Register DGrid with the OpenAI-compatible provider registry
- Add `dgrid` to platform types, key creation, and the Keys page provider list
- Seed `dgridai/free` with free-tier limits and fallback config
- Document DGrid Free Models Router in the README
- Add provider, key route, and migration coverage

## Verification

- `npm run build`
- `npm test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 # FreeLLMAPI
 
-**One OpenAI-compatible endpoint. Sixteen free LLM providers. ~1.7B tokens per month.**
+**One OpenAI-compatible endpoint. Eighteen free LLM providers. ~1.7B tokens per month.**
 
-Aggregate the free tiers from Google, Groq, Cerebras, NVIDIA, Mistral, OpenRouter, GitHub Models, Cohere, Cloudflare, HuggingFace, Z.ai (Zhipu), Ollama, Kilo, Pollinations, LLM7, OVH AI Endpoints, and OpenCode Zen — plus any custom OpenAI-compatible endpoint (llama.cpp, LM Studio, vLLM, local Ollama) — behind a single `/v1/chat/completions` endpoint. Keys are stored encrypted. A router picks the best available model for each request, falls over to the next provider when one is rate-limited, and tracks per-key usage so you stay under every free-tier cap.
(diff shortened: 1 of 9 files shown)


tashfeenahmed:
Thanks. The provider registration part is fine, but model rows can't ship as DB migrations anymore: model data is owned by the hosted catalog now, and this also currently conflicts on migrations. Could you re-cut with just the provider plumbing? Separately, before wiring a third-party aggregator in, I'd want some signal on DGrid's stability and terms; if you have links to their docs/ToS, drop them here.

tashfeenahmed:
Thanks for the provider work. One policy thing blocks this in its current shape: model and catalog rows must not ship as migrations. The catalog is delivered by the published catalog service, which is also where tier gating happens, so seeding those rows in a migration puts them outside that path and they end up diverging from what the service publishes. That's what the comment in migrations.ts is there to prevent, so it needs to stay as-is rather than be rewritten.

Happy to take the rest, though. If you drop the migration and the comment rewrite and keep the provider plumbing — the registry entry, PLATFORMS, the KeysPage changes, and the tests — I can merge it after a rebase on current main.

tashfeenahmed:
Thanks for the provider work here, and for putting the README notes together with it.

I am going to close this one, but not because the idea was wrong. Two things have not moved since June. The catalog policy still stands, so model rows cannot ship as DB migrations, and the file this PR edits (`server/src/db/migrations.ts`) no longer exists in main at all, so the change would need rebuilding against the current migrations directory regardless. The question about DGrid's stability and terms is also still open, and I would want an answer on that before routing traffic through a third party aggregator.

None of that is a no forever. If you still want DGrid in, a fresh PR with just the provider and key plumbing, no model rows, plus a link to their docs and terms, is very welcome and I will look at it quickly. Sorry it went quiet on your end for so long.

PR #386, opened 2026-06-26
feat(providers): add AgentRouter with per-token model discovery

thatssoheil:
## What

Adds **AgentRouter** ([agentrouter.org](https://agentrouter.org)) as a provider — a free, no-card OpenAI-compatible gateway that fronts Claude, GPT, GLM and more. Two things make it a bit different from the existing providers, and both are handled here:

### 1. It gates on the client User-Agent

AgentRouter's "New API" backend only accepts a **Claude Code-style `User-Agent`** (`claude-cli/<version> (external…)`). Any other client — the OpenAI SDK, `curl`, a browser, or FreeLLMAPI's generic adapter — gets `401 "unauthorized client detected"` on **both** `GET /v1/models` (so a key would show as *invalid*) and `POST /v1/chat/completions` (so chat would 502).

The provider pins that User-Agent via `extraHeaders` — the **same mechanism Routeway already uses** to get past its Cloudflare UA filter:

```ts
register(new OpenAICompatProvider({
  platform: 'agentrouter',
  name: 'AgentRouter',
  baseUrl: 'https://agentrouter.org/v1',
  extraHeaders: { 'User-Agent': 'claude-cli/1.0.0 (external, cli)' },
}));
```

### 2. Models are per-token, not a fixed catalog

An AgentRouter token is scoped to a **user-chosen set of models** in AgentRouter's own panel, and `GET /v1/models` returns exactly that set. So instead of seeding a catalog, the Keys page **discovers** the token's models and lets the user pick:

- New `OpenAICompatProvider.listModels(apiKey)` reads the token's `/v1/models` (sending the same auth + UA).
- `POST /api/keys/agentrouter/discover` returns those IDs; the dashboard auto-fills them into an editable list so the user can trim before adding.
- `POST /api/keys/agentrouter` registers the chosen IDs **bound to the key** (`models.key_id`) and appends them to the fallback chain — reusing the same per-key registration + delete-cascade the custom-endpoint flow uses. Re-submitting replaces the set so edits take effect.

The shared bits (`collectModelEntries`, `registerModelEntries`) are extracted from the existing `/custom` handler so both flows stay in sync.

## Why

It's a genuinely free, no-card gateway to frontier models (Claude/GPT/GLM), which fits the project's free-tier aggregation goal. The only blocker to using it through FreeLLMAPI was the UA gate; this clears it the way the codebase already clears Routeway's.

## Files

- `shared/types.ts` — `agentrouter` added to `Platform`.
- `server/src/providers/index.ts` — provider registration (UA pinned).
- `server/src/providers/openai-compat.ts` — `listModels()` discovery.
- `server/src/routes/keys.ts` — `/agentrouter/discover` + `/agentrouter` routes; shared registration helpers; delete-cascade generalized to `agentrouter`.
- `server/src/services/provider-quota.ts` — quota pool wiring.
- `client/src/pages/KeysPage.tsx` + `FallbackPage.tsx` + `i18n/locales/en.json` — "Add AgentRouter" section (paste key → **Fetch models** → edit → add).
- `README.md` — provider list + ToS table (UA requirement disclosed).
- Tests: `__tests__/routes/agentrouter.test.ts` (discover / add / replace / cascade) and additions to `__tests__/providers/openai-compat.test.ts` (pinned UA on outgoing calls, `listModels` parsing).

## Verification

- `npm test` — **599 passing** (incl. the new tests; the existing 20 custom-provider tests still pass, guarding the shared-helper refactor).
- `npm run build` — green (server `tsc` + client `vite`).
- Live-probed against a real AgentRouter token (per CONTRIBUTING's no-invented-facts rule): the generic UA → `401 "unauthorized client detected"`; the `claude-cli` UA → authenticates, `/v1/models` returns the token's scoped set, and chat completes.

## Note on Terms of Service

AgentRouter is aimed at CLI coding tools and enforces the `claude-cli` User-Agent as its client gate; this PR sends that UA so a self-hosted, single-user proxy of your own token works. Flagged in the README ToS table as ⚠️ Caution so users can make their own call.


thatssoheil:
Closing — withdrawing this change for now. Thanks!

PR #393, opened 2026-06-27
feat(embeddings): accept optional dimensions parameter for MRL truncation

jasnoorgill:
## Summary

OpenAI-compatible `/v1/embeddings` accepts an optional `dimensions` integer that is forwarded to providers supporting **Matryoshka Representation Learning (MRL)**. On MRL-aware backends (NVIDIA NeMo NIM, Google Gemini Embedding, OpenAI `text-embedding-3-*`, OpenRouter) the model truncates its vector to the requested size **server-side**. Providers that ignore unknown fields (HuggingFace, Cloudflare) see byte-identical request bodies when the parameter is absent.

## Why this is needed

The local Hindsight memory server uses embedded **pgvector**, which hard-caps column dimensions at **2000**. NVIDIA's `llama-nemotron-embed-1b-v2` emits a native **2048-dim** vector — over the limit, rejected at INSERT time.

Two workarounds exist today:

1. **Switch to BGE-M3** (1024 dims via Cloudflare/HuggingFace) — fits, but lower retrieval quality.
2. **Truncate client-side** after the upstream call — works, but doubles request bytes and complicates batch APIs.

**MRL** is the third option that didn't exist until now: the model itself supports a smaller `output_dim` natively, with quality degraded gracefully (1536-dim Nemotron still beats 1024-dim BGE-M3 on retrieval benchmarks per the NVIDIA NIM docs). This PR enables it as a first-class request parameter.

## Usage

```bash
curl -X POST http://10.13.13.1:3001/v1/embeddings \
  -H "Authorization: Bearer <unified-key>" \
  -H "Content-Type: application/json" \
  -d '{"model":"nvidia/llama-nemotron-embed-1b-v2","input":"hello","dimensions":1536}'
```

Returns 1536-dim vector; same call without `dimensions` returns 2048.

## Compatibility matrix

| Provider | Supports `dimensions`? | Behavior when omitted | Behavior when set |
|----------|----------------------|---------------------|------------------|
| google (Gemini Embedding) | ✅ (MRL) | native dim | truncated |
| nvidia (NeMo NIM) | ✅ (MRL: 128/256/384/512/768/1024/1536/2048) | 2048 | truncated |
| openrouter | ✅ (passthrough) | native dim | forwarded to underlying provider |
| github (text-embedding-3-*) | ✅ (MRL) | native dim | truncated |
| cloudflare (BGE-M3 etc.) | ❌ ignores | 1024 | 1024 (param dropped) |
| huggingface (feature-extraction) | ❌ ignores | native | native (param dropped) |
| cohere (v2/embed) | ❌ different schema | native | native (param dropped) |

No provider rejects the field with a 400 — verified via curl. The parameter is gated by Zod as `z.number().int().positive()`, so callers can't send `0`, negative, or fractional values.

## Files changed

- `server/src/routes/proxy.ts` (+6 / -1) — Zod schema accepts `dimensions`; route forwards it.
- `server/src/services/embeddings.ts` (+24 / -9) — `openAiStyleEmbed`/`callProvider`/`runEmbeddings` thread the optional param; body construction conditionally adds it.

## Test plan

- ✅ All **17 existing embeddings tests pass** (no behavior regression on the default path).
- ✅ Manual end-to-end verified on this host:
  - `dimensions: 1536` → 1536-dim vector returned
  - omitted → 2048-dim vector returned (native Nemotron dim)
  - `dimensions: 0` → 400 (Zod rejects)
  - `dimensions: 99999` → upstream 400 with clear message

## Why a separate PR, not bundled

This change is **additive** — every existing call site continues to work exactly as before. It does not touch the process-safety-net, the router, the fallback chain, or any rate-limit code. It is therefore reviewable independently of the larger crash-fix work (`#391` area) and can land without coordinating with in-flight refactors.

## Backwards compatibility

- Existing clients that don't send `dimensions` see byte-identical request bodies on the wire.
- Existing clients that DO send `dimensions` (if any) will start receiving truncated vectors. The OpenAI API has accepted this parameter for ~3 years (since `text-embedding-3-small` shipped in Jan 2024) so any client using it is already aware of the behavior.

### DIFF
--- a/server/src/__tests__/services/embeddings.test.ts
+++ b/server/src/__tests__/services/embeddings.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import type { MockedFunction } from 'vitest';
 import { initDb, getDb } from '../../db/index.js';
 import { encrypt } from '../../lib/crypto.js';
 import { resolveFamily, getDefaultFamily, runEmbeddings, EmbeddingsError } from '../../services/embeddings.js';
@@ -194,15 +195,24 @@ describe('embeddings service', () => {
     });
 
     describe('dimensions parameter (MRL truncation)', () => {
+      // Typed fetch mock so .mock.calls is properly indexed without `as any`.
+      // The cast on assignment to globalThis.fetch is unavoidable because
+      // globalThis.fetch in lib.dom is typed loosely; this is the same
+      // pattern the existing tests in this file use.
+      function mockFetch(impl: typeof fetch): MockedFunction<typeof fetch> {
+        const m = vi.fn(impl) as MockedFunction<typeof fetch>;
+        globalThis.fetch = m as unknown as typeof fetch;
+        return m;
+      }
+
       it('forwards dimensions to NVIDIA NeMo NIM in the request body', async () => {
         addKey('nvidia');
-        const fetchMock = vi.fn(async () => okEmbeddingResponse(1536));
-        globalThis.fetch = fetchMock as any;
+        const fetchMock = mockFetch(async () => okEmbeddingResponse(1536));
 
         await runEmbeddings('llama-nemotron-embed-vl-1b-v2', ['hello'], 1536);
 
         expect(fetchMock).toHaveBeenCalledTimes(1);
-        const body = JSON.parse(fetchMock.mock.calls[0][1].body);
+        const body = JSON.parse(String((fetchMock.mock.calls[0][1] as RequestInit).body));
         expect(body.dimensions).toBe(1536);
         // The provider-prefixed model id (nvidia/<id>) is what reaches the upstream;
         // the bare id is used internally for family resolution only.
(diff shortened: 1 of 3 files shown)


jasnoorgill:
Added test coverage in a follow-up commit (`2e2177e`) per CONTRIBUTING.md's "Every PR should include a test" guidance.

**8 new tests** under `server/src/__tests__/services/embeddings.test.ts` → `describe('dimensions parameter (MRL truncation)')`:

1. Forwards `dimensions` to NVIDIA NeMo NIM in the request body (pinned wire format)
2. Omits `dimensions` from the upstream body when not requested (no-op default)
3. Returns the truncated dimension reported by the upstream response
4. Forwards `dimensions` to the Google provider
5. Forwards `dimensions` to the GitHub provider
6. Forwards `dimensions` to the OpenRouter provider
7. **Preserves byte-identical upstream bodies when `dimensions` is undefined** — snapshot comparison against the pre-feature code path
8. Passes `dimensions` through the family failover chain (nvidia → openrouter both see it)

All 25 tests in the file pass. Diff is now +149/-10 across 3 files (was +32/-10 across 2 files before the tests).

The PR description's "Test plan" section has been substantively met — the manual curl checks are now backed by deterministic CI tests.

jasnoorgill:
Follow-up commit (`6725871`): type-cleanup for the new tests.

Replaced the loose `vi.fn()` + `as any` mock pattern with a small `mockFetch()` helper that returns `MockedFunction<typeof fetch>`, and added `as RequestInit` casts on body-access lines (same pattern the existing tests in this file already use at line 143).

Verification:
- `npx tsc --noEmit server/src/__tests__/services/embeddings.test.ts` — **0 errors in my new tests (lines 200+)**; the 5 remaining errors are pre-existing in the original 17 tests (lines 95–156) and out of scope for this PR.
- `npm run build:server` (the actual CI build) — clean, exits 0.
- All 25 tests in the file pass at runtime.
- No behavior change — pure type cleanup.

The new tests now match the file's existing typing style while being strictly more correct.

PR #397, opened 2026-06-27
refactor(db): introduce Db/DbStatement/DbFactory interfaces

LoneRifle:
## Motivation

`better-sqlite3` is a native Node.js module. Any code that imports
`Database` from it at the type level is directly coupled to that
concrete dependency, which makes unit-testing hard: you either pull in
the native binary or you mock the entire module.

This PR introduces a thin structural interface (`Db` / `DbStatement` /
`DbFactory`) that the rest of the server code depends on instead.
The native module is confined to a single `betterSqliteFactory` function
inside `db/index.ts`.

## Changes

- **`server/src/db/types.ts`** (new) — `Db`, `DbStatement`, `DbFactory`
  interfaces. `better-sqlite3` is structurally compatible with `Db` via a
  single `as unknown as Db` cast in the factory function.

- **`server/src/db/index.ts`** — `connectDb` and `initDb` accept an
  optional `{ factory?: DbFactory }` option. Tests (and future alternative
  drivers) can inject a fake or in-memory implementation without touching
  the filesystem.

- All other server modules that previously held `import type Database from
  'better-sqlite3'` or `import type { Database } from 'better-sqlite3'`
  now import from `../db/types.js` instead:
  `db/migrate/runner.ts`, `db/migrate/cli.ts`, `db/migrate/defaults.ts`,
  `db/migrate/TEMPLATE.ts`, `db/migrations/…legacy_baseline.ts`,
  `db/model-pricing.ts`, `lib/crypto.ts`, `services/router.ts`,
  `services/catalog-sync.ts`, `services/quirks.ts`.

## Tests

`server/src/__tests__/db/factory-injection.test.ts` (new) — two fast unit
tests confirm that:

1. The injected factory receives the resolved path.
2. No filesystem operations occur when a stub factory is supplied — the
   test passes with a path that does not exist.

All 619 existing tests continue to pass.


LoneRifle:
Hi @tashfeenahmed! Would you be keen to have a look at this please? There should be no behavourial change, and resulting test code coverage should go up.

PR #410, opened 2026-06-28
feat(analytics): hourly aggregates + lifetime/30d counters survive raw-row prune

jasnoorgill:
## Problem

The `requests` table is silently capped by `REQUEST_ANALYTICS_MAX_ROWS` (default 100k). Combined with the 90d `REQUEST_ANALYTICS_RETENTION_DAYS` window, this caused the UI's 30d/7d/24h totals to hover at the cap regardless of actual traffic — visible in the DB as exactly `maxRows+N` rows and oldest rows ~8d old despite the 90d retention window.

For example, a deploy with sustained >100k requests/30d would show a flat 100k line in the analytics dashboard indefinitely.

## Fix

Add two durable aggregates that the analytics summary reads from instead of the pruned raw table:

- **`request_hourly`** — one row per UTC hour with counts + tokens, pruned at >30d. Covers 24h / 7d / 30d UI windows from ~720 rows.
- **`settings`** — lifetime `total_requests` / `total_input_tokens` / `total_output_tokens` / `first_request_at`, never pruned.

Both updated in the **same transaction** as the raw `INSERT INTO requests` in `lib/request-log.ts`, so the aggregates never disagree with the row count.

The hourly prune piggybacks on the existing 60s prune loop, gated once per 24h — no extra scheduler.

## Changes

- `server/src/db/migrations/20260628_120000_request_aggregates.ts` (new) — creates `request_hourly` and seeds lifetime counters from any still-present raw rows
- `server/src/db/migrate/defaults.ts` — register new migration
- `server/src/lib/request-log.ts` — atomic upserts in same tx as raw insert
- `server/src/services/request-retention.ts` — hourly prune (gated once/24h) + table-existence guard
- `server/src/routes/analytics.ts` — `/summary` and `/timeline` now read from aggregates; `/summary` exposes new `lifetimeTotalRequests` field
- Tests:
  - `server/src/__tests__/services/request-retention.test.ts` — hourly prune + 24h gate
  - `server/src/__tests__/routes/analytics.test.ts` — updated test helpers to mirror production aggregates
  - `server/src/__tests__/db/migrate/roundtrip.test.ts` — include new migration in expectations

## Compatibility

- Fresh installs work correctly: empty `request_hourly`, lifetime counters seeded from `COUNT(*) = 0`. First request populates them.
- Existing installs: backfilled from surviving raw rows on migration. Note that data already pruned before this migration is unrecoverable from the aggregate (raw rows are gone) — the lifetime counters are best-effort and start from the migration-time `COUNT(*)`.
- UI: no frontend changes required. `AnalyticsPage.tsx` reads `summary.totalRequests` and `summary.firstRequestAt` directly, which now reflect accurate values past the cap.
- New `lifetimeTotalRequests` field in `/summary` is additive; clients that don't read it are unaffected.

## Tests

All 671 tests pass (`npm test`). Build clean (`npm run build`).

## Verification on live install

After deploy + migration on the running instance:
- Raw `SELECT COUNT(*) FROM requests WHERE created_at >= now-30d` → 100000 (capped, as before)
- `SELECT SUM(total_requests) FROM request_hourly WHERE hour >= now-30d` → 100002+ (climbing)
- `SELECT value FROM settings WHERE key='total_requests'` → 100002+ (climbing)
- Both aggregates incrementing in real time, decoupled from the raw-row prune.

### DIFF
--- a/server/src/__tests__/db/migrate/roundtrip.test.ts
+++ b/server/src/__tests__/db/migrate/roundtrip.test.ts
@@ -7,6 +7,7 @@ import { up as runLegacyBaseline } from '../../../db/migrations/20260101_000000_
 const LEGACY_BASELINE_FILENAME = '20260101_000000_legacy_baseline.ts';
 const CUSTOM_PROVIDER_MODALITIES_FILENAME = '20260627_000001_custom_provider_modalities.ts';
 const CATALOG_MODEL_STATE_FILENAME = '20260627_000002_catalog_model_state.ts';
+const REQUEST_AGGREGATES_FILENAME = '20260628_120000_request_aggregates.ts';
 
 interface SchemaRow {
   type: string;
@@ -60,6 +61,7 @@ describe('migration round trip', () => {
         LEGACY_BASELINE_FILENAME,
         CUSTOM_PROVIDER_MODALITIES_FILENAME,
         CATALOG_MODEL_STATE_FILENAME,
+        REQUEST_AGGREGATES_FILENAME,
       ]);
     } finally {
       db.close();
--- a/server/src/__tests__/routes/analytics.test.ts
+++ b/server/src/__tests__/routes/analytics.test.ts
@@ -26,6 +26,7 @@ function insertRequest(createdAt: string) {
     INSERT INTO requests (platform, model_id, status, input_tokens, output_tokens, latency_ms, error, created_at)
     VALUES ('test', 'test-model', 'success', 1, 2, 3, NULL, ?)
   `).run(createdAt);
+  upsertAggregate(db, createdAt, 'success', 1, 2);
 }
 
 function insertTokensRequest(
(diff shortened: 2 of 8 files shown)


tashfeenahmed:
Thanks @jasnoorgill, the hourly-aggregate design here is exactly right, and surviving the raw-row prune is a real fix. Before merging I scoped the branch down to just the analytics feature and fixed one correctness bug. Summary of what changed on the branch:

**Rescoped to analytics-only.** The branch had picked up four unrelated commits from your integration branch: the `fix(proxy)` AbortError work (that's PR #409), the `chore(health)` log line (that's PR #394), and the `INTEGRATION_BRANCH.md` / `update-integration-branch.sh` tooling. I dropped those from this PR so it's a single concern. #409 and #394 stay as their own PRs (I merged #394 separately); nothing is lost.

**Fixed a boundary-day undercount.** The writer (`logRequest.hourKey` plus the migration backfill) stores bucket keys in SQLite's native `'YYYY-MM-DD HH:00:00'` text (space separator, same as `created_at`), but `readAggregateSince` and the retention prune compared against a `.replace(' ', 'T')` cutoff. Since `' '` (0x20) sorts before `'T'` (0x54), every bucket on the window's boundary day was excluded from the summary and over-pruned. For the 24h range that silently dropped everything before midnight. The unit tests missed it because the seed helpers also used the `'T'` form, so writer and reader moved together. Fix: compare on the native space format everywhere (the timeline reader already did), make the seed helpers mirror the real writer, and add a regression test that drives the actual `logRequest` writer and pins the stored key shape.

Full suite green (681 tests). Merging once CI passes here.

PR #419, opened 2026-06-29
feat(custom): multi-key rotation pool per endpoint

danscMax:
## What

Lets a single custom (OpenAI-compatible) endpoint rotate across **several interchangeable API keys**, instead of being pinned to one.

## Why

The router already round-robins across a platform's keys, but a custom model was pinned to a single key by an exact `key_id` match (#212). For self-hosted / proxied OpenAI-compatible endpoints that hand out multiple keys (rate-limit headroom, multiple accounts behind one base_url), there was no way to spread load across them.

## How

- **Routing** (`router.ts`): a custom model now matches keys by its endpoint's `base_url` rather than the exact `key_id`. Every key sharing that `base_url` joins the round-robin (cooldowns / rpm / rpd / tpm / tpd limits still applied per key); keys belonging to *other* custom endpoints stay excluded, so the #212 isolation is preserved. If the endpoint's `base_url` can't be resolved it falls back to the legacy exact-key match; legacy rows (`key_id NULL`) keep the old any-key behaviour.
- **Registration** (`keys.ts`): `POST /api/keys/custom` gains an optional `extraKeys: string[]`. The primary key still backs the model bindings; the extras form the rotation pool. Extras are trimmed, de-duped, and never a copy of the primary. Re-submitting an endpoint rebuilds its pool (primary row and its model bindings preserved). The response returns `poolKeys` = primary + extras.

## Tests

New isolated test covering pool build, dedupe (duplicate / blank / self-copy dropped), primary preservation, and pool rebuild on re-submit. Full suite green (685 passed).

### DIFF
--- a/server/src/__tests__/routes/custom-provider.test.ts
+++ b/server/src/__tests__/routes/custom-provider.test.ts
@@ -358,3 +358,40 @@ describe('Custom Provider Endpoints', () => {
     });
   });
 });
+
+// Isolated DB so the global custom-key counts in the block above stay intact.
+describe('custom provider key rotation pool', () => {
+  let app: Express;
+  beforeAll(() => {
+    process.env.ENCRYPTION_KEY = '0'.repeat(64);
+    initDb(':memory:');
+    getDb().prepare("DELETE FROM settings WHERE key = 'active_profile_id'").run();
+    app = createApp();
+    dashToken = mintDashboardToken();
+  });
+
+  it('builds a pool from extraKeys and replaces it on re-submit (primary preserved)', async () => {
+    const base = 'http://127.0.0.1:9991/v1';
+    const r1 = await post(app, '/api/keys/custom', {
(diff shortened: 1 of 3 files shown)


tashfeenahmed:
Thanks for the careful work here, and sorry to be the bearer of bad news: I'm passing on this one as a matter of policy, not code quality (the implementation is tidy and the writeup was clear).

FreeLLMAPI's premise is stacking free tiers that each user is individually entitled to, for personal experimentation. Nearly every provider's terms are one account, one key, with limits attached to that account. A first-class rotation pool exists mostly to multiply those limits across several keys, and shipping it as a feature endorses that pattern, including against the very free-tier providers this project depends on staying in good standing with. That holds even scoped to custom endpoints, since a custom endpoint is very often just a provider's base URL.

For a genuinely legitimate multi-key setup (say, a self-hosted gateway that issues several equivalent keys), the existing escape hatch is to add the endpoint twice as two custom entries with different keys and let the fallback chain spread load across them.

Closing, but genuinely appreciated, and I'd be glad to review other contributions from you.

PR #431, opened 2026-07-01
Feat/auxiliary routing

zhaoxp-xyz:
for hermes model


tashfeenahmed:
Thanks for the ambitious work here. Blocking problems as it stands: (1) no migration creates the auxiliary_config table, so the /api/auxiliary routes fail at runtime on a fresh install; (2) debug leftovers are still in the diff (window.__REACT_ERRORS__ injection in main.tsx, .bak entries in gitignore); (3) only 2 of 6 locales are covered; (4) it now conflicts with main. Bigger picture: this adds a speculative parallel routing subsystem with 15 task types, which is a large design commitment. I'd rather start from a short design issue describing the two or three task types you actually use, and grow from there. Happy to discuss in an issue; leaving this open for a bit if you want to slim it down.

zhaoxp-xyz:
Thanks for the review! I've rebased against the latest main and addressed all the blocking issues:

**Fixed:**
- ✅ **Migration added** — `20260701_000000_create_auxiliary_config.ts` creates the `auxiliary_config` table (with index + FK to models)
- ✅ **Debug code removed** — `__REACT_ERRORS__` injection in main.tsx deleted
- ✅ **`.gitignore` cleaned** — all `*.bak` entries removed
- ✅ **Locales filled** — es, fr, it, pt-BR now have `visionTab`, `coderTab`, `groupsTab` keys matching en/zh-CN
- ✅ **Rebased against latest main** — zero conflicts, clean 4-commit history

**Open questions:**
On the design scope — I use auxiliary routing with the task types that correspond to my local model chain (vision, coder, webextract, compression, video, image generation). The 14 task types in VALID_TASK_TYPES map to capabilities supported by FreeLLMAPI's auxiliary chain system. I'm happy to discuss the right set in a design issue, but for now the API is runtime-dynamic — unused task types are inert until configured.

Let me know if anything else needs changing! 👀

tashfeenahmed:
First, credit where it is due. You came back and fixed all four blockers I raised, the migration, the debug leftovers, the gitignore entries and the locales, and you did it in a day. That was a good response to review and I appreciate it.

I am still closing this, and the reason is the one I flagged as the bigger picture rather than anything you did wrong. A parallel routing subsystem with fifteen task types is a large design commitment for the project to take on in one step, and I am not confident enough in the shape of it to maintain it long term. That is a judgement call on scope, not a comment on the quality of the code.

What I would genuinely like is a short issue describing the two or three task types you actually use in practice, and what routing behaviour you need for them. If the narrow version proves itself we can grow it from there, and a lot of this work would carry straight over. Please do open that issue, and thanks again for the effort you put in.

Issue #430, opened 2026-07-01
user and password for .exe setup


86TheCactus:
Where is my username and password for the windows .exe setup? I was never given a prompt to set it up and I cant figure out how to edit the .db using sql

tashfeenahmed:
Good news: there's no username or password to find. The Windows desktop app signs the dashboard in automatically with a hidden local account, so it never prompts you for credentials and you never need to touch `freeapi.db` with SQL.

The only credential you need is your **unified API key** (the `freellmapi-…` token you point your OpenAI/Anthropic client at). Grab it either way:

- **Tray popover** → click the FreeLLMAPI tray icon → **Copy Key**.
- **Dashboard** → tray → **Open Dashboard** → the key is in the **Keys** page header.

If your goal was to log into the dashboard specifically, just use tray → Open Dashboard; it opens already authenticated. Documented this in #448 so the next person isn't stuck.

Issue #189, opened 2026-06-03
Can not remove custom added


onlivee:
when i try to add custom key,then remove it.because i added a wrong one.
and i found it still in the fallback label.

and i try to test in playground label.using the right one custom model also report something wrong like 

Error: Provider error (xxxxx): Unexpected non-whitespace character after JSON at position 583 (line 27 column 2)

tashfeenahmed:
Both parts of this are fixed in #197, thanks for the clear report.

1. Removing a custom key now also removes its models from the fallback chain, so you will not see stale entries anymore.

2. That JSON error usually means the base URL points at an API that is not OpenAI compatible. If you are using Ollama, use http://host:11434/v1 rather than the native /api path. The error message now says this directly instead of showing the raw parse error.

Pull the latest main and you should be good. If the playground error persists with the /v1 path, please reopen with the base URL shape you are using.

onlivee:
> Both parts of this are fixed in [#197](https://github.com/tashfeenahmed/freellmapi/pull/197), thanks for the clear report.
> 
> 1. Removing a custom key now also removes its models from the fallback chain, so you will not see stale entries anymore.
> 2. That JSON error usually means the base URL points at an API that is not OpenAI compatible. If you are using Ollama, use http://host:11434/v1 rather than the native /api path. The error message now says this directly instead of showing the raw parse error.
> 
> Pull the latest main and you should be good. If the playground error persists with the /v1 path, please reopen with the base URL shape you are using.

ok,thank you!

PR #439, opened 2026-07-01
fix(provider): distinguish Cloudflare-truncated 200 from non-OpenAI-compat endpoint

jasnoorgill:
## Summary

Splits the catch-block error at `server/src/providers/openai-compat.ts:152` into two distinct messages so operators stop chasing a base-URL config bug that doesn't exist.

## Motivation

Live observation on `10.13.13.1:3001` (7-day window, Kilo platform only): **139 of 179 errors (78%)** were logged as `Kilo Gateway returned 200 with a non-JSON body — the endpoint is not OpenAI-compatible.`

Of those 139:
- 51 rows hit `latency_ms` ≈ **600,000** (Cloudflare edge idle-keepalive dropped the upstream socket mid-response)
- 59 rows sat in the 1–5 minute bucket (same pattern, slightly under the 600s ceiling)
- 89 rows were `poolside/laguna-m.1:free`, 50 rows were `nvidia/nemotron-3-super-120b:free` — both routed through Kilo's free-tier queue, both behind Cloudflare

The error message pointed operators at the Kilo base URL (`https://api.kilo.ai/v1`), which was **correct**. The actual cause was upstream truncation. Two distinct failure modes, one misleading message.

The truncation-vs-wrong-endpoint distinction is the difference between "rebuild / reconfigure your base URL" and "retry, or pick a different model/upstream."

## Changes

`server/src/providers/openai-compat.ts:149-189` — catch-block now classifies the parse error:

| SyntaxError pattern | Content-Type | New message |
|---|---|---|
| `Unexpected end of JSON input` | any | `… response body was truncated mid-stream (likely an idle-keepalive timeout at an upstream proxy or CDN, e.g. Cloudflare's 600s edge limit). Retry, or switch to a faster model/upstream.` |
| `Unexpected non-whitespace character after JSON at position <n>` | `application/json` (not NDJSON) | same truncation message |
| `Unexpected non-whitespace character after JSON at position <n>` | `application/x-ndjson`, `text/event-stream`, missing | unchanged: `… endpoint is not OpenAI-compatible` |
| anything else | any | unchanged: wrong endpoint |

The Content-Type check is the load-bearing piece. NDJSON bodies (Ollama's native `/api/chat`, etc.) produce the same SyntaxError text as a mid-stream truncation, but the upstream's Content-Type betrays which one it is.

## Test plan

- **3 new vitest cases** in `server/src/__tests__/providers/openai-compat.test.ts`:
  - `distinguishes a truncated 200 body (CDN keepalive) from a wrong-endpoint 200 body (#430)` — `Unexpected end of JSON input` + `application/json` → truncation message
  - `attributes an "after JSON at position" parse error with Content-Type: application/json to CDN truncation (#430)` — mid-position error + `application/json` → truncation message
  - `does not flag NDJSON Content-Type as a truncation (#430 regression guard)` (+ explicit `not.toThrow` companion test) — same SyntaxError text + `application/x-ndjson` → still wrong-endpoint message
- The existing `#189` test (`Unexpected non-whitespace character after JSON at position <n>` with no Content-Type) still passes — empty/missing Content-Type falls through to the safe default.

**Test count**: `origin/main` baseline = **479 passing**, this branch = **483 passing** (+4 net: 3 new tests + 1 already-existing that still passes). Zero regressions on the openai-compat suite.

(26 test FILES fail to load in this environment due to a missing `multer` dep — unrelated to this PR, same on `origin/main`.)

## Compatibility

- No new env vars, no new dependencies.
- No DB schema changes, no migration needed.
- Pure error-message change at the throw site. `isRetryableError` in `lib/error-classify.ts` matches on substring `aborted` / `fetch failed` / `upstream 4xx|5xx` — the new message contains none of those tokens, so retry behavior is unchanged from a substring-match perspective. The router will still fail over to the next key for these errors via the generic retryable classifier; if you want the retry policy tuned per cause, that's a follow-up.

## Live verification

Commit `3d90453` (same change as this PR) is deployed on the `integration/all-my-prs` branch and running on `10.13.13.1:3001` (systemd `freellmapi.service`, build `2026-07-01 13:34:16` BST). The next Kilo free-tier CF-truncated request will surface the new message in the `requests.error` column instead of the misleading base-URL hint.

## Issue

References operator session 2026-07-01 ("Kilo/Poolside non-OpenAI-compatible — troubleshoot similar errors"). No upstream issue number yet — happy to file one if requested.

🤖 Generated with [Hermes Agent](https://nousresearch.com)

### DIFF
--- a/server/src/__tests__/providers/openai-compat.test.ts
+++ b/server/src/__tests__/providers/openai-compat.test.ts
@@ -159,6 +159,89 @@ describe('OpenAICompatProvider', () => {
     ).rejects.toThrow(/not OpenAI-compatible/);
   });
 
+  it('distinguishes a truncated 200 body (CDN keepalive) from a wrong-endpoint 200 body (#430)', async () => {
+    // Cloudflare-fronted free-tier upstreams (Kilo routing Nemotron / Poolside Laguna)
+    // deliver a partial JSON body when the 600s edge idle-keepalive fires mid-response.
+    // Node's JSON.parse surfaces that as "Unexpected end of JSON input" — different from
+    // the NDJSON / native-API case above. Operators were chasing a base-URL bug that
+    // didn't exist because both paths produced the same error string.
+    const headers = new Headers({ 'content-type': 'application/json' });
+    vi.spyOn(global, 'fetch').mockResolvedValueOnce({
+      ok: true,
+      status: 200,
+      headers,
+      json: () => Promise.reject(new SyntaxError('Unexpected end of JSON input')),
+    } as any);
+
+    let caught: Error | null = null;
+    try {
+      await provider.chatCompletion('key', [{ role: 'user', content: 'hi' }], 'model');
+    } catch (err) {
+      caught = err as Error;
+    }
+    expect(caught).not.toBeNull();
+    expect(caught!.message).toMatch(/truncated mid-stream/);
+    expect(caught!.message).toMatch(/Cloudflare's 600s edge limit/);
+    expect(caught!.message).not.toMatch(/not OpenAI-compatible/);
+  });
+
+  it('attributes an "after JSON at position" parse error with Content-Type: application/json to CDN truncation (#430)', async () => {
+    // Same edge-keepalive case but the body parses far enough to reach a
+    // mid-token garbage character before EOF. Content-Type is the only signal
+    // that distinguishes this from a real NDJSON upstream (next test).
+    const headers = new Headers({ 'content-type': 'application/json' });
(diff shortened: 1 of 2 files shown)


tashfeenahmed:
Reviewed and merging — this is a clean split of two genuinely different failure modes, and the Content-Type guard against NDJSON misclassification is exactly right. Note: I removed the `(#430)` reference from the title since #430 in this repo is an unrelated Windows setup issue (the number probably came from your fork's numbering). Thanks for the detailed live data backing this up!

PR #452, opened 2026-07-03
lmao

iceyxsm:


tashfeenahmed:
Thanks for this — and sorry for the unlucky timing. Your rate-limit persistence commit is dated May 25; main fixed that exact bug (#88) on May 26 in 7cc751a, with counters now derived from a persisted `rate_limit_usage` event log and write-through cooldowns in `rate_limit_cooldowns`, so restarts no longer reset daily quota. The OpenCode Zen provider also landed separately a while back, and the specific free promo models here were retired in V25 after they went dead upstream; model data is now managed via the live catalog rather than repo seeds. That leaves nothing in this branch that isn't already on main, so I'm closing it. You clearly found a real bug independently, and PRs against current main are very welcome — the fork-feature backlog in the issues has several items if you're looking for one.

PR #458, opened 2026-07-03
fix(catalog-sync): add enabled to model overrides map to prevent undefined SQL column crash

cagedbird043:
### Summary

Fixes a crash during catalog sync re-application after boot (`cached catalog re-apply failed: no such column: undefined`).

### Root Cause

When a model has overrides stored in `model_overrides` (such as `{"enabled": false}`), `applyModelOverrides()` in `model-state.ts` maps each JSON key to a SQLite column using `OVERRIDE_COLUMNS`. Because `enabled` was missing from `OVERRIDE_COLUMNS`, `OVERRIDE_COLUMNS['enabled']` evaluated to `undefined`. This constructed a broken query (`UPDATE models SET undefined = ? ...`) that failed on boot.

### Changes

- Added `enabled?: boolean` to `ModelOverridePatch` and `enabled: 'enabled'` to `OVERRIDE_COLUMNS` in both `model-state.ts` and `declarative-config.ts`.
- Added boolean-to-integer conversion for `enabled` in `toDbValue()`.
- Added a defensive filter (`.filter(k => k in OVERRIDE_COLUMNS)`) in `applyModelOverrides()` to ensure unmapped keys in JSON never result in SQL column errors.

### DIFF
--- a/server/src/services/declarative-config.ts
+++ b/server/src/services/declarative-config.ts
@@ -302,6 +302,7 @@ function upsertModel(db: Database.Database, input: z.infer<typeof modelSchema>):
     contextWindow: 'context_window',
     supportsVision: 'supports_vision',
     supportsTools: 'supports_tools',
+    enabled: 'enabled',
   };
   for (const key of Object.keys(patch) as Array<keyof ModelOverridePatch>) {
     assignments.push(`${columnMap[key]} = ?`);
--- a/server/src/services/model-state.ts
+++ b/server/src/services/model-state.ts
@@ -15,6 +15,7 @@ export interface ModelOverridePatch {
   contextWindow?: number | null;
   supportsVision?: boolean;
   supportsTools?: boolean;
+  enabled?: boolean;
 }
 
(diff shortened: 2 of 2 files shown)


tashfeenahmed:
Thanks for digging into this. I tried to reproduce the crash on current main and could not: both writers of overrides_json build their patch from an explicit allowlist that excludes `enabled` (declarative-config.ts and routes/models.ts), and upsertModelOverrides filters again through OVERRIDE_COLUMNS, so an `enabled` key can't be persisted there on a stock install. I suspect your DB had rows written by an older build or a fork.

That said, the defensive filter in applyModelOverrides is worth keeping: filtering patch keys to known columns kills the whole class of 'SET undefined' crashes. Could you trim the PR to that filter plus a regression test (a stored override containing an unknown key must not crash catalog re-apply), and drop the `enabled` column-map addition? Adding `enabled` to OVERRIDE_COLUMNS would let a stored override silently force-flip models.enabled on every catalog sync, which we don't want.

PR #460, opened 2026-07-03
feat(embeddings): add quota summary and TokenUsageBar card to embeddings page

cagedbird043:
### Summary

Adds a quota summary and multi-account pooled budget card (`TokenUsageBar`) to the Embeddings page, aligning its UI design with the main Models page.

### Changes

- In `server/src/routes/embeddings.ts`, parse quota labels (e.g. Cloudflare Neurons/day, HuggingFace monthly credits) and scale by active key count to calculate `totalBudget`, `totalUsed`, and per-family budgets in `/api/embeddings/usage`.
- Export `TokenUsageBar` and `TokenUsageData` from `FallbackPage.tsx`.
- In `EmbeddingsPage.tsx`, render `<TokenUsageBar>` at the top of the Embeddings tab when total budget is available.

image

### DIFF
--- a/client/src/pages/EmbeddingsPage.tsx
+++ b/client/src/pages/EmbeddingsPage.tsx
@@ -2,6 +2,7 @@ import { useState } from 'react'
 import { Link } from 'react-router-dom'
 import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query'
 import { ArrowDown, ArrowUp } from 'lucide-react'
+import { TokenUsageBar } from './FallbackPage'
 import { apiFetch } from '@/lib/api'
 import { Button } from '@/components/ui/button'
 import { Switch } from '@/components/ui/switch'
@@ -36,7 +37,9 @@ interface EmbeddingsData {
 }
 
 interface UsageData {
-  families: { family: string; requestsToday: number; tokensMonth: number }[]
+  totalBudget?: number
+  totalUsed?: number
+  families: { family: string; displayName?: string; platform?: string; budget?: number; used?: number; requestsToday: number; tokensMonth: number }[]
 }
 
 function formatTokens(n: number): string {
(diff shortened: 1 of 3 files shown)


tashfeenahmed:
Thanks. Two things before this can go in: (1) parseEmbeddingQuotaLabel hardcodes equivalences (10K neurons/day approximated as 30M tokens/mo, $0.10 as 2M tokens) inferred from quota label strings, plus a silent cloudflare default; those numbers will drift and mislead. If we can't compute a real figure from recorded usage, it's better to show the quota label as-is. (2) TokenUsageBar is imported from FallbackPage into EmbeddingsPage; a shared component should move to components/ rather than being owned by a sibling page. Also note #459 and #457 merged, so a rebase picks up pooled scaling and key load-balancing. Happy to review a revision.

tashfeenahmed:
Shipped in #695 — thanks for this, and sorry it sat so long.

Your idea was the right one: the Embeddings page showed a quota label per provider but never totalled anything, and the Models page had had a summary card at the top for ages. #695 puts that card on **Embeddings, Image and Audio**, in the same slot and the same chrome as the chat tab's budget bar. Image and Audio had no usage reporting at all before, so your PR ended up improving two pages it never touched.

One thing changed on the way in, and it is worth explaining rather than just landing quietly. Your `parseEmbeddingQuotaLabel` bridged the missing budget with fixed conversion factors:

```js
return kNeurons * 100_000 * 30;          // 10K neurons/day ~ 30M tokens/month
return Math.round(dollars * 20_000_000); // $0.10/mo ~ 2M tokens
```

That is a completely reasonable thing to reach for — there is no budget column on `embedding_models` to divide by, so something has to fill the gap. But running it against the live catalog produces `92.0M remaining`, and Neurons-per-token varies by model while dollar credits are not tokens at all, so the figure is not derived from anything. Since it renders as free capacity people plan against, it is the same class of problem as a wrong rate limit in the catalog: wrong for everyone, silently.

I went looking for a better denominator and there isn't one yet. Of the 16 enabled embedding and media rows, exactly one (`100 rpm · 1K req/day`) has a cap the dashboard could honestly count against. Five are Cloudflare Neurons, which we do not log and which are pooled across every Cloudflare model including chat. Four are per-minute rates with no cumulative cap. Six are not quotas at all — `MP3 output - multilingual` is a format description.

So the card reports what is actually measured: month-to-date spend, how it splits across providers, requests today, and each provider's quota label printed verbatim for the reader to judge. A real budget bar on these tabs needs structured quota fields in the catalog (amount, unit, period) rather than a display string, which is a catalog change rather than a dashboard one. Worth doing, and your PR is what surfaced it.

You are credited as `Co-authored-by` on the squashed commit and you were already in the [Contributors list](https://github.com/tashfeenahmed/freellmapi#contributors). Closing this one as superseded.

PR #462, opened 2026-07-04
feat(analytics): persist selected time range in localStorage

jasnoorgill:
## Summary

Persists the selected analytics time range (24h / 7d / 30d) in `localStorage`
so the user lands back on the window they were inspecting after a reload or
new session. Same persistence pattern as the existing `theme` toggle
(`client/index.html`) and the `freellmapi.locale` preference
(`client/src/i18n/I18nProvider.tsx`).

Storage key: `freellmapi.analytics.range`. Values are validated against
`['24h', '7d', '30d']`; invalid or missing entries fall back to the existing
7d default so a corrupted entry never bricks the page. `localStorage`
access is wrapped in `try/catch` for SSR / private-mode safety (same
pattern as `client/src/lib/api.ts:18-26`).

## Motivation

The analytics page is a frequent revisit — most operators land on 7d or 30d,
not the default, and a hard reload (or new browser session) always snapped
them back to 7d. Small but persistent annoyance; the existing `theme` and
`freellmapi.locale` patterns show the project already considers
client-level preferences worth persisting, so this is consistent with the
project's UX conventions.

## Changes

| File | Change |
|---|---|
| `client/src/pages/AnalyticsPage.tsx` | Added `useEffect` import; `useState<TimeRange>('7d')` → `useState<TimeRange>(loadStoredRange)`; new `useEffect` to persist `range` on change; added `STORAGE_KEY`, `TIME_RANGES`, `DEFAULT_RANGE` constants and `loadStoredRange()` helper. |

**1 file, +34 / -2.** No new dependencies. No server changes. No new
migrations. No i18n changes (the tab labels use existing
`analytics.range24h` / `range7d` / `range30d` keys).

## Test plan

- `npx tsc -b` in `client/` — clean, 0 errors.
- `npx tsc -b` in `server/` — 9 errors in `server/src/routes/keys.ts` from
  upstream's `multer` typing bug (PR #438, pre-existing on `origin/main`
  at `340f414`). **Not introduced by this PR** — same errors reproduce on
  a clean `git checkout origin/main`.
- `npx vitest run` in `server/` — 524/524 passing, 0 failures, 0 regressions.
- Live verification: deployed on `integration/all-my-prs @ f81f781`; the
  rebuilt client bundle (`dist/assets/index-DJVe5VdN.js`) contains the new
  storage key string and the read+set localStorage calls. Restarted
  service, `MainPID=860700` running.

Manual smoke:
1. Open `/analytics`, click 30d.
[...]

Issue #438, opened 2026-07-01
I added Agnes AI, but it doesn’t appear in the model list.


gongjurenzhangwei:
I added Agnes AI, but it doesn’t appear in the model list.

kgqq:
一样的问题，只开这个就显示没有配置模型 。

sonlyhong:
helo

tashfeenahmed:
Thanks for reporting — @gongjurenzhangwei and @kgqq, you're hitting the same thing.

**Why it happens:** Agnes AI is a recently added provider. Its models are published to the premium catalog first and only appear for free-tier installs once they age into the monthly free catalog (that promotion is coming shortly). Until then the provider shows up in the add-key list, so you can save an Agnes key — but there are no Agnes models in your catalog yet, which is why the model list looks empty. I confirmed a fresh install has 0 Agnes model rows.

**Two ways to use Agnes right now:**
1. Add it as a **Custom (OpenAI-compatible)** provider: base URL `https://apihub.agnes-ai.com/v1`, your Agnes key, and the model id (e.g. `agnes-2.0-flash` or `agnes-1.5-flash`). That path doesn't depend on the catalog.
2. If you have a Premium license key, the live catalog already includes Agnes.

**What I changed (#446):** adding a key for a provider that has no models in your current catalog no longer fails silently — the dashboard now tells you the key was saved but no models are available yet, and points you to the options above. Leaving this issue open until that's merged and Agnes is promoted to the free catalog.

tashfeenahmed:
The dashboard-side fix is now merged (#446): adding an Agnes (or any) key for a provider with no models in your current catalog now shows a clear notice instead of a silent empty list, and points you to the custom-provider workaround. Keeping this open until Agnes is promoted into the free catalog.

PR #463, opened 2026-07-04
feat(analytics): sticky range + sortable per-model table columns

jasnoorgill:
## Summary

Two small but related client-side improvements to the analytics page,
shipped together because they share infrastructure and would otherwise
conflict on the same lines of `AnalyticsPage.tsx`:

1. **Sticky time range** — the selected 24h/7d/30d range now persists
   in `localStorage` so reloads and new sessions land on the user's
   last choice.
2. **Sortable per-model table** — every column in the "per-model
   breakdown" table (except `pinned`, which renders `—`) is now
   clickable to sort. 3-state cycle: `unsorted (API order) → asc →
   desc → unsorted`. Last selection persists in `localStorage` too.

Both use the same persistence shape as the existing `theme` toggle
and `freellmapi.locale` preference. Same `useState` + `useEffect`
pattern, same try/catch on every storage access.

## Why combined

Originally shipped as PR #462 (sticky range only). Adding the sort
work on top of #462's branch required a rebase that would have
re-written #462's commits — at that point the cleanest move was a
single PR covering both. The two features are thematically adjacent
(analytics-page UX) and small individually, so combining avoids
two PRs reviewing the same file.

## Changes

| File | Change |
|---|---|
| `client/src/pages/AnalyticsPage.tsx` | +188 / -11. Added `useMemo` to imports, `STORAGE_KEY`/`loadStoredRange`/sort types, `SortableHeader` helper component, `sort` state + persistence effect, `onHeaderClick` reducer, `sortedByModel` memoized sort, replaced 8 of 9 column headers with `<SortableHeader>` (pinned stays static), added `lucide-react` ArrowUp/ArrowDown/ChevronsUpDown imports. |

**1 file, +188 / -11 across 2 commits.**

## Sort behavior

| State | Indicator | Order |
|---|---|---|
| Unsorted | ChevronsUpDown (faded) | API-returned (requests DESC) |
| Asc | ArrowUp | A→Z / 0→9 |
| Desc | ArrowDown | Z→A / 9→0 |

- Click another column → starts at asc on the new column.
- Click the same column three times → back to API order, storage
  key removed.
- Null values sort last in both directions (Excel/Sheets convention).
- Right-aligned columns flip the indicator with `flex-row-reverse` so
  the label stays closest to the data.
- `aria-sort` attribute is set on each header button for screen readers.

`pinned` is intentionally not sortable: it renders `—` for zero-pinned
rows, and the 0/>0 distinction is meaningless to sort on.

## localStorage keys

| Key | Value | When written |
|---|---|---|
| `freellmapi.analytics.range` | `'24h'` \| `'7d'` \| `'30d'` | On range change |
| `freellmapi.analytics.byModelSort` | `{"column": "...", "direction": "asc"\|"desc"}` or absent | On sort change (removed on unsort) |

Both keys are validated against an allowlist on read so a corrupted
entry falls back to the default rather than crashing the page. Both
writes are wrapped in try/catch for SSR / private-mode safety.

## Test plan

- `npx tsc -b` in `client/` — clean, 0 errors.
- `npx tsc -b` in `server/` — 9 errors in `server/src/routes/keys.ts`
  from upstream's `multer` typing bug (PR #438, pre-existing on
  `origin/main` at `340f414`). **Not introduced by this PR** —
  same errors reproduce on a clean `git checkout origin/main`.
- `npx vitest run` in `server/` — 524/524 passing, 0 regressions.
- Live verification: deployed on `integration/all-my-prs @ d08383b`.
  Service running on PID 864746, client bundle has both
  `freellmapi.analytics.range` and `freellmapi.analytics.byModelSort`
  storage keys + the sort logic.

Manual smoke:
1. Open `/analytics`, click 30d. Hard-reload — opens on 30d.
2. Click any column header (e.g. "Latency") — sorts asc, indicator
   shows ↑, header text darkens.
3. Click again — sorts desc, indicator shows ↓.
4. Click a third time — back to API order, indicator faded.
5. Click a different column — starts at asc on the new column.
6. Reload — your last sort is restored.
7. Hard-reload after clicking through to unsort — the key is
   removed from localStorage, the page loads in API order.

## Compatibility

No breaking changes. Defaults are unchanged (7d range, API-natural
row order). The persistence is additive. New `localStorage` keys
are namespaced under `freellmapi.*` per the existing convention.

## Supersedes

Closes #462 — the sticky-range half of this work is included here.

References operator sessions 2026-07-04. No upstream issue number
yet — happy to file one if requested.

### DIFF
--- a/client/src/pages/AnalyticsPage.tsx
+++ b/client/src/pages/AnalyticsPage.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import { useQuery } from '@tanstack/react-query'
 import {
   BarChart, Bar, XAxis, YAxis, CartesianGrid, Tooltip, ResponsiveContainer,
@@ -13,6 +13,26 @@ import { formatSqliteUtcToLocalTime } from '@/lib/utils'
 import { useI18n } from '@/i18n'
 
 type TimeRange = '24h' | '7d' | '30d'
+
+const TIME_RANGES: readonly TimeRange[] = ['24h', '7d', '30d']
+const STORAGE_KEY = 'freellmapi.analytics.range'
+const DEFAULT_RANGE: TimeRange = '7d'
+
+// Read the previously-selected range from localStorage. Invalid or missing
+// values fall back to the 7d default so a corrupted entry never bricks the
+// page; SSR-safety follows the same try/catch shape as `lib/api.ts`.
+function loadStoredRange(): TimeRange {
+  if (typeof window === 'undefined') return DEFAULT_RANGE
+  try {
+    const stored = window.localStorage.getItem(STORAGE_KEY)
+    if (stored && (TIME_RANGES as readonly string[]).includes(stored)) {
+      return stored as TimeRange
+    }
+  } catch {
+    /* localStorage unavailable (private mode, etc.) — use default */
+  }
+  return DEFAULT_RANGE
+}
 
 function formatTokens(n?: number): string {
   if (!n) return '0'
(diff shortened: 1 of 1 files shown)


tashfeenahmed:
Thanks, the sticky range and sortable columns work looks good and CI was green. Main has since shipped the redesigned Analytics page (#467) plus a new Recent Calls table (#474), so this now conflicts on AnalyticsPage.tsx. Could you rebase onto latest main and re-apply the sticky range and sort on the new table layout? Happy to merge once it's green again.

jasnoorgill:
Rebased, filed https://github.com/tashfeenahmed/freellmapi/pull/490

Thanks
Jasnoor

Issue #440, opened 2026-07-01
[Security] Missing SSRF guard on custom-provider base_url — cloud IMDS / internal ports reachable


chirag127:
## Problem
`routes/keys.ts:260` validates `customProviderSchema.baseUrl` with only `z.string().url()`. Anyone with dashboard access can set `base_url` to:
- `http://169.254.169.254/latest/meta-data/` (AWS IMDS)
- `http://metadata.google.internal/` (GCP IMDS)
- `http://localhost:<internal-admin-port>`
- `http://<private-network-IP>`

The proxy then makes authenticated outbound requests to these targets with the user-supplied API key attached — a classic SSRF vector that also exfiltrates internal responses back through the `/v1/chat/completions` return path.

## Suggested fix
Add a URL parser that:
1. Resolves DNS → checks resulting IP is not private (RFC 1918) or link-local
2. Blocks cloud-metadata addresses (169.254.169.254, metadata.google.internal, etc.)
3. Optionally: env flag `FREELLMAPI_ALLOW_PRIVATE_PROVIDER_URLS=true` for local-LLM users (llama.cpp / Ollama / LM Studio)

OmniRoute has a reference implementation at `src/shared/network/outboundUrlGuard.ts` (three-tier gate: public-by-default, opt-in local, opt-in private, IMDS always blocked).

Thanks for the great work on freellmapi!

tashfeenahmed:
Fixed in #443 (merged). Custom-provider base URLs now go through an outbound guard (`server/src/lib/url-guard.ts`): cloud metadata and link-local addresses are always blocked, private/loopback ranges are allowed by default (so local Ollama / LM Studio / llama.cpp keep working) and can be blocked with `FREEAPI_BLOCK_PRIVATE_PROVIDER_URLS=true` on hosted installs. Enforced both at save time and again at request time. Thanks @chirag127 for the detailed report and the reference to OmniRoute's implementation.

PR #472, opened 2026-07-06
fix: 自定义 provider URL 默认限制 localhost，新增 ALLOW_REMOTE_CUSTOM_PROVIDERS…

xd-afk-eng:
… 安全开关
## 安全加固

1. **自定义 provider URL 默认限制 localhost**
   - 新增 URL 校验，非 localhost 的 HTTPS URL 需要显式设置 `ALLOW_REMOTE_CUSTOM_PROVIDERS=true`
   - 写入 config 时和运行时解析 provider 时双重校验

2. **防止历史恶意 URL**
   - 即使之前配置了恶意 base_url，升级后也会被拦截

3. **测试验证**
   - npm run test -w server -- custom-provider: 12 passed ✅
   - npm run build -w server: 通过 ✅

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,3 +25,8 @@ REQUEST_ANALYTICS_MAX_ROWS=100000
 # Vite dev dashboard. Only set this if you serve the dashboard from a
 # different host than the API (e.g. http://my-server.local).
 # DASHBOARD_ORIGINS=http://my-server.local,http://192.168.1.50
+
+# Custom providers are restricted to localhost by default to avoid turning the
+# server into a private-network request proxy. Set this only when you need a
+# remote OpenAI-compatible HTTPS endpoint.
+# ALLOW_REMOTE_CUSTOM_PROVIDERS=true
--- a/server/src/__tests__/routes/custom-provider.test.ts
+++ b/server/src/__tests__/routes/custom-provider.test.ts
@@ -48,6 +48,10 @@ describe('resolveProvider (#117)', () => {
     expect(resolveProvider('custom', '   ')).toBeUndefined();
   });
 
+  it('refuses a custom provider bound to a private remote address', () => {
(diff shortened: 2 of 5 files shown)


xd-afk-eng:
## Security Fix

1. **Custom provider URLs default to localhost only**
   - Added URL validation: remote HTTPS URLs require explicit `ALLOW_REMOTE_CUSTOM_PROVIDERS=true`
   - Validated at both write-time (config save) and runtime (provider resolution)

2. **Legacy URL protection**
   - Previously configured malicious base_urls are blocked after upgrade

3. **Testing**
   - `npm run test -w server -- custom-provider`: 12 passed ✅
   - `npm run build -w server`: passed ✅

tashfeenahmed:
Thanks for the hardening work. I agree with the goal but can't take this as-is: (1) isPrivateIpv6 runs startsWith checks for fc/fd/fe80 against the raw hostname, so legitimate public domains beginning with those letters (fd.io for example) get blocked; the check must only apply to actual IPv6 literals. (2) Hard-blocking private IPs even with the flag set, and requiring https for remote hosts, breaks the most common self-hosted setup: pointing at Ollama or LM Studio on another LAN machine over http. That's a first-class use case for this project, so a localhost-only default is too aggressive here. There's already an opt-in FREEAPI_BLOCK_PRIVATE_PROVIDER_URLS guard and cloud metadata endpoints are always blocked. The direction I'd rather take is hardening the exposed-install story (first-run setup changes are landing separately) instead of restricting LAN providers. If you'd like to rework this into fixing the IPv6 literal detection inside the existing url-guard, that would be very welcome.

tashfeenahmed:
Thanks for raising this, but main now ships a stronger guard for custom provider URLs from #440 (DNS resolution, cloud metadata IP blocking, IPv6 encoding checks, and a FREEAPI_BLOCK_PRIVATE_PROVIDER_URLS option), so I am closing this as superseded.

PR #481, opened 2026-07-06
feat: opt-in response cache for /v1/chat/completions

tashfeenahmed:
## What this does

Adds an **opt-in, exact-match response cache** for non-streaming `POST /v1/chat/completions`. When enabled, a successful completion is stored keyed by a SHA-256 canonical hash of the request, so an **identical** later request is replayed from an in-memory LRU without touching any provider: zero quota cost, near-zero latency, and one fewer 429 on the way to the daily reset. This advances the "cache" item on the 2.0 build list.

Exact match only, no fuzzy matching, so a near-miss can never return a different prompt's answer.

## Opt-in contract

**Enable (default OFF), any one of:**
- Env var `RESPONSE_CACHE=true`
- Setting `response_cache_enabled` = `1` (dashboard-toggleable at runtime via `PUT /api/cache/config`, no restart). The setting wins over the env var; clear it to fall back to the env default.

**Per-request override (one call):**
- `X-FreeLLM-Cache: on` (aliases `force`, `yes`, `1`, `true`) forces caching even when globally off
- `X-FreeLLM-Cache: off` (aliases `bypass`, `skip`, `no`, `0`, `false`) bypasses read and write
- Standard `Cache-Control: no-store` / `no-cache` also bypasses

**Response header:** `X-FreeLLM-Cache: HIT | MISS | OFF` (hits also set `X-Routed-Via: cache`).

**Tuning env vars:**
- `RESPONSE_CACHE_TTL_SECONDS` (default 3600)
- `RESPONSE_CACHE_MAX_TEMPERATURE` (default 1.0; requests above this are never cached)
- `RESPONSE_CACHE_MAX_ENTRIES` (default 5000; least-recently-used evicted past this)

**Admin API** (under `/api/cache`, behind `requireAuth`): `GET /stats` (enabled/config + entries/hits/saved tokens), `PUT /config` (`{ enabled: boolean | null }`), `DELETE /` (flush).

## Design notes

- **Key** is the request, not the route: `'auto'`/omitted model normalize together, key order is stabilized across clients, and `tools`/`tool_choice` are part of the key so a different tool set never collides.
- **Temperature-gated**: only requests at/below the max temperature are cached (high temperature wants variety).
- **Bounded in-memory LRU** with TTL expiry; truncated turns (`finish_reason: 'length'`) are never cached.
- **Streaming always bypasses** in v1 (never read from or written to the cache).
- **Quota-safe**: a hit skips `recordRequest`/`recordTokens` and is not re-logged as provider usage, so cached replays do not inflate quota or analytics.
- **Fail-safe**: the settings lookup that decides the master switch is wrapped, so an uninitialized DB disables the cache instead of throwing on the proxy hot path (mirrors `services/ratelimit.ts`).

## Credit

Ported and adapted from [@nandukmelath](https://github.com/nandukmelath)'s fork of freellmapi (MIT). Reworked to current main:

- Converted the fork's SQLite-backed store to an **in-memory LRU** (no migration; matches the intended LRU design and avoids the Db-interface seam entirely).
- Added the **settings-table runtime toggle** (the fork was env-only).
- Scoped to **exact-match, non-streaming**: dropped the fork's semantic-embedding and streaming-cache layers as out of scope for v1 (they can follow later).
- Uses main's `sanitizeResponse` on the cached body and current header conventions.

## Tests

- **24 unit** (`services/cache.test.ts`): key canonicalization/stability, model + tools isolation, TTL expiry, LRU eviction including recency (spares a recently-read entry), hit accounting / saved-token stats, directive parsing, env switch, and the settings-backed runtime toggle.
- **9 proxy integration** (`routes/proxy-cache.test.ts`): hit/miss with provider-call counting, quota preserved on a hit, header force/bypass, high-temperature skip, streaming-request bypass, and the stats + flush admin endpoints.
- Full server suite green: **862 tests / 87 files passing** with `vitest run --pool=forks`. `tsc --noEmit` clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,23 @@ PORT=3001
 # set to 0 to disable the cap. Example — cap OpenRouter at 50 requests/day:
 # PROVIDER_DAILY_REQUEST_CAP_OPENROUTER=50
 
+# Opt-in response cache. When on, a successful non-streaming /v1/chat/completions
+# answer is stored in a bounded in-memory LRU keyed by a canonical hash of the
+# request, so an IDENTICAL later request is served from memory without spending
+# any provider quota. Exact match only (no fuzzy matching), so a near-miss never
+# returns a different prompt's answer. Off by default; set to true to enable.
+# Can also be toggled at runtime via the dashboard (PUT /api/cache/config), and
+# per-request with the `X-FreeLLM-Cache: on|off` header. A restart flushes it.
+# RESPONSE_CACHE=false
+# How long a cached answer stays fresh, in seconds (default: 3600 = 1 hour).
+# RESPONSE_CACHE_TTL_SECONDS=3600
+# Only cache requests at/below this temperature; higher temperatures want fresh
+# variety, so they are never cached. Default 1.0 caches everything when enabled;
+# lower it (e.g. 0.2) to cache only near-deterministic calls.
+# RESPONSE_CACHE_MAX_TEMPERATURE=1.0
+# Hard cap on stored entries; the least-recently-used are evicted past this.
+# RESPONSE_CACHE_MAX_ENTRIES=5000
+
 # Custom-provider URL policy. Cloud metadata and link-local addresses
 # (169.254.169.254, metadata.google.internal, fe80::/10, ...) are ALWAYS
 # blocked as custom provider base URLs. By default localhost and private LAN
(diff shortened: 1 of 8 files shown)


tashfeenahmed:
Addressed the review finding on cache-key coverage in cd6cad0: `computeCacheKey` now also keys `response_format`, `stop`, `n`, `seed`, `presence_penalty`, `frequency_penalty`, `logit_bias`, `logprobs`, and `top_logprobs` (key shape v2), so requests differing only in one of those knobs can never be served each other's cached answer. The proxy call site passes the normalized `stop` and reads the schema-stripped knobs from the raw request body (explicit null treated as absent). New tests cover key divergence per knob, identical-with-knobs stability, and an integration check that `response_format: json_object` vs plain is a MISS while an exact repeat still HITs. Cache tests now 38 (28 unit + 10 integration); full suite 867 passing, tsc clean.

PR #498, opened 2026-07-08
feat(security): harden admin endpoints and security middleware

s-uryansh:
## Summary

Several defense-in-depth security improvements for the admin surface.

### Changes

- Enable a stricter CSP using Helmet
- Add rate limiting for authenticated admin endpoints
- Require password re-authentication before exporting API keys
- Prevent untrusted `X-Forwarded-For` headers unless `trust proxy` is enabled
- Sanitize production 5xx error responses while preserving server-side logging

Intention of these changes are for security hardening. No workflow of actual application should be changed by this commit


tashfeenahmed:
Thanks for putting real work into this — the security direction is right, but it's too much in one PR to land as-is. Could you split it? The rate limiter, the X-Forwarded-For trust change, and the 5xx sanitization I want as they are, and those can merge quickly on their own.

The other two need more thought. The CSP needs proof that a production SPA build still loads — the old comment about the hashed-asset loader was there for a reason, so please verify against a real `npm run build` output, not just the dev server. And the `x-reauth-password` header on /api/keys/export is a breaking change for the desktop app, which doesn't send it; that one has to be opt-in rather than enforced by default.

Also, the rate-limit-header test contradicts itself: `does not set rate limit headers on /api/ping` then asserts `x-ratelimit-limit` is `60`. Worth deciding which behavior you actually want there.

PR #509, opened 2026-07-11
feat(server): make per-provider chat HTTP timeout configurable via PROVIDER_TIMEOUT_<PLATFORM>

jasnoorgill:
## What

Hardcoded provider timeouts become overridable via env var:

```
PROVIDER_TIMEOUT_GROQ=30000
PROVIDER_TIMEOUT_NVIDIA=180000
PROVIDER_TIMEOUT_COHERE=45000
PROVIDER_TIMEOUT_CLOUDFLARE=60000
PROVIDER_TIMEOUT_AIHORDE=240000
PROVIDER_TIMEOUT_CUSTOM=0   # disable timeout
```

Set to `0` to disable the timeout entirely (no AbortSignal fires).

## Coverage

All 25 chat platforms honour `PROVIDER_TIMEOUT_<PLATFORM>`:

| Default | Platforms |
| --- | --- |
| 60s | google |
| 90s | nvidia |
| 120s | ollama, aihorde, custom |
| 15s (generic) | groq, cerebras, mistral, openrouter, github, cohere, cloudflare, zhipu, huggingface, kilo, pollinations, llm7, opencode, ovh, agnes, reka, siliconflow, routeway, bazaarlink, ainative |

This is split across **two commits** because the initial plumbing missed
three providers that didn't take a constructor `timeoutMs` option:

1. `feat(server): make per-provider chat HTTP timeout configurable via PROVIDER_TIMEOUT_<PLATFORM>`
   — adds the helper + plumbing for google / nvidia / ollama / custom and
   every OpenAICompatProvider instance (16 platforms).
2. `feat(server): extend PROVIDER_TIMEOUT_<PLATFORM> to cohere, cloudflare, aihorde`
   — adds the constructor option to the three remaining providers that
   had hardcoded call sites (CohereProvider, CloudflareProvider, AIHordeProvider).

## Why

Operators currently have no way to tune a single provider's HTTP
timeout without forking. A few concrete cases this fixes:

1. Local llama.cpp / vLLM / Ollama on CPU — 15s is far too tight even
   for a single token of output from a multi-second-per-token model.
   Today the only escape valve is editing `providers/index.ts`.
2. Slow-but-functional upstreams (e.g. an aggregator that proxies
   through a contended endpoint) — needs more headroom than the
   default.
3. Aggressive debugging — set `PROVIDER_TIMEOUT_GROQ=0` to see the
   real upstream latency without AbortError interference.
4. AI Horde's queued generations — 120s default is sometimes not
   enough during off-peak hours with sparse volunteer workers.

## Pattern

Mirrors the existing `PROVIDER_DAILY_REQUEST_CAP_<PLATFORM>` family
in `services/ratelimit.ts`:

- Read once at module load when each provider is registered (matches
  the daily-cap resolution model — restart required to change).
- Per-platform override; missing/unparseable values fall back to the
  built-in default.
- Added `/^PROVIDER_TIMEOUT_[A-Z0-9_]+$/ to KNOWN_UNDOCUMENTED_PATTERNS
  in `env-drift.ts` so the new pattern doesn't trigger .env drift
  warnings.

## Testing

- New unit test: `server/src/__tests__/providers/index.test.ts`
  covers defaults, env override, `0` semantics, garbage/negative
  values, overriding providers with non-default built-ins, the new
  cohere/cloudflare/aihorde regression coverage, and registry integrity.
- Updated `server/src/__tests__/lib/env-drift.test.ts` to lock the
  new allowlist regex.
- Full suite: 934/934 pass on a clean `origin/main` checkout with
  both commits applied.
- Live-verified on my fork's deployment (systemd unit restarted,
  /api/health returns 401 as expected).

## Docs

`PROVIDER_TIMEOUT_<PLATFORM>` block added to `.env.example`
directly under the existing `PROVIDER_DAILY_REQUEST_CAP_<PLATFORM>`
block, listing every built-in default and examples for Groq and
Cohere.

## Files

- `server/src/providers/index.ts` — `getProviderTimeoutMs` helper
  + 7 call sites + DEFAULT_PROVIDER_TIMEOUTS_MS map (5 entries).
- `server/src/providers/cohere.ts` — new constructor option +
  threaded through 2 of 3 `fetchWithTimeout` sites (validateKey keeps
  a short 10s hardcoded timeout — key checks should be fast).
- `server/src/providers/cloudflare.ts` — same shape as Cohere.
- `server/src/providers/aihorde.ts` — replaces the module-level
  `HORDE_TIMEOUT_MS` constant with a constructor option.
- `server/src/lib/env-drift.ts` — one-line allowlist entry.
- `server/src/__tests__/providers/index.test.ts` — new (180 lines).
- `server/src/__tests__/lib/env-drift.test.ts` — two-line addition.
- `.env.example` — documentation block.

Total: +214 / −23 across 8 files.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,13 @@ PORT=3001
 # set to 0 to disable the cap. Example — cap OpenRouter at 50 requests/day:
 # PROVIDER_DAILY_REQUEST_CAP_OPENROUTER=50
 
+# Per-provider HTTP timeout for outbound chat/embedding requests, named
+# PROVIDER_TIMEOUT_<PLATFORM> (platform name upper-cased, in milliseconds).
+# Overrides the built-in default for that provider; set to 0 to disable the
+# timeout entirely. Built-in defaults: google=60000, nvidia=90000, ollama=120000,
+# custom=120000, all others=15000. Example — give Groq 30s instead of 15s:
+# PROVIDER_TIMEOUT_GROQ=30000
+
 # Wall-clock budget for provider failover, in milliseconds (default: 45000).
 # A request that keeps failing over stops STARTING new attempts once this much
 # time has passed and returns the exhaustion error instead (the first attempt
--- a/.env.example
+++ b/.env.example
@@ -42,8 +42,9 @@ PORT=3001
 # PROVIDER_TIMEOUT_<PLATFORM> (platform name upper-cased, in milliseconds).
 # Overrides the built-in default for that provider; set to 0 to disable the
 # timeout entirely. Built-in defaults: google=60000, nvidia=90000, ollama=120000,
-# custom=120000, all others=15000. Example — give Groq 30s instead of 15s:
+# aihorde=120000, custom=120000, all others=15000. Examples:
 # PROVIDER_TIMEOUT_GROQ=30000
+# PROVIDER_TIMEOUT_COHERE=45000
 
 # Wall-clock budget for provider failover, in milliseconds (default: 45000).
 # A request that keeps failing over stops STARTING new attempts once this much
--- a/server/src/__tests__/lib/env-drift.test.ts
+++ b/server/src/__tests__/lib/env-drift.test.ts
@@ -50,6 +50,8 @@ describe('env drift warnings', () => {
         'OLD_FLAG=true',
         'NODE_ENV=production',
         'PROVIDER_DAILY_REQUEST_CAP_GROQ=50',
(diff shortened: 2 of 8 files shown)


tashfeenahmed:
Thanks for this, the idea was right and it has now landed in #567 reworked against the current provider defaults (all platforms covered, safe 0-disables semantics, plus the stream stall timer), with you credited as co-author on the commit and in the README.

PR #537, opened 2026-07-14
test: add unit coverage for inferQuotaPoolKey

ashstru9-cloud:
## Summary
- `inferQuotaPoolKey` in `server/src/services/provider-quota.ts` had no unit tests, despite being a pure function that maps platforms/models to quota pool keys.
- Adds tests covering the openrouter free/account split, several fixed shared-pool platforms, and the fallback behavior for unrecognized platforms (with and without a model id).

## Test plan
- [x] `npx vitest run src/__tests__/services/provider-quota.test.ts` passes (4/4)

### DIFF
--- a/server/src/__tests__/services/provider-quota.test.ts
+++ b/server/src/__tests__/services/provider-quota.test.ts
@@ -1,0 +1,27 @@
+import { describe, it, expect } from 'vitest';
+import { inferQuotaPoolKey } from '../../services/provider-quota.js';
+
+describe('inferQuotaPoolKey', () => {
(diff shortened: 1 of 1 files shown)


tashfeenahmed:
Thanks for the PR. Closing it because inferQuotaPoolKey already has unit coverage (added in #479), and this file conflicts with that suite directly.

That said, four of your assertions cover cases the existing tests miss: google, cerebras, huggingface, and whitespace model ids. If you rebase and append just those four to the existing describe block, I'll happily merge that.

PR #557, opened 2026-07-17
feat(keys): show key counts in provider selector

iisyw:
Display the number of configured keys as a badge before each provider in the add-key dialog.


iisyw:
e5a0206f37c4e22ffdd07deb5a85dbeb

Please take a look at this PR. The changes are minor and low-risk, yet I believe this is necessary; it allows for a clear view of how many keys have been added for each supplier and makes it easier to spot new suppliers that can be added.

PR #561, opened 2026-07-17
feat: add local Ollama provider support with prefix stripping

niko4244:
## Summary

Two changes to enable using a local Ollama instance (http://127.0.0.1:11434) instead of the Ollama Cloud API:

### 1. Local Ollama base URL (providers/index.ts)

Changes the Ollama provider `baseUrl` from `https://ollama.com/v1` to `http://127.0.0.1:11434/v1`, and renames from `'Ollama Cloud'` to `'Local Ollama'`.

### 2. Prefix stripping (providers/openai-compat.ts)

Strips the `ollama/` prefix from model IDs before sending requests to Ollama's API. FreeLLMAPI stores model IDs as `ollama/<name>` but the local Ollama API expects bare model names (e.g., `qwen3.5:4b`, not `ollama/qwen3.5:4b`).

This is a generic fix — any FreeLLMAPI user running a local Ollama instance will hit this prefix mismatch.

## Backwards Compatibility

- The prefix stripping is a no-op when no `ollama/` prefix exists
- The base URL can be overridden via provider config for users who still want the cloud endpoint

## Testing

Tested end-to-end with 19 local Ollama models (qwen3.5, qwythos, gemma3, qwen3-coder, etc.) — all route correctly through the local instance. Also tested cloud providers (nvidia, minimax, mistral) to confirm the change doesn't affect non-Ollama routing.

### DIFF
--- a/server/src/providers/index.ts
+++ b/server/src/providers/index.ts
@@ -94,23 +94,14 @@ register(new OpenAICompatProvider({
 }));
 
 // Moonshot direct integration was dropped in V4 (paid-only); MiniMax direct
-// was dropped in V4 (superseded by the OpenRouter route).
-
-// Ollama Cloud — OpenAI-compatible. Free plan: 1 concurrent model, 5h session
-// caps, GPU-time-based quota (not per-token). Many catalog models on the
-// /v1/models list are subscription-only — Free returns 403 with an explicit
-// "this model requires a subscription" message. Catalog rows are filtered to
-// confirmed-Free entries.
-//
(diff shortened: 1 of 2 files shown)


tashfeenahmed:
Thanks for the PR. I can't take this one: it repoints the shared Ollama Cloud provider at 127.0.0.1:11434, which would break the hosted Ollama catalog for everyone who isn't running a local daemon.

The good news is local Ollama already works: add a custom provider with your local base URL and the models come through. If you're up for it, a docs PR showing that setup would genuinely help people.

PR #570, opened 2026-07-20
fix: change default PORT from 3001 to 8080 for Railway

tusharbusiness:


tashfeenahmed:
This breaks your documented Docker setup; Railway users should just set PORT in their service settings

PR #588, opened 2026-07-23
fix: extend NVIDIA NIM stream timeout

### LINKED ISSUE #584: [Bug] NVIDIA NIM stream stalled timeout (90000ms) still occurs on large context requests
Bug Description
Despite recent updates aimed at fixing timeout issues (such as PR #567), the freellmapi proxy still terminates connections on large context requests after exactly 90 seconds if no data chunks are received from the NVIDIA NIM stream.

When working with heavy context (e.g., around 100k+ input tokens), the model takes longer than 90 seconds to generate the first token or process chunks, triggering a hardcoded stream stall timeout.

Steps to Reproduce

1. Use freellmapi with a Kilo Code (or similar VS Code extension) setup.
2. Send a request with a large context/prompt to an NVIDIA-hosted model (e.g., deepseek-ai/deepseek-v4-pro).
3. Wait for the model to process. If processing/TTFT takes longer than 90 seconds, the proxy aborts the request.

Terminal Log / Error Output

`[server] [Proxy] 12:44:14 fail 0dd9bf a0 nvidia - deepseek-ai/deepseek-v4-pro lat=121418ms err="NVIDIA NIM stream stalled: no data for 90000ms (timeout)"`

`[server] [Proxy] 13:01:09 fail a25550 a0 nvidia - minimaxai/minimax-m2.7 lat=90005ms err="The operation was aborted (nvidia, chat, 90s)"`

Additional Context

- It seems the previous fix might have addressed other providers or timeouts, but the NVIDIA NIM provider still has a hardcoded 90-second stall/inactivity limit (90000ms).
- Would it be possible to increase this limit or make the NVIDIA stream timeout configurable via environment variables or config settings?

Thanks for your hard work on this project!

ousamabenyounes:
## Summary
- raise NVIDIA NIM's built-in request timeout from 90s to 180s
- give NVIDIA NIM streams a matching 180s inactivity watchdog while preserving `PROVIDER_STREAM_STALL_TIMEOUT_MS` as the global override
- add regression coverage for both the provider registration timeout and the actual SSE stream watchdog path

## Test verification (RED -> GREEN)
- Upstream-base RED: with only the new regression test present, `npm run test -w server -- src/__tests__/providers/reasoning-timeouts.test.ts` failed because NVIDIA was still registered at `90000` instead of `180000`, and the stream test recorded delays `[777, 90000, 90000, 90000]` instead of including `180000`.
- GREEN focused: `npm run test -w server -- src/__tests__/providers/reasoning-timeouts.test.ts src/__tests__/lib/provider-timeout.test.ts` passed: `2 passed`, `17 passed`.
- Full local validation: `npm run test:migrations && npm test && npm run build` passed: server `108` files / `1085` tests, build passed.
- Review gate: independent staged-diff review passed after confirming the stream path now passes the provider-specific watchdog to the shared SSE reader.

Fix #584

### DIFF
--- a/server/src/__tests__/providers/reasoning-timeouts.test.ts
+++ b/server/src/__tests__/providers/reasoning-timeouts.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, afterEach } from 'vitest';
 import { resolveProvider } from '../../providers/index.js';
 import { CloudflareProvider } from '../../providers/cloudflare.js';
+import type { ChatCompletionChunk } from '@freellmapi/shared/types.js';
 
 // Regression guard for the 2026-07-11 live-sweep finding: platforms hosting
 // hidden-reasoning or buffered non-streaming models (zhipu glm-4.7-flash 41s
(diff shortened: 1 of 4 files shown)


tashfeenahmed:
Thanks @ousamabenyounes — the diagnosis here was correct and valuable: the 90s stall watchdog really was killing healthy NIM prefills on large contexts (#584). We landed a generalization of it in 99076fc (PR #605): instead of raising nvidia's stall constant (which would also slow dead-connection detection to 3 minutes), streams now give the *first byte* its own budget — the platform's chat timeout, with nvidia raised to 180s exactly as your PR proposed — while mid-stream stalls still fail fast at 90s. That fixes the same bug for ollama, custom endpoints, Cloudflare GLM and Google, which shared it. Also added the per-platform override you'll recognize: PROVIDER_STREAM_STALL_TIMEOUT_NVIDIA. You're credited in the README. Closing as superseded.

PR #589, opened 2026-07-23
fix: keep notifications visible longer

### LINKED ISSUE #586: Notifications close too fast
Notifications that show in the bottom right corner of the screen close on their own and they do so too fast.

Image

For example, this message closes in like 5 seconds, which is not enough time to read it.

ousamabenyounes:
## Summary
- keep default toast notifications visible until the user dismisses them
- preserve explicit auto-dismiss durations for callers that still need temporary toasts
- cover persistent defaults and explicit durations in the client toast tests

## Test verification (RED -> GREEN)
- Upstream-base RED: `npm run test -w client -- src/lib/toast.test.ts` failed before the production change with `expected 10000 to be null` and `expected 12000 to be null`.
- GREEN focused: `npm run test -w client -- src/lib/toast.test.ts` passed: client `1` file / `3` tests.
- Browser verification: tested in headless Chromium through Vite with the actual `Toaster` component; a success toast stayed visible after `13s` and disappeared after clicking dismiss.
- Full local baseline: `./run-ci.sh` on latest `main` passed `npm install`, migrations `3/3`, server `119` files / `1180` tests, client `1` file / `3` tests, and build.
- Full local final: `./run-ci.sh` on the PR commit passed the same commands with the same test counts and build result.

Fix #586

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
+    "test": "vitest run",
     "lint": "eslint .",
     "preview": "vite preview"
   },
@@ -44,7 +45,8 @@
     "globals": "^17.4.0",
     "typescript": "~6.0.2",
     "typescript-eslint": "^8.58.0",
-    "vite": "^8.0.4"
+    "vite": "^8.0.4",
+    "vitest": "3.2.4"
   },
   "engines": {
     "node": ">=20.18.0 <25.0.0",
--- a/client/src/components/toaster.tsx
+++ b/client/src/components/toaster.tsx
@@ -33,6 +33,7 @@ function Toast({ toast }: { toast: ToastItem }) {
   const { t } = useI18n()
 
   useEffect(() => {
+    if (toast.duration === null) return
     const timer = window.setTimeout(() => dismissToast(toast.id), toast.duration)
     return () => window.clearTimeout(timer)
   }, [toast.id, toast.duration])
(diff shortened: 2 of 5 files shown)


emv33:
Why not keep the notification open until the user dismisses it? It's already small enough as it is and full of text. Not everybody can notice a notification appeared, then read (and understand) all that text in 12 seconds.

ousamabenyounes:
Thanks for the review @emv33. I pushed `a4a152c` to keep default toast notifications visible until the user dismisses them, while preserving explicit auto-dismiss durations for callers that need temporary toasts.

Validation: reproduced the reviewer-requested behavior change with `client/src/lib/toast.test.ts` RED (`10000`/`12000` instead of `null`), then reran the focused test GREEN. Full local `./run-ci.sh` baseline and final both passed with the same counts; browser verification in Chromium confirmed the toast stayed visible after 13s and dismissed on click.

GitHub Actions currently reports `action_required` for both CI and Docker on this fork PR before any jobs start, so there are no remote job logs to inspect yet.

tashfeenahmed:
Thanks @ousamabenyounes — right problem (#586, errors genuinely vanished too fast to read), and your vitest bootstrap for the client was a genuinely good idea that we kept. We landed a variant in 99076fc (PR #605): rather than making every toast persistent (which leaves 'Copied!' confirmations stuck until clicked), errors are sticky while success/info stay on a 5s timer, timers pause on hover and window blur, and the 4-toast cap now never evicts an unread error (the case where a persistent error could still vanish under your approach). Client vitest is in with 15 tests. You're credited in the README. Closing as superseded — appreciate both PRs!

PR #620, opened 2026-07-27
Feat/zai glm5 models

mithucste30:


tashfeenahmed:
Thanks for putting this together, and for the detail in the migration comments.

We are not going to take this one. The catalog criteria is recurring free credits, and the GLM Coding Plan is a paid subscription starting at $18/mo with no free tier, so it sits outside what we list.

One note for any future provider PR: model definitions belong in the hosted catalog rather than a DB migration. Migrations ship models to every install on update, which bypasses the catalog tier gate.

Appreciate the work you put into this.

Issue #411, opened 2026-06-28
Per-client API keys with their own system prompt (backend half of #388)


tashfeenahmed:
Splitting the second half of #388 into its own tracking issue. The **Playground UI** half (an optional client-side system-prompt box) shipped in #390. This issue covers the **backend** half, which #390 intentionally left out.

### Goal
Let each downstream client/product have its own API key with its own server-enforced system prompt, instead of every caller inheriting the single hardcoded identity. The existing unified/admin key stays backward compatible (it maps to the default prompt).

### Proposed components (from @kairwang01's draft in #388)
- **db:** a `client_profiles` table (`token_hash` lookup plus an encrypted copy of the key for masked display), plus accessors.
- **lib/system-prompt:** `resolveAuth(token)` (unified key to default; profile key to its own prompt; empty to neutral passthrough) and `prependSystemPrompt(messages, prompt)`.
- **proxy.ts / responses.ts:** authenticate via `resolveAuth` and inject the resolved prompt server-side.
- **routes/client-profiles:** CRUD (create mints a profile key, rotate, enable/disable), mounted under `/api/client-profiles`.
- **client:** a Profiles dashboard page plus nav/route to manage them.

### Why server-side (vs the #390 client approach)
#390 prepends the system message client-side, which is perfect for the Playground but trusts the caller. The point of this issue is a prompt the client cannot override: enforced at the gateway and keyed to the credential, so a downstream product gets a fixed identity regardless of what it sends.

### Acceptance
- A profile key resolves to its own prompt; the unified key keeps the default; an empty profile prompt is a neutral pipe.
- Profile keys cannot escalate to the admin/unified scope.
- CRUD, rotate, and masked display work from the dashboard.
- Tests cover `resolveAuth` precedence and the proxy/responses injection path.

Refs #388, #390.

PR #621, opened 2026-07-27
feat: add profile-scoped API workspaces

simmorsal:
Add named API profiles with dedicated URL slugs and unified API keys while preserving /v1 and /default/v1 for the built-in Default profile.

Scope user-facing behavior to the requested profile, including model allowlists and routing, Fusion and Anthropic mappings, embeddings, media, cache and guardrails, outbound proxy settings, analytics, MCP endpoints, and generated API usage examples.

Add database migrations for existing installations and profile-aware request history. Introduce a Profiles dashboard with persistent selection, profile creation and editing, icon selection, animated profile cards, API-key management, and per-profile proxy controls. Update the documentation and add integration coverage for URL, key, settings, proxy, and analytics isolation.


tashfeenahmed:
Thanks for this, really solid work. The migration up path is careful, key isolation works, and the whole existing suite still passes. I went through it in depth and there are some things I need fixed before it can go in:

1. Per-profile proxy settings never reach background traffic. Health checks, catalog sync and key probes still read the old global settings rows, which the dashboard no longer writes. The UI says proxy active while background provider calls go direct, and the config breaks on restart. That's a real IP leak for anyone routing through a VPN or Tor.
2. Key rotation is split-brained. Regenerating updates profiles.api_key but leaves settings.unified_api_key stale, and /api/settings/api-key without X-Profile-ID returns the dead key. Related: any /api write without that header lands in the invisible global row and silently no-ops. Simplest fix is probably having the dashboard middleware fall back to the Default profile when the header is absent.
3. ensureAllModelsInProfiles runs on every routing decision and every /v1/models call. That's an N+1 write loop on the request hot path. Please move it to startup or catalog sync.
4. The down() migration destroys every custom workspace: keys get re-minted at random and slugs are recomputed from names on the next up(). Seed a custom profile in roundtrip.test.ts before the down/up cycle and you'll see it fail. db:migration:down is a documented command so this needs to be safe.
5. auto:<profile-name> now hard 400s with no deprecation note. At minimum the error should point at the /<slug>/v1 replacement and the README should mention the removal instead of just deleting the section.
6. Analytics: the no-profile view now drops all non-Default traffic, profile_request_hourly is never pruned, and deleting a profile orphans its rows (no FK, no cleanup).
7. PATCH /api/models/:id with fallbackEnabled updates the global fallback_config even when a custom profile is selected, which leaks per-profile toggles into the baseline that seeds new profiles.
8. Frontend: creating a profile bounces selection back to Default (the setTimeout correction effect in profile-provider races the profiles refetch), the active profile is read from localStorage at request time so two tabs desync and write to the wrong profile, and regenerating a key invalidates ['unified-key', id] while five consumers (Playground, command palette, detail pages) still cache under bare ['unified-key'] and keep serving the revoked key.
9. The proxy dispatcher cache in lib/proxy.ts is a single global keyed on URL, so two profiles with different proxies rebuild the undici agent on every request and never close the old one.

A few smaller ones while you're in there: dashboardProfileMiddleware mounts before requireAuth so it's a pre-auth profile id oracle, the response cache and sticky sessions should key on profile id rather than slug (renames strand entries), and the Profiles page is hardcoded English while everything else ships 60 locales.

On tests: none of the new tests actually send a chat completion through a named workspace, so allowlist enforcement, the per-profile Anthropic mapping and /<slug>/mcp are all unexercised. At least one end to end test through /coding/v1 would give me a lot more confidence.

The feature itself is exactly the direction I want, happy to merge once these are addressed.

tashfeenahmed:
Hi @simmorsal, thanks for the huge amount of work here. This was genuinely one of the most thorough PRs this repo has had. I'm closing it for timing reasons, not quality.

Client profiles landed on main (#411, merged via #758) while this was open. They cover the per-key part of workspaces and store tokens as SHA-256 digests, so I can't take a second credential system with a plaintext key column on the auth path. Main has also moved a lot since your base (54 commits touch 29 of your 51 files), so a rebase would basically be a rewrite.

Some pieces of this are still genuinely missing from main, and I'd love to see them as small separate PRs built on client profiles:

1. Per-client request attribution in analytics. Right now you can't tell which client generated which traffic, and your design solved that.
2. Making client profile keys work on the Anthropic, MCP, Gemini and Ollama routes. They only work on /v1 today.
3. A per-client model allowlist.

Any of those would be very welcome, and they're each a fraction of the size of this branch. Thanks again for the effort you put in.

PR #635, opened 2026-07-28
Add in-dashboard update checker

### LINKED ISSUE #491: RFC: in-dashboard update checker for self-hosted installs
## Problem

There's no way to know when a new version is available without manually checking the repo. For self-hosted installs (git clone, Docker), users only find out about updates if they happen to visit GitHub or read the release notes.

## What I'm thinking

A lightweight "Check for updates" section in the Settings page that:

1. Hits the GitHub API to compare the local HEAD commit against the remote `main` branch
2. Shows whether an update is available, with the commit message and date
3. Displays deployment-appropriate update instructions (docker compose pull / git pull + npm build)

No automatic updates, no in-place upgrades. Just honest information + a link to the install docs.

## Why commits instead of versions

I noticed `server/package.json` is stuck at `0.2.1` while git tags go up to `v0.4.1`, and there are no formal GitHub Releases. Comparing commit SHAs works regardless of version tagging discipline.

## Questions for the maintainer

1. Is this something you'd want in the project, or is it out of scope?
2. Would you prefer commit-based checking (as described above) or version-based (requires keeping `package.json` version in sync with tags)?
3. Any preferences on where the UI lives — Settings page, a dedicated "Updates" section, or somewhere else?
4. Should it check on a schedule (like the catalog sync does) or only on manual button click?

Happy to implement this if there's interest.

4riful:
## Summary

- add a manual, authenticated update checker to Settings using commit ancestry against official `main`
- distinguish current, available, ahead, diverged, unknown, and unsupported installations
- show a focused progress modal with recent commit changelog and installation-specific update guidance
- embed source commit metadata in Docker and Desktop builds
- add translated UI copy across all 60 locales

## Reliability and privacy

- cache successful checks for five minutes and coalesce concurrent requests
- validate and bound all GitHub response metadata
- fall back to GitHub’s public Atom feed when REST API authentication or rate limits fail
- keep checks manual and disclose that they contact GitHub

## Testing

- `npm test` (1,503 server, 44 CLI, 15 client tests)
- `npm run build`
- `npm run build:all` in `desktop/`
- focused ESLint for `settings-dialog.tsx`
- i18n parity: 60 locales, 677 keys
- live authenticated check against GitHub

Closes #491


tashfeenahmed (changes requested):
Reviewed this properly — read the whole diff, ran the suites locally, and captured the actual outbound request. The engineering is good: response validation is thorough, the error paths are stable and don't leak upstream detail, the 404/401/403/429 handling is right, and the test file genuinely covers the interesting cases (cache, coalescing, Atom fallback, malformed payloads, no-identity). Nothing touches licensing, tiers, catalog sync, or any other phone-home path — `update.ts` imports `node:child_process` and `express` and nothing else. Locales are mechanically sound and `check:i18n` passes at 60/677.

Two things need to change before this can go in.

**1. There's no way to turn it off.**

Self-hosted boxes need an off switch for anything that talks to the internet, and there isn't one here — no env var, no setting. Airgapped and policy-restricted installs currently get a Settings section that can only fail, and each click costs two outbound attempts plus a `logger.error`.

Please add an opt-out following the existing convention (`# FREELLMAPI_COMPRESSION=off`, `# FREELLMAPI_CONTEXT_HANDOFF=...`), e.g. `FREELLMAPI_UPDATE_CHECK=off`. When it's off I'd expect `/check` and `/status` to report disabled without fetching or shelling out to git, and the UI section to hide rather than render a dead button.

**2. Drop the ambient `GITHUB_TOKEN` pickup.**

`resolveIdentity`/`performCheck` read bare `process.env.GITHUB_TOKEN` and attach it as `Authorization: Bearer`. I confirmed it on the wire:

```
GET https://api.github.com/repos/tashfeenahmed/freellmapi/compare/<local-sha>...main
  Accept: application/vnd.github+json
  User-Agent: freellmapi-update-checker
  X-GitHub-Api-Version: 2022-11-28
  Authorization: Bearer <whatever GITHUB_TOKEN happens to be>
```

The URL and UA are fine — the local SHA is inherent to an ancestry check and I have no problem with it. The token is the issue. `GITHUB_TOKEN` is not currently read anywhere else in this repo, so this PR introduces a new implicit read of one of the most generically-named variables there is. Plenty of Compose users already have `GITHUB_TOKEN` in a shared root `.env` for unrelated tooling, and it's frequently a broad-scope PAT. Silently borrowing it turns an anonymous update ping into one authenticated as that user's GitHub account, which is more than the check needs and more than anyone opted into.

Same host so it isn't exfiltration, but it should be explicit: namespace it (`FREELLMAPI_UPDATE_GITHUB_TOKEN` or similar) so it's only used when someone deliberately sets it for this feature.

**3. `.env.example` needs the new vars.**

Related to the above and a concrete bug on its own: nothing was added to `.env.example`, so anyone who actually sets the token gets flagged by our own env-drift checker as an unknown key on startup. Document the token var and the kill switch there. `FREELLMAPI_COMMIT_SHA` / `FREELLMAPI_INSTALL_METHOD` are build-injected and read from the process env rather than the file, so those don't trip drift — but a commented note wouldn't hurt.

---

Non-blocking, take or leave:

- Failures aren't negatively cached. Every retry re-hits the API and then the Atom feed. A short negative TTL (30–60s) would stop a stuck box from re-dialing on every click.
- `execFileSync('git', ...)` with a 5s timeout runs inside the request handler, so first open of Settings can block the event loop. Lazy is right; async would be better.
- Self-built Docker images (`docker build` with no `--build-arg`) end up as installation `unknown`, not `docker`, because the install-method assignment is gated behind a valid SHA. Cosmetic, but those users get the Releases link instead of the `docker compose pull` hint.
- `MAX_ATOM_FEED_LENGTH` is checked after `await atomResponse.text()`, so the cap doesn't actually bound what gets read into memory.
- `ARG FREELLMAPI_COMMIT_SHA` sits above the runtime `COPY` layers, so a changing SHA invalidates cache for all of them. Moving it below the COPYs keeps the layer cache useful.
- The updates section reuses `premium.checkForUpdates` and `premium.lastChecked`. Fine today, fragile if those strings get restructured — worth their own keys under `settings.`.

Fix 1–3 and I'll merge. CI is green and the rest of this is solid work.

4riful:
Addressed all three requested changes:

- Added `FREELLMAPI_UPDATE_CHECK=off`. Both `/api/update/check` and `/api/update/status` return `status: disabled` before identity resolution, so disabled mode performs no Git subprocess and no network request. The Settings section hides when that state is returned.
- Replaced ambient `GITHUB_TOKEN` pickup with the explicit `FREELLMAPI_UPDATE_GITHUB_TOKEN`; generic `GITHUB_TOKEN` is ignored. Regression tests cover both header behaviors.
- Documented both variables in `.env.example`, including the build-injected SHA/install-method note, so env-drift recognizes user configuration.

I also preserved `installation: docker` for self-built images without embedded SHA metadata, without falling through to Git discovery.

Rebased/merged current `main` and reran validation: server 1,650 tests, CLI 44, client 30, i18n 60 locales/684 keys, root production build, desktop `build:all`, focused ESLint, and `git diff --check` all pass.

4riful:
Hello there, Can I have a response about this Pr ?

4riful:
Still on this, sorry for the gap. Rebased onto current `main`; the branch is mergeable again and CI-green locally.

One thing worth raising before you re-review. Since your review, #718 added `AppVersionRow` to Settings → General — a manual release-version check against `releases/latest` — and #703 added `app-version.ts` plus `GET /api/settings/version` behind it. Merging this branch as-was would have put two update checkers, two buttons and two outbound GitHub pings in the same pane, so I made this the single surface rather than adding a second one:

- The collapsed row is now one `Row` in the wordless idiom you used: `FreeLLMAPI  v0.6.8  →  a1b2c3d  ⟳`. Nothing new to translate for the row itself.
- Words stay behind the button, in the dialog, where the commit list and the per-deployment upgrade command are — those strings were already localised.
- `/api/update/status` and `/check` now return `version` via `getAppVersion()`, so the client makes one request instead of two. `GET /api/settings/version` is untouched, tests and all.
- `compareVersions` and the client-side `releases/latest` fetch are gone; the server does the comparing against the local commit, so a source or container install that tracks `main` still gets a true answer rather than one that depends on tagging discipline.

That deletes code you wrote three weeks ago, so I'd rather ask than assume: happy to put this back as its own section and leave `AppVersionRow` exactly as it is, if you'd prefer the smaller diff. Your call and I'll turn it around quickly either way.

Also picked up the non-blocking notes from your review:

- Failures are negatively cached for 45s, so a box that can't reach GitHub stops re-dialling the API and then the Atom feed on every click.
- `git rev-parse HEAD` runs through `execFile` instead of `execFileSync`, so first open of Settings no longer blocks the event loop. The test seam is now `execFile` and returns `{ stdout }`.
- `MAX_ATOM_FEED_LENGTH` is enforced on the way in — `content-length` first, then a byte ceiling while streaming. Checking `.length` after `await response.text()` bounded nothing, as you said.
- The dialog no longer borrows `premium.checkForUpdates` / `premium.lastChecked`; both now exist under `settings.*` in all 60 locales, carrying each locale's existing translation.
- `ARG FREELLMAPI_COMMIT_SHA` moved below the runtime `COPY` layers.

Install-method preservation for self-built images went in last round and keeps its test.

Validation: server 1,890 · CLI 44 · client 69 · `check:i18n` 60 locales / 773 keys · `npm run build` · eslint on the changed files.

tashfeenahmed:
Design question answered: yes, one update surface is right, and your fold of the release-version row into the checker is exactly what I wanted. Verified locally against current main: clean merge, 1950 server tests, i18n across 60 locales, client build all green. Merging. Thanks for sticking with this one through the wait.

tashfeenahmed (approved):
All three requested changes verified fixed, and the #718 fold is the right call. Tested against current main: 1950 server tests, 60-locale i18n, client build all green.

Issue #488, opened 2026-07-07
Custom API Providers, Auto Load and weights


euphoria360:
Hi.
Currently, in order to use a custom provider (like [bynara.id](https://router.bynara.id/)) we have to manually add supported models.
This is not very useful as these providers add and remove models frequently.
For example now that we have sonnet-5, they have added it to their models.

I usually have to manually query what they have available every now and then:
`curl -s [URL]/v1/models -H "Authorization: Bearer [API-Key]" |jq -r '.data[] | " \(.id) (\(.owned_by // .provider // "unknown"))"'`
then add them to the **Models Box**.
Can this be done automatically by **FreeLLMAPI** itself?

another Issue is that all of the added models by custom providers have same default weights **(Reliability: 50, Speed: 50, Intelligence: 0)**.
This makes them literally unusable with _Routing Strategies_, unless we choose Manual mode.
Is there a way to assess their weights in the same way FreeLLMAPI does it for supported providers?

tashfeenahmed:
Both halves shipped in #649, going out in v0.6.2. The custom-endpoint form and every saved custom key now have a "Fetch models" button that hits your endpoint's `/models` with your own key and gives you a checkbox list of what's there, with already-registered ones locked — that replaces the `curl … | jq` loop. It's a one-click fetch rather than a background auto-refresh; if you want it to poll on its own, open a separate issue for that. The weights half is fixed too: new custom models now seed at the local catalog's median tier/intelligence/speed instead of `Custom`/0, so they actually compete in the auto strategies, and ranks you've tuned yourself are never overwritten on re-registration. Closing.

euphoria360:
unfortunately the weights doesnt work. 
i played with the intelligence weight,set it 1 or 4 or 10 and saved it. but nothing changes and "Intelligence" is still 52.

<img width="1117" height="391" alt="Image" src="https://github.com/user-attachments/assets/90f9d0c6-9440-4478-9085-36b94cad7d88" />

PR #649, opened 2026-07-28
Auto-discover models on a custom endpoint, and stop burying them at intelligence 0 (#488)

tashfeenahmed:
Closes half of #488 — the "can this be done automatically" part, plus the weights complaint.

Custom providers change their model list constantly, so keeping it current means re-running `curl .../v1/models | jq` and pasting ids by hand. And every custom model was inserted at `size_label 'Custom'` (not a capability tier, so the bandit scored its intelligence axis 0) with rank 50, which meant it lost every auto-route — exactly what the issue reports.

**Discovery.** `POST /api/keys/custom/discover-models` (dashboard auth) GETs `${baseUrl}/models` through the provider adapter's existing catalog fetch — the same one `validateKey` uses, extracted rather than copied — and returns the ids with the ones already registered on that endpoint flagged. It is tolerant of what relays actually send (`data` / `models` / `result`, bare arrays, plain strings, `id`/`name`/`model`/`slug`), caps the body at 2 MB and the list at 500, and maps an unreachable box, an upstream 401 and an unreadable answer to their own statuses. Nothing is written, and it reads only the operator's own base URL with their own key.

**Bulk add.** `POST /api/keys/custom` now accepts a `keyId` as well as a `baseUrl`, so the add that follows discovery binds to the endpoint credential the list came from. That stays consistent with the per-endpoint key pool from #619/#640: a model already on the endpoint keeps its binding, and every key of the pool still serves it. The response reports created vs. already-registered.

**Seeding.** A new custom model now starts at the local catalog's median tier/intelligence/speed instead of the floor — unknown is not the same as worst. Single adds, bulk adds and declarative config all use it, and re-registering still never overwrites ranks you have tuned yourself.

**UI.** "Fetch models" on the custom-endpoint form and on each saved custom key, opening a checkbox list with the already-added models locked.

This is discovery on your *own* configured endpoint with your *own* key. It does not fetch or refresh the provider catalog and does not touch catalog sync.

New tests cover the happy path against a mocked relay, the parsing variants, auth, the size cap, upstream failures, the endpoint binding, dedup and the median seeding. Full server suite (1604) and client build pass.


Aldo-f:
As there is auto discover for custom providers we do not need to add a list of models? 
Nor need that field to be required.

Screenshot_20260731_124839

Another tiny bug:
When providing a custom name for a custom provider the name will be replaced to 'Custom'.

image

PR #658, opened 2026-07-29
Fix custom endpoint model identity (#651)

### LINKED ISSUE #651: Custom relay models collide on (platform, model_id) — per-endpoint identity
Every custom relay is stored under `platform = 'custom'`, and the `models` table is unique on `(platform, model_id)`. So two relays that both offer, say, `deepseek-v3.1` collapse into a single row: one `enabled` flag, one set of ranks, one binding anchor. Turning the model off because relay A is broken also turns it off for relay B, and there is no way to keep it on for one and not the other.

The scoring side has the same shape. Reliability and speed stats key on `platform:model_id`, so both relays feed one bucket — a relay that times out half the time drags down the copy served by the relay that's fine, and the router can't tell them apart when it picks. This is what @qq97693453 was describing in #619 (and the cross-provider half of #580, which is fixed for catalog platforms because those have distinct `platform` values — `custom` is the case that doesn't).

**Already shipped as groundwork**

- #640 — per-endpoint key pools; adding a second key to the same `base_url` no longer overwrites the first, and nothing gets silently rebound.
- #649 — model discovery against a specific endpoint, and adds bind to the key the list came from.
- #645 — timeouts now cost speed score, so a slow relay is at least measured honestly within its bucket.

**Remaining**

Model identity has to be keyed per endpoint rather than per platform:

- schema migration off `UNIQUE(platform, model_id)` — either an endpoint discriminator in the models table or a junction table;
- re-key the stats buckets so reliability/speed are per (endpoint, model) instead of per (platform, model), with a backfill story for existing rows;
- `fallback_config` fixups — entries currently point at the merged row, and the `platform:model_id` hard pin has no way to name one custom endpoint;
- UI: the models list needs to show which endpoint a custom row belongs to once duplicates can coexist.

#486 proposed a `custom_key_bindings` junction table for the key-isolation problem; the same shape is prior art for the identity problem and worth reading before designing this.

nicyoong:
# Fix custom endpoint model identity (#651)

## Summary

Fixes #651 by allowing multiple custom OpenAI-compatible endpoints to register
the same upstream `model_id` without overwriting each other's model row or
routing traffic to the wrong base URL.

Custom chat models are now identified by their stable custom endpoint and
`model_id`. Built-in providers retain their existing `(platform, model_id)`
identity.

## What changed

- Add a file-per-migration schema change for stable custom endpoint records,
  endpoint-aware custom key associations, and endpoint-scoped custom-model
  uniqueness.
- Preserve existing model ids and fallback/profile references while rebuilding
  the legacy `models` table without its global custom-model uniqueness rule.
- Keep pricing columns during the table rebuild and make the migration
  reversible while no duplicate custom model ids exist across endpoints.
- Register, re-register, list, route, score, group, delete, and budget custom
  chat models by endpoint identity rather than only `platform + model_id`.
- Attribute new request logs, routing statistics, fallback token usage, and
  analytics to `models.id`, so two relays exposing the same model no longer
  share performance or usage data.
- Make declarative custom-provider configuration endpoint-aware.
- Update built-in-only test fixtures to use targetless SQLite upserts, since
  the legacy global `ON CONFLICT(platform, model_id)` target no longer exists.

## Behaviour before and after

Before:

1. Register `shared-model` on relay A.
2. Register `shared-model` on relay B.
3. The single `custom/shared-model` row is rebound to relay B; relay A is no
   longer independently routable.

After:

1. Register `shared-model` on relay A.
2. Register `shared-model` on relay B.
3. Two independent model rows exist, each routes through its own endpoint and
   each can be removed without affecting the other.

## Scope

This PR intentionally covers custom **chat** models, which is the collision
path described by #651. Custom embedding and media-model tables retain their
existing key-based identity and are out of scope for this focused fix.

## Tests

Ran successfully:

```text
npm run build -w server
npm run test -w server -- \
  src/__tests__/routes/analytics.test.ts \
  src/__tests__/routes/custom-provider.test.ts \
  src/__tests__/routes/custom-provider-multikey.test.ts \
  src/__tests__/db/migrate/roundtrip.test.ts \
  src/__tests__/services/declarative-config.test.ts \
  src/__tests__/services/declarative-config-profiles.test.ts
```

The focused run passed 70 tests. The added regression test verifies that two
custom endpoints with the same model id receive different database ids, route
to their respective base URLs, and retain the surviving model when the other
endpoint is deleted.

I also ran the full server suite. All issue-related tests and the affected
routing, analytics, migration, ratelimit, and model-retirement tests passed.
The suite has two Windows-specific failures in `db/hardening.test.ts`: POSIX
permission assertions expect `0600`, while the Windows filesystem reports
mode `054`. These failures are unrelated to this change.

## Migration notes

Existing custom keys are assigned to normalized endpoint records. Existing
custom request history is backfilled only when its key still identifies the
endpoint unambiguously; ambiguous orphaned rows intentionally remain without a
`model_db_id` rather than being attributed to the wrong relay.

### DIFF
--- a/server/src/__tests__/db/migrate/roundtrip.test.ts
+++ b/server/src/__tests__/db/migrate/roundtrip.test.ts
@@ -21,6 +21,7 @@ const REQUEST_SERVED_MODEL_FILENAME = '20260726_000005_request_served_model.ts';
 const ATTEMPT_ERROR_SUMMARY_FILENAME = '20260726_000006_attempt_error_summary.ts';
 const AGENT_COMPATIBILITY_FILENAME = '20260727_000001_agent_compatibility.ts';
 const TOMBSTONE_PROVENANCE_FILENAME = '20260728_000001_tombstone_provenance.ts';
+const CUSTOM_ENDPOINT_MODEL_IDENTITY_FILENAME = '20260729_000001_custom_endpoint_model_identity.ts';
 
 interface SchemaRow {
   type: string;
@@ -88,6 +89,7 @@ describe('migration round trip', () => {
         ATTEMPT_ERROR_SUMMARY_FILENAME,
         AGENT_COMPATIBILITY_FILENAME,
         TOMBSTONE_PROVENANCE_FILENAME,
+        CUSTOM_ENDPOINT_MODEL_IDENTITY_FILENAME,
       ]);
     } finally {
       db.close();
--- a/server/src/__tests__/routes/custom-provider.test.ts
+++ b/server/src/__tests__/routes/custom-provider.test.ts
@@ -309,6 +309,46 @@ describe('Custom Provider Endpoints', () => {
     });
   });
 
+  describe('same model on independent custom endpoints (#651)', () => {
+    it('keeps distinct model rows, routes, and lifecycle when endpoint model ids collide', async () => {
+      const db = getDb();
+      db.prepare("DELETE FROM fallback_config WHERE model_db_id IN (SELECT id FROM models WHERE platform = 'custom')").run();
+      db.prepare("DELETE FROM models WHERE platform = 'custom'").run();
+      db.prepare("DELETE FROM api_keys WHERE platform = 'custom'").run();
+
+      const first = await post(app, '/api/keys/custom', {
+        baseUrl: 'http://127.0.0.1:18081/v1', model: 'shared-relay-model', label: 'Relay one',
+      });
+      const second = await post(app, '/api/keys/custom', {
+        baseUrl: 'http://127.0.0.1:18082/v1', model: 'shared-relay-model', label: 'Relay two',
+      });
+      expect(first.status).toBe(201);
+      expect(second.status).toBe(201);
+      expect(first.body.modelDbId).not.toBe(second.body.modelDbId);
+
+      const rows = db.prepare(`
+        SELECT m.id, m.custom_endpoint_id, k.base_url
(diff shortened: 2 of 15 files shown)


tashfeenahmed:
Thanks for this — and sorry for the collision. I had a fix for #651 in flight at the same time and merged it in #659 earlier today, so this now conflicts against main. Closing on those grounds only, not on quality: your write-up of the problem was accurate and the approach is sound.

For what it's worth, we landed in a similar place — a rebuilt `models` table that preserves row ids so fallback/profile references survive, endpoint-scoped uniqueness for custom rows, built-ins keeping plain `(platform, model_id)`, and a reversible migration that refuses when real duplicates exist. Ours keys the scope on the normalized base URL rather than a separate endpoint record.

Two things you did that we didn't, which I'd take as follow-ups if you're interested:

- **Attributing request logs, routing stats and analytics to `models.id`** instead of the platform/model_id pair. We re-keyed stats but left the analytics attribution alone, and yours is the cleaner shape.
- **First-class endpoint records.** We used an opaque scope string. A real endpoint table is a better home for per-key group tags, which is exactly what #657 needs next — same-site keys on Chinese relays belong to different 分组 with different model access, so keys sharing a base URL aren't interchangeable there.

Either would be welcome as its own PR.

nicyoong:
Yes. I will follow your main repo and redo the changes as their own PRs. Thank you

PR #660, opened 2026-07-29
feat: Add endpoint exposure settings and provider key indicator

Vigno04:
Description
This PR introduces two main UI/UX improvements to the dashboard and API behavior:

1. Endpoint Exposure Settings Added a new Endpoints section in the Settings dialog to control exactly which models are exposed to the clients querying the /v1/models API endpoints.
You can now toggle exposure for: All LLM models, Image models, Embedding models, Audio models, and the various virtual routing models (Auto Balanced, Auto Intelligent, Auto Fast, and Fusion).
The backend routes (Proxy, Anthropic, Gemini) have been updated to securely honor these toggles and filter the model catalog before sending it to the client.
Also updated the fallback strategy preview so that the routing strategy test on the dashboard relies on a local preview parameter instead of overriding global configuration.
2. Configured Provider Indicator Added a helpful visual indicator to the "Add Key" dialog. When choosing a platform from the dropdown menu, any provider that already has an API key configured will display a small glowing green dot on the left. This makes it instantly obvious which providers you've already set up without having to cross-check.

Type of change
 New feature (non-breaking change which adds functionality)
 Enhancement (UI/UX improvement)


suantea:
Stale check (2026-09-05): open ~5 weeks with no review, conflicting with upstream main (`proxy.ts`/`router.ts`/FallbackPage have all changed heavily — incl. #1027 quota display and #1105 keyless-model dropping, which touch the same surfaces). The endpoint-exposure control is NOT yet covered upstream, so the idea still has value. Author: please rebase onto main if pursuing, or close to trim the open-PR backlog.

tashfeenahmed:
Closing this one as stale: it is 200+ commits behind with conflicts across proxy, router and the Keys page. The provider key indicator already exists on main (add-key-form shows added platforms via GET /api/keys/providers), and the routing chain rewrite here would reopen the empty-chain bug fixed in #1021. The endpoint exposure toggles are still a good idea and would be welcome as a fresh, small PR that filters the listings only. Thanks for the contribution.

PR #668, opened 2026-07-29
Proposal: Attribute request history to concrete model rows

nicyoong:
# Proposal: Attribute request history to concrete model rows

## Summary

This is a focused follow-up to my earlier [PR #658](https://github.com/tashfeenahmed/freellmapi/pull/658), which addressed the custom-relay model-identity collision in #651.

While #658 was open, the maintainer merged [PR #659](https://github.com/tashfeenahmed/freellmapi/pull/659), which solved the core collision by making custom models unique per normalized endpoint scope. #658 was therefore closed because it conflicted with `main`, not because its approach was rejected. The closing feedback specifically invited two parts of #658 as separate follow-ups:

- attribute request logs, routing statistics, and analytics to `models.id`;
- introduce first-class endpoint records for future per-key model groups (#657).

This PR implements the first item only, adapted to #659's existing `endpoint_scope` design.

## Problem

After #659, two custom relays can correctly have separate `models` rows for the same upstream `model_id`. However, historical and new requests were still identified primarily by `(platform, model_id)`.

That leaves an ambiguity for same-named custom relays. In particular, analytics joins can match a request to both rows, duplicating request counts and pricing. It also keeps routing performance history tied to a derived string key instead of the durable row that routing actually selects.

## Changes

- Add a reversible migration that introduces nullable `requests.model_db_id` with an index.
  - Backfill catalog requests where identity is unambiguous.
  - Backfill custom requests only when the saved key still resolves to the matching normalized endpoint scope.
  - Leave orphaned custom history `NULL` rather than attributing it to the wrong relay.
  - Keep the column intentionally free of a foreign-key constraint so request history survives model deletion.
- Resolve and persist the concrete `models.id` whenever a request is logged.
- Key router statistics, monthly usage, speed-rank writeback, key scoring, and round-robin state by `model_db_id`.
- Join analytics to `models.id`, so duplicated relay model IDs remain separate for pricing and per-model reporting.
- Retain a safe catalog-only fallback for legacy request rows that predate the migration.

## Scope

This PR intentionally does **not** introduce a `custom_endpoints` table or alter the endpoint-scope identity model from #659. That is the separate follow-up suggested for #657, where endpoint records may be useful for per-key model-group metadata.

## Validation

Added regression coverage for:

- logging a request against the specific model row selected for a relay;
- independent scoring for two relays exposing the same upstream model ID;
- independent analytics counts and pricing for those relays;
- migration round-trips and rerunnable catalog migration state.

Ran successfully:

```text
npm run build -w server
npm test -w server -- \
  src/__tests__/routes/custom-endpoint-identity.test.ts \
  src/__tests__/routes/analytics.test.ts \
  src/__tests__/services/router-speed-scoring.test.ts \
  src/__tests__/routes/custom-provider.test.ts \
  src/__tests__/routes/proxy-model-groups.test.ts \
  src/__tests__/routes/routed-via-header.test.ts \
  src/__tests__/services/catalog-sync.test.ts \
  src/__tests__/db/migrate/roundtrip.test.ts
```

The focused run passed 64 tests. The full server suite also passed all affected tests; its only remaining failures are the repository's existing Windows-specific POSIX permission assertions in `db/hardening.test.ts`.

### DIFF
--- a/server/src/__tests__/db/migrate/roundtrip.test.ts
+++ b/server/src/__tests__/db/migrate/roundtrip.test.ts
@@ -22,6 +22,7 @@ const ATTEMPT_ERROR_SUMMARY_FILENAME = '20260726_000006_attempt_error_summary.ts
 const AGENT_COMPATIBILITY_FILENAME = '20260727_000001_agent_compatibility.ts';
 const TOMBSTONE_PROVENANCE_FILENAME = '20260728_000001_tombstone_provenance.ts';
 const CUSTOM_MODEL_ENDPOINT_IDENTITY_FILENAME = '20260729_000001_custom_model_endpoint_identity.ts';
+const REQUEST_MODEL_ATTRIBUTION_FILENAME = '20260730_000001_request_model_attribution.ts';
 
 interface SchemaRow {
   type: string;
@@ -90,6 +91,7 @@ describe('migration round trip', () => {
         AGENT_COMPATIBILITY_FILENAME,
         TOMBSTONE_PROVENANCE_FILENAME,
         CUSTOM_MODEL_ENDPOINT_IDENTITY_FILENAME,
+        REQUEST_MODEL_ATTRIBUTION_FILENAME,
       ]);
     } finally {
       db.close();
--- a/server/src/__tests__/routes/analytics.test.ts
+++ b/server/src/__tests__/routes/analytics.test.ts
@@ -253,6 +253,7 @@ describe('Analytics API', () => {
   function insertRaw(opts: {
     platform?: string;
     modelId?: string;
+    modelDbId?: number | null;
     keyId?: number | null;
     status?: 'success' | 'error';
     inputTokens?: number;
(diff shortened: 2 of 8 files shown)


tashfeenahmed:
Thanks for the careful writeup, and sorry this went unanswered for so long. The analytics half is now fixed on main: #958 gives custom endpoints their own identity, so /by-model groups by platform, endpoint and model_id with an endpoint scoped models join, and the duplicated counts and pricing you describe are gone. The routing half is covered too, since modelStatsKey already keys stats per endpoint and the community reliability priors from #744 are keyed on that string. So I am closing this one as superseded.

There is still one real point in here that main does not solve: request identity is derived at read time from api_keys.base_url, so deleting a key or editing its base_url retroactively reattributes old history. Stamping a nullable requests.model_db_id at write time would freeze it. If you are up for a much smaller PR with just the migration and backfill (no FK, like you had it) plus logRequest stamping the column, and no router changes, I would take that. Note main widened logRequest to keyId: number | null since your base.

PR #670, opened 2026-07-30
fix(i18n): rewrite zh-TW translation

yfdyh000:
Multiple AI tools were used, along with immature human review. The original translation contained numerous machine translation errors.


tashfeenahmed:
Merged, thanks. This was a bigger cleanup than the description suggested. The previous file had "router" as `銑刀`, "lifetime license" as `終身駕照`, "clipboard" as `夾板`, and Sonnet and Haiku as the poem forms, so this is a real improvement rather than a polish pass.

Structure checked out too: key set unchanged, no placeholder drift, validator clean across all 60 locales.

If you ever feel like doing the same for another locale, the model family names (`keys.familyOpus`, `familySonnet`, `familyHaiku`) are translated as poetry in quite a few of them.

PR #672, opened 2026-07-30
cli: publish freellmapi to npm (#671)

### LINKED ISSUE #671: npx freellmapi commands return npm E404 because package is not published
I’m using the FreeLLMAPI Windows desktop application and attempted to connect it to Claude Code using the command documented in the README and Clients & Coding Agents guide:

```powershell
npx freellmapi launch --url http://127.0.0.1:31415
```

The command fails with:

```text
npm error code E404
npm error 404 Not Found - GET https://registry.npmjs.org/freellmapi - Not found
npm error 404
npm error 404 The requested resource 'freellmapi@*' could not be found or you do not have permission to access it.
npm error 404
npm error 404 Note that you can also install from a
npm error 404 tarball, folder, http url, or git url.
```

The same problem would appear to affect the other documented commands, including:

```text
npx freellmapi setup-claude
npx freellmapi setup-codex
npx freellmapi launch-codex
```

### Environment

- OS: Windows 11
- FreeLLMAPI Desktop: v0.5.0
- Node.js: v24.11.1
- npm: v11.6.2
- Installation method: Windows desktop installer

The repository contains a CLI workspace whose `package.json` names the package `freellmapi`, but the package does not appear to be available from the public npm registry.

Expected behavior:

The documented `npx freellmapi` commands should download and run the CLI.

Actual behavior:

npm returns an E404 because it cannot locate the package.

Current workaround:

Claude Code can be launched manually by setting `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, and the relevant model environment variables before running `claude`.

Would it be possible to either:

1. Publish the CLI package to npm;
2. Correct the package name in the documentation if it is published under another name; or
3. Update the documentation and desktop Agents page with working installation/manual configuration instructions?

It would also be helpful for the Windows desktop application’s Agents page to provide a PowerShell-ready manual configuration block when automated CLI setup is unavailable.

tashfeenahmed:
Fixes #671.

`npx freellmapi setup-claude` (and every other documented generator and launcher) fails with npm `E404`. The `cli/` workspace already builds a complete, tested package named `freellmapi` — it has simply never been pushed to the registry. The name is still unregistered.

## What this changes

**`.github/workflows/cli-release.yml`** — publishes the package.

- Runs on pushes to `main` touching `cli/**`, and on `workflow_dispatch` for the first publish.
- Guarded by a registry lookup: if `freellmapi@<version>` already exists, the job skips. Ordinary `cli/src` edits are a no-op; shipping a new version means bumping `version` in `cli/package.json` in the same PR that changes the code. Re-runs are safe.
- Before publishing it packs the tarball, installs it into a scratch project, and runs the binary. A `files` array that drops a runtime import passes unit tests but breaks every consumer, and this is the only step that catches it.
- Publishes with `--provenance`, so npm shows the tarball's link back to this repo and commit.
- If `NPM_TOKEN` is missing the job fails with an explicit message instead of a confusing error from inside `npm publish`.

**`cli/README.md` + `cli/LICENSE`** — the package shipped neither, so its npm page would have rendered blank and the tarball carried no license text. That is a poor look for a package whose entire job is to be trusted enough to `npx`. Both are added to `files`, along with `homepage`, `bugs`, and `keywords`.

**Docs** — the documented command omitted the credential:

```
$ npx freellmapi setup-claude --url http://localhost:3001
freellmapi: No API key supplied. Pass --api-key or set FREELLMAPI_API_KEY.
```

Fixing only the E404 would have walked the reporter straight into this, so `README.md` and `docs/clients.md` now show `--api-key` and the `FREELLMAPI_API_KEY` alternative.

No CLI source changes.

## Verification

Packed the tarball and installed it into a clean project, as a consumer would:

```
$ npx freellmapi --help
FreeLLMAPI coding-agent setup
...
$ npx freellmapi list
claude	Anthropic Messages	root
codex	OpenAI Responses	/v1
...
$ npx freellmapi setup-claude --url http://localhost:3001 --api-key testkey --dry-run
freellmapi: Could not reach the FreeLLMAPI gateway at http://localhost:3001 ...
```

The last one gets past the credential check and stops only at the unreachable local gateway, which is correct. `FREELLMAPI_API_KEY` behaves identically.

Tarball is 12 files / 12 kB. Repo suite green: 44 CLI tests, 64 client tests, i18n across 60 locales, `npm run build` clean.

## Before this can ship

The workflow needs an **`NPM_TOKEN`** repository secret — an npm automation token on an account that will own `freellmapi`. Once it is set, run the workflow manually to publish `0.3.0`; after that it is automatic on version bumps.

Worth noting: the name is unregistered while a public README instructs people to `npx freellmapi`. Anyone can claim it and serve arbitrary code to users following our own docs, so this is worth landing sooner rather than later.

### DIFF
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -8,16 +8,29 @@ name: CLI release
 # Triggers:
 #   - push to main touching cli/**   publish when cli/package.json's version is
 #                                    not on the registry yet; no-op otherwise
-#   - workflow_dispatch              manual run (used for the very first publish)
+#   - workflow_dispatch              manual run
 #
 # The version guard makes this idempotent: ordinary cli/src changes re-run the
 # job, see the version already published, and skip. To ship a new version, bump
 # `version` in cli/package.json in the PR that changes the code.
 #
-# Requires the NPM_TOKEN secret (npm automation token with publish rights).
-# Until it is set, an automatic run warns and skips instead of failing, so
-# merging this workflow does not redden main. A manual run still fails loudly —
-# asking for a publish and silently getting none would be worse.
+# Auth is npm trusted publishing (OIDC) — no NPM_TOKEN, no long-lived secret.
+# npm revoked all classic tokens in December 2025, and granular tokens with
+# "bypass 2FA" are being wound down through 2026 (they lose direct publish
+# around January 2027), so a token-based workflow would have been born
+# deprecated. OIDC mints a short-lived credential per job instead, and npm
+# attaches provenance automatically.
+#
+# One-time setup on npmjs.com (Package settings → Trusted Publisher):
+#   Organization or user: tashfeenahmed
+#   Repository:           freellmapi
+#   Workflow filename:    cli-release.yml
+# Trusted publishing cannot bootstrap a package that does not exist yet — npm
+# has no package settings page to configure until the first version is up — so
+# 0.3.0 was published manually. Every version after it goes through here.
+#
+# Requires npm >= 11.5.1 and Node >= 22.14.0 for OIDC, hence Node 24 below
+# (still inside the repo's engines range of >=20.18 <25).
 
 on:
   workflow_dispatch:
@@ -29,7 +42,7 @@ on:
 
 permissions:
   contents: read
-  id-token: write # npm provenance
+  id-token: write # OIDC credential for trusted publishing
 
 jobs:
   publish:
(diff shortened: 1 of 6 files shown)


tashfeenahmed:
Pushed b159e4f — the token check now skips instead of failing on automatic runs.

As originally written, merging this would have put a red X on main: the push trigger fires, `0.3.0` is unpublished, the token check exits 1. Since the secret can't be added yet, the first consequence of landing the fix would have been a broken main.

A `workflow_dispatch` run still fails loudly — explicitly asking for a publish and silently getting none is worse than an error. The step summary now reports which of the three cases it hit (already published / no token / published).

This makes the PR safe to merge before `NPM_TOKEN` exists.

PR #676, opened 2026-07-30
Fix: routing exhaustion misreported as retry-budget timeout

### LINKED ISSUE #666: Timeout when using Local Ollama through a Custom Provider despite longer upstream timeout
# Timeout when using Local Ollama through a Custom Provider despite longer upstream timeout

## Environment

* **FreeLLMAPI:** Latest version
* **Provider:** Custom (Local Ollama)
* **Model:** `gemma4:e4b`
* **API Client:** OpenAI Python SDK
* **Operating System:** Windows
* **Local LLM:** Ollama

## Problem

I'm using FreeLLMAPI as an OpenAI-compatible proxy in front of a local Ollama instance.

Some requests take longer than expected (which is normal for a local model), but instead of waiting for the configured provider timeout, FreeLLMAPI aborts the request and returns a 502 error.

The dashboard also marks the provider as:

* `penalized`
* `cooldown`
* `failing`

with repeated messages like:

```text
The operation was aborted (custom, chat, 120s)
```

The client ultimately receives:

```text
502 - {
  "error": {
    "message": "All 1 routed attempt(s) failed with upstream provider errors (stopped early: retry time budget 45s exceeded). This is a provider-side failure, not a problem with your request; retry, or check provider status. Attempt trail: custom/gemma4:e4b key1: timeout. Last error: The operation was aborted (custom, chat, 120s)",
    "type": "provider_error",
    "code": "upstream_failed"
  }
}
```

## Observation

The error message appears inconsistent:

* The upstream provider reports a **120-second timeout**.
* However, the router reports:

```text
stopped early: retry time budget 45s exceeded
```

This makes it look like an internal router retry budget is expiring before the configured upstream timeout can complete.

If that's intentional, it would be helpful if this behavior were documented or configurable.

## Expected behavior

If the provider timeout is configured to 120 seconds (or longer), I would expect FreeLLMAPI to wait for that timeout before aborting the request.

Alternatively, if the router has its own retry time budget, it should either:

* be configurable,
* automatically adapt for local providers like Ollama,
* or clearly document that it overrides the provider timeout.

## Questions

1. Is the `retry time budget` configurable?
2. Does it intentionally override the provider timeout?
3. Is there a recommended configuration for local Ollama models that may legitimately require more than 120 seconds for some requests?

Any clarification would be appreciated.

Image

maanas1234:
## Summary
Fixes #666.

The wall-clock retry budget (default 45s, `DEFAULT_FALLBACK_TIME_BUDGET_MS`) was checked *before* checking whether another route/key even existed. With a single custom/Ollama key configured, the one legitimate attempt could legitimately take its full configured `timeoutMs` (e.g. 120s for custom providers). By the time the loop looped around to consider a second attempt, elapsed time already exceeded the 45s budget, so it reported `stopped early: retry time budget 45s exceeded` — even though there was no second key to retry anyway. The real reason was routing exhaustion (no more candidates), not the time budget.

The 120s provider timeout itself was never truncated — this was purely a misleading error message.

## Fix
In `server/src/lib/fallback-loop.ts`, swapped the order: `hooks.route(attempt)` is now checked first, so a routing dead-end reports honestly via `onRoutingExhausted`. The wall-clock budget check now only fires when a route candidate genuinely exists but time is already spent skipping it. Also moved the budget check inside the lease-release `finally` block so no in-flight lease leaks on that early return path.

## Test plan
- [x] `npx vitest run src/__tests__/lib/fallback-loop.test.ts src/__tests__/routes/fallback-hardening.test.ts` — 41/41 pass
- [x] Full suite: 1238/1240 pass (2 pre-existing failures are unrelated Unix file-permission-bit assertions, not reproducible/relevant on this platform)

### DIFF
--- a/server/src/lib/fallback-loop.ts
+++ b/server/src/lib/fallback-loop.ts
@@ -768,17 +768,10 @@ export async function runFallbackLoop(hooks: FallbackHooks): Promise<void> {
       return;
     }
 
-    // Wall-clock budget: refuse to START another retry once spent. The first
-    // attempt always runs; a slow attempt is never aborted mid-flight (that is
-    // the TODO(fallback-v2) hedging work), it just becomes the last one.
-    if (attempt > 0 && budgetMs > 0 && Date.now() - startedAt >= budgetMs) {
-      hooks.onExhausted(
-        exhaustedRetryError(lastError, maxRetries, { attempts, timedOut: true, budgetMs }),
-        { attempts, timedOut: true },
-      );
-      return;
-    }
-
+    // Routing exhaustion (no more keys/candidates) takes priority over the
+    // wall-clock budget below: if there was never another attempt to make,
+    // say so, rather than blaming a budget that nothing was actually cut
+    // short by.
     let route: RouteResult;
     try {
       route = hooks.route(attempt);
@@ -792,11 +785,23 @@ export async function runFallbackLoop(hooks: FallbackHooks): Promise<void> {
 
     // Everything from here to the end of the iteration runs inside a finally that
     // frees the route's in-flight lease. Every exit — success, auth rotation,
-    // retryable continue, fatal, breaker trip, contract violation — passes through
-    // it, so no path can leak a lease and leave the key's concurrency budget short.
-    // Success accounting happens inside dispatch, so the persisted counters are
-    // already written by the time the provisional lease goes away.
+    // retryable continue, fatal, breaker trip, contract violation, wall-clock
+    // budget — passes through it, so no path can leak a lease and leave the
+    // key's concurrency budget short. Success accounting happens inside
+    // dispatch, so the persisted counters are already written by the time the
+    // provisional lease goes away.
     try {
+    // Wall-clock budget: refuse to START another retry once spent. The first
+    // attempt always runs; a slow attempt is never aborted mid-flight (that is
+    // the TODO(fallback-v2) hedging work), it just becomes the last one.
+    if (attempt > 0 && budgetMs > 0 && Date.now() - startedAt >= budgetMs) {
+      hooks.onExhausted(
+        exhaustedRetryError(lastError, maxRetries, { attempts, timedOut: true, budgetMs }),
+        { attempts, timedOut: true },
+      );
+      return;
+    }
+
     let outcome: DispatchOutcome;
     try {
       outcome = await hooks.dispatch(route, attempt);


Arman-Espiar:
Thanks for the explanation.

One thing I'm still unsure about is the `120s` timeout.

I did **not** configure a 120-second timeout anywhere.

In my application, I'm only setting the OpenAI Python SDK timeout:

```python
client = OpenAI(
    ...,
    timeout=600,
)
```

So the client is willing to wait up to 600 seconds.

The error message still reports:

```text
The operation was aborted (custom, chat, 120s)
```

Could you clarify where this 120-second timeout comes from?

* Is it the default timeout for Custom providers?
* Is it configured internally by FreeLLMAPI?
* Or is it coming from the Ollama provider implementation?

I'm trying to understand why the upstream request is being aborted after 120 seconds even though my client timeout is much longer.

maanas1234:
the 120 sec is the default setting in freellm.

suantea:
Nice, minimal fix — reordering so routing exhaustion (no candidates left) is diagnosed before the wall-clock budget matches what an operator actually observes: with a single custom/Ollama key, a legitimate slow attempt (e.g. 120s custom timeout) exhausted the budget, and the error blamed the budget even though nothing was cut short. A few notes:

1. **The reorder is the right call**, and the comment now explains it well. One thing worth double-checking: the budget check still runs when `attempt > 0`, so the first attempt is still never budget-capped — good, that behavior is unchanged.
2. **Tests**: I don't see a test asserting the new precedence (exhaustion-before-budget). Given this is exactly the kind of ordering that silently regresses, a case with a single key whose first attempt is slow + no second candidate, asserting the exhaustion error (not `timedOut: true`), would lock the behavior in.
3. Minor: the diff comment says "never another attempt to make" — for the `groupChain.length === 0` case that's precise; worth confirming the hook also fires correctly when a provider-level failure skips the whole platform (#788), which is the sibling exhaustion path.

Overall the change is small and correct; the missing test is the only thing I'd add before merge.

tashfeenahmed:
Thanks for digging into this and for the clear write-up of the #666 scenario, it was a genuinely confusing error message. Closing because main fixed it another way: 44af06b (#751/#752) changed the loop-top guard to `attempt > 1` so the second attempt always reaches `hooks.route()` and a single-key dead end now reports routing exhaustion rather than the budget, 951712e (#694) rewrote the misleading note, and 1d2226a (#828) made the budget genuinely abort a stalled attempt so `timedOut` is now truthful. If you still see exhaustion misreported as a budget timeout on current main (for example at attempt 2 or later, where the budget check does still run before `route()`), please reopen or push a rebased version with the precedence test suantea suggested and we will take another look.

PR #678, opened 2026-07-30
feat: Sync with upstream/main and integrate Combos routing (Fallback & Round-Robin)

mohammadsaleh40:
### 📌 Summary
This PR synchronizes my `develop` branch with the latest `upstream/main`, resolving all merge conflicts while successfully preserving the newly developed **Combos** feature alongside upstream's recent improvements (such as the refined `resolveStickyPreference` logic and ModelScope provider updates).

### 🚀 Key Changes & Features
- **Combos Feature Integration:** Fully preserved and integrated the named model groups feature, supporting both `fallback` (strict chain) and `round-robin` (rotational) routing strategies.
- **Smart Conflict Resolution in `proxy.ts`:** Carefully merged the upstream's improved `resolveStickyPreference` logic for auto-routed models with the custom Combo resolution logic. The Combo logic now correctly utilizes `resolveRequestedIdToMembers` for unified model groups without breaking upstream's sticky session guarantees.
- **Upstream Enhancements:** Incorporated all recent upstream updates, including ModelScope provider improvements, UI/branding updates, and security/cooldown refinements.
- **Configuration Harmony:** Ensured that local development adjustments (e.g., port configurations) align with upstream's environment variable standards to prevent future conflicts.

### 🛠️ Testing & Verification
- ✅ TypeScript compilation passes without errors (`npx tsc --noEmit`).
- ✅ Verified that Combo routing (both `fallback` and `round-robin`) correctly resolves model IDs and respects the defined order.
- ✅ Confirmed that upstream's auto-routing and sticky session logic remain fully functional and unaffected.
- ✅ No breaking changes introduced to the existing API contract.

### 📝 Notes for Reviewer
The most critical conflict was in `server/src/routes/proxy.ts` regarding the `isAutoModel` block. I resolved this by ensuring the upstream's `resolveStickyPreference` is applied first, followed by the Combo strategy evaluation, guaranteeing that both features coexist harmoniously. 

Thank you for reviewing this PR! Please let me know if any adjustments or further refinements are needed.


tashfeenahmed:
Thanks for all the work on this, and sorry it sat while main moved underneath it. Most of what Combos does has since landed under different names: named ordered model groups are now profile chains (every custom profile shows up as auto:<name> in /v1/models and routes through that chain, with drag to reorder on /models/chat), fusion with a judge is on /models/fusion with a saved panel config, and the transport error fallback is already in error-classify.ts. So I am closing this one as superseded rather than adding a third page for the same idea.

One thing here is genuinely missing on main and I would like to have it: round robin across the models in a chain. Right now we only round robin between keys of a single model. Your round-robin.ts and its test are the right shape. If you want to send a small follow up that adds a round-robin strategy in scoring.ts plus a rotation cursor over the resolved chain in router.ts, with the sticky limit knob on the existing Fallback page, I would merge that happily. Worth a look at #1033 first since it touches the same area.

PR #669, opened 2026-07-30
fix(i18n): improve zh-CN translation quality

yfdyh000:

PR #683, opened 2026-07-31
docs(i18n): pin the zh glossary, land the two agreed zh-CN fixes

tashfeenahmed:
Follow-up to the terminology discussion on #669, which carried on after the merge and mostly resolved itself. This lands the two points both reviewers ended up agreeing on and writes the rest down.

### String changes

- `settings.compressionDescription`: `请求数据量` to `请求大小`. @NirvanaCh7 pointed out the semantic scope shifts: `请求大小` is the size of one request, while `请求数据量` reads as a cumulative total across many, which is the wrong sense here. @yfdyh000 did not object. It also matches zh-TW, which already renders the same English string as `請求大小`.
- `models.providerIdsHint`: `指定经由该提供方` to `会将请求锁定到该提供方`. @yfdyh000 withdrew the original wording once it was clear there is no fallback chain involved. The new wording restores the predicate and matches "pin one provider" in the source.

### docs/translating.md

Two zh passes landed in two days and the same terms came up twice, so this writes down what is settled: `提供方`, the `词元` / `令牌` split, `编程`, `导出`, `撤销`, `试验台`, `权衡` vs `权重`, and the superlatives on the strategy presets.

It also records two things that were not obvious from the files alone:

- `词元` is decided with known dissent. It is the [national standard term](https://termonline.cn) and it disambiguates from `令牌` for credentials, but most Chinese AI products still show `Token` in Latin. Anyone wanting to revisit it should open an issue rather than fold it into another PR.
- zh-TW diverges on purpose (`Token` in Latin, `提供者`, `遊樂場`) and should not be synced to zh-CN term for term.

Plus a short general section on the validator and key parity, linked from CONTRIBUTING.md.

`npm run check:i18n` passes: 60 locales, 725 keys.

Thanks to @yfdyh000 and @NirvanaCh7 for working this through in detail.


vi9ayy911:
HI Tashfeen are u from india?

PR #686, opened 2026-07-31
fix(proxy): inject estimated usage frame when streaming upstream omit…

oppih:
## Problem

Some OpenAI-compatible upstreams (e.g. **OpenCode Zen**) never echo a final usage frame in streaming responses, even when the client requests `stream_options.include_usage`. FreeLLMAPI's streaming path only passes through upstream usage frames, so for those upstreams the client receives a stream with **no usage block at all**.

Strict OpenAI-compatible clients treat a missing usage block as "no accounting happened": **Hermes Agent** (and similar) skips per-call token/cost/`billing_provider` writes entirely when `response.usage` is absent. Verified end-to-end:

- Hermes → FreeLLMAPI (streaming, `deepseek-v4-flash-free` routed to OpenCode Zen) → `billing_provider=NULL`, `api_call_count=0`, `input_tokens=0` in every session, while the same request routed to providers that *do* echo usage (Gemini, Ling) accounts perfectly.
- Direct probe: 52 stream chunks from the opencode route, 0 usage frames; the non-streaming path on the same route *does* return real usage.

## Fix

At stream end, when the upstream never echoed a usage frame **and** the client requested `stream_options.include_usage`, inject the proxy's own estimate — the same `estimatedInputTokens` / `totalOutputTokens` values the non-stream path already reports — as a standard usage-only chunk (`choices: []`), placed after the finish chunk and before `[DONE]`.

Behavior is fully gated:

- Upstreams that echo usage keep their **real** frame, unchanged (captured from any frame position — usage-only or content-attached — so a mid-stream real block also suppresses the estimate).
- Clients that never request usage see no extra frames.
- Only `include_usage=true` + usage-less upstream triggers the injected estimate.

> **Note:** the injected frame is an **estimate**, not the upstream's final accounting. The OpenAI wire format has no standard flag for estimated usage, and adding a custom field could break strict clients that reject unknown usage keys — the very clients this injection exists for. Token counts use the same length/4 heuristics as the proxy's non-streaming estimate path.

## Tests

`proxy-stream-usage.test.ts` covers:

- usage-less upstream + `include_usage=true` → estimated usage frame injected (positive token counts, empty choices)
- usage-less upstream + no `include_usage` → no frame injected
- upstream with usage → upstream frame passed through unchanged
- usage-only frame emitted mid-stream → real frame is the only usage frame, no estimate
- usage attached to a content frame (non-final position) → real usage preserved, no estimate


tashfeenahmed:
Heads up: #845 just merged, so real upstream usage frames now flow through when the provider reports them. This PR is still valuable for providers that stay silent, but it will need a rebase and can narrow to the inject when absent case, since the stream_options schema handling now exists on main.

suantea:
Good catch — OpenCode Zen (and similar upstreams) skipping the final usage frame leaves strict-mode clients hanging. A few observations from reading the diff:

1. **Estimated-frame shape**: injecting a usage block with estimated token counts is the pragmatic fix, but please make the estimate explicitly marked as such if the wire format allows it (or document it in the PR), so a client doing cost accounting doesn't mistake an estimate for the upstream's final accounting.
2. **When to inject**: the diff gates on "stream ended without a usage frame". Worth confirming the estimate only fires when `stream_options.include_usage` was actually requested — injecting an unsolicited usage frame could surprise clients that didn't opt in.
3. **Tests**: the new `proxy-stream-usage.test.ts` covers no-usage and with-usage upstreams — nice. I'd add a third case: upstream that emits usage in a non-final position (mid-stream), asserting the estimate is NOT injected on top of a real block.

Implementation looks clean otherwise — happy to help validate against a live OpenCode Zen stream once it merges.

oppih:
Thanks for the review on both counts.

**Rebase / narrowing (re: #845):** Rebased on latest `main`. Since #845 now owns the `stream_options` schema parsing and upstream forwarding, this PR is narrowed to exactly the "inject when absent" case: at stream end, if the upstream never emitted a usage frame AND the client requested `stream_options.include_usage`, the proxy injects its own estimate. Providers that echo usage are untouched (real frame passes through), and clients that never opt in see no extra frame.

**On the three review points:**

1. **Marking the estimate** — I considered a custom flag, but the OpenAI wire format has no standard field for estimated usage, and adding a non-standard key risks exactly the strict clients this fix exists for (some reject unknown usage keys outright). So it's documented instead: the PR description and a code comment both state the injected frame is an estimate built from the same length/4 heuristics the non-streaming path already reports. Happy to add a marker field if you'd prefer it on the wire — just say the word.

2. **When it fires** — confirmed: injection is gated on `parsed.data.stream_options?.include_usage` only. No opt-in, no frame.

3. **Mid-stream usage** — added two tests: (a) usage-only frame emitted mid-stream (before the finish chunk), (b) usage attached to a content frame in a non-final position. Both assert the real usage is the only usage frame and no estimate is injected. Writing test (b) exposed a real gap in the original code: usage was only captured inside the `!choice` branch, so usage riding on a content frame was dropped and the estimate would have fired on top of a real block. The capture now happens for any frame carrying `usage`, with `choices: []` forced on re-emit so content is never duplicated.

All 5 tests in `proxy-stream-usage.test.ts` pass, plus the 15-test `proxy-stream-integrity` suite, `tsc` clean. The new commit is `f5db1bd`.

suantea:
Thanks for the thorough follow-up — all three points landed well:

1. **Estimate marking**: the documented-not-on-wire call is reasonable; the strict-client concern is real and a non-standard key would be worse. Agreed.
2. **Opt-in gating**: confirmed  only — good.
3. **Mid-stream usage**: excellent catch — the usage-on-content-frame gap (dropped because capture only happened in the  branch) is exactly the kind of edge the review was fishing for, and forcing  on re-emit is the right fix. The two new tests lock it in.

Rebase narrowing against #845's stream_options ownership also reads clean. Happy with this as-is.

PR #692, opened 2026-08-01
fix(keys): carry base_url for custom endpoints through export/import

### LINKED ISSUE #687: [Bug] Export of keys does not contain custom keys
When I export all my keys and then import them I want all of them.

What needs to be updated:
`.env` export, `.csv` export and `.json` export -> Add the `base_url` for custom keys.

Image

-> "Will export 40 keys" -> untrue, as the custom keys are skipped it could be less/none.

suantea:
## What & why

Fixes #687. Custom OpenAI-compatible endpoints were dropped from key export and skipped on import:

- `.env` / `.csv` exports had no `base_url` field, and JSON import ignored the `baseUrl` the export already wrote.
- Both import routes (`/api/keys/import` and `/api/keys/import-selected`) rejected the `custom` platform outright, so an export → import round-trip silently lost every custom provider — the "Will export N keys" count was also misleading.

## Changes

- **`.env` export** — custom endpoints now emit a numbered pair `CUSTOM_BASE_URL_<n>` + `CUSTOM_KEY_<n>` so the upstream URL survives.
- **`.csv` export** — header gains an optional `base_url` column (4th), populated for custom keys.
- **`.json` export** — already carried `baseUrl`; the import path now honors it.
- **Import routes** — `/import` and `/import-selected` restore custom endpoints: normalized `base_url` stored on the `api_keys` row, `'no-key'` sentinel kept for auth-less local servers. Keys without a base URL are still skipped with a clear reason.
- **Export filter** — `no-key` custom rows (local auth-less servers) are no longer dropped from exports, since the endpoint is the valuable part.
- Parser (`key-parser.ts`) threads `baseUrl` through `ParsedKey`, CSV 4th column, and the `.env` pairing convention.

## Tests

- `key-parser.test.ts`: export JSON carries `baseUrl`; CSV 4th column; `.env` `CUSTOM_KEY`/`CUSTOM_BASE_URL` pairing.
- New `keys-export-import-custom.test.ts`: round-trip for all three formats (export → wipe → import → row restored with `base_url`), plus `import-selected` happy path and rejection without a base URL.

Verified with `npm run test -w server` (1756 passing; 3 pre-existing environment failures unrelated to this change — 2 file-permission checks on Windows, 1 flaky perf assertion).

### DIFF
--- a/client/src/components/keys/import-keys-section.tsx
+++ b/client/src/components/keys/import-keys-section.tsx
@@ -85,6 +85,7 @@ export function ImportKeysSection({ onImported }: { onImported?: () => void } = 
       keyName: row.keyName,
       keyValue: row.keyValue,
       platform: row.platform,
+      baseUrl: row.baseUrl,
     }))
 
   function updateRow(index: number, patch: Partial<ImportRow>) {
--- a/server/src/__tests__/lib/key-parser.test.ts
+++ b/server/src/__tests__/lib/key-parser.test.ts
@@ -111,6 +111,27 @@ describe('key parser', () => {
     expect(result!.skipped).toHaveLength(0);
   });
 
+  it('carries baseUrl through export JSON parsing (#687)', () => {
+    const exportJson = JSON.stringify({
+      version: 1,
+      exportedAt: '2026-07-06T12:00:00Z',
+      source: 'freellmapi',
+      keys: [
+        { platform: 'custom', key: 'sk-relay-123', label: 'My Relay', baseUrl: 'https://relay.example.com/v1' },
+      ],
+    });
+    const result = parseExportJson(exportJson);
+    expect(result).not.toBeNull();
+    expect(result!.keys).toEqual([
+      {
+        rawKey: 'My Relay=sk-relay-123',
+        prefix: 'CUSTOM_',
+        platform: 'custom',
+        baseUrl: 'https://relay.example.com/v1',
+      },
+    ]);
+  });
+
   it('returns null for non-export JSON', () => {
     expect(parseExportJson('{"foo":"bar"}')).toBeNull();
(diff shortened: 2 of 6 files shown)


tashfeenahmed:
Thank you for this, and sorry for the collision — #717 landed the export side and the `POST /api/keys/import` route a few hours before I got to this, so the overlapping part is already on main and the branch no longer merges.

Two things worth recording, because your PR is the reason I found them.

**You were right that the round-trip was still broken, and further than #717 went.** You wired `baseUrl` through `importKeySchema`, the `/preview` response and the client payload. That path genuinely was untouched — the dashboard never calls `/api/keys/import`, it calls `/preview` → `/import-selected`, so #717's fix isn't reachable from the UI at all.

**But that wiring alone doesn't finish the job.** I applied your changes to current main, rebuilt, and drove the real Add-key dialog: the preview row still shows an empty "Choose provider", the checkbox still can't be ticked, and the button still reads "Import 0", disabled. `custom` isn't in `importablePlatforms` at all (it lives in `CUSTOM_GROUP`, deliberately kept out of that dropdown), so `platformFromPreview` returns `''`, the row can never be selected, and the request is never sent. The base URL you carefully threaded through never leaves the browser.

With that one extra piece added, the same flow gives "Imported 1; failed 0." and the endpoint comes back with its `base_url` intact.

I'm opening a follow-up PR with that, built on main's `resolveCustomEndpointKey` (endpoint pooling, #619) and the `assessProviderUrl` SSRF check that `/import` already applies — an import file can name `169.254.169.254` just as easily as a manual add can, and the raw `INSERT` here would also duplicate a row on re-import.

Closing this as superseded, but the gap was yours to find. Credited in the README.

PR #693, opened 2026-08-01
fix(csp): only emit upgrade-insecure-requests for HTTPS requests

### LINKED ISSUE #682: Production CSP upgrade-insecure-requests breaks dashboard on plain HTTP (LAN / IP access)
### Summary

v0.6.6 introduced stricter CSP via Helmet [](https://github.com/tashfeenahmed/freellmapi/pull/498). The new policy includes `upgrade-insecure-requests`, which breaks the dashboard when accessed over plain HTTP (e.g. LAN IP). **v0.6.5 and earlier work fine.**

### Description

Production responses now include:

```http
Content-Security-Policy: ...; upgrade-insecure-requests

When opening http://<lan-ip>:port, the document loads over HTTP, but the browser upgrades /assets/*.js, /assets/*.css, and favicon.svg to HTTPS. The server has no TLS on that port → net::ERR_SSL_PROTOCOL_ERROR → blank page.
Asset paths in HTML are relative (/assets/...), so the failure is entirely due to the CSP directive.
Steps to reproduce

Run ghcr.io/tashfeenahmed/freellmapi:v0.6.6 (or :latest) with HOST_BIND=0.0.0.0.
Open http://<lan-ip>:<port> from another device.
Blank page; DevTools Network shows document 200/304, assets fail with ERR_SSL_PROTOCOL_ERROR.

Expected
Same as v0.6.5: dashboard loads over intentional plain HTTP on a trusted LAN (as documented with HOST_BIND=0.0.0.0).
Actual
Blank page on any non-HTTPS origin.
Environment

Broken: v0.6.6 / :latest
Working: v0.6.5
Access: HTTP only, LAN IP (Synology Docker)
Browser: Chrome (also in incognito)

Workarounds

Pin image to ghcr.io/tashfeenahmed/freellmapi:v0.6.5
Put HTTPS reverse proxy in front
Strip/rewrite CSP at the proxy to remove upgrade-insecure-requests

Suggested fix

Only add upgrade-insecure-requests when the request is already HTTPS (or X-Forwarded-Proto: https), or
Provide an env flag (e.g. CSP_UPGRADE_INSECURE_REQUESTS=false) for HTTP/LAN installs, or
Document that plain-HTTP LAN access requires a TLS terminator after v0.6.6

This is a regression from the security hardening in #498. Happy to test a fix.

suantea:
## What & why

Fixes #682. helmet's default CSP ships `upgrade-insecure-requests`, which tells the browser to fetch every subresource over HTTPS. On plain-HTTP LAN installs (`HOST_BIND=0.0.0.0`, no TLS terminator) that blanks the dashboard: the document loads over HTTP, `/assets/*` get upgraded to HTTPS, the server has no TLS on that port, and every asset fails with `ERR_SSL_PROTOCOL_ERROR`.

## Changes

- Disable `upgrade-insecure-requests` in the helmet defaults.
- Re-add it via middleware only when the request already arrived over HTTPS (direct TLS, or `X-Forwarded-Proto: https` behind a reverse proxy). Trusting that header is harmless — a spoofed value can only make the caller's own browser demand HTTPS for that one response.
- Tests: plain HTTP omits the directive; `X-Forwarded-Proto: https` keeps it.

Verified with `npx vitest run src/__tests__/routes/csp-security.test.ts` (7/7 passing).

### DIFF
--- a/client/src/components/keys/import-keys-section.tsx
+++ b/client/src/components/keys/import-keys-section.tsx
@@ -85,6 +85,7 @@ export function ImportKeysSection({ onImported }: { onImported?: () => void } = 
       keyName: row.keyName,
       keyValue: row.keyValue,
       platform: row.platform,
+      baseUrl: row.baseUrl,
     }))
 
   function updateRow(index: number, patch: Partial<ImportRow>) {
--- a/server/src/__tests__/lib/key-parser.test.ts
+++ b/server/src/__tests__/lib/key-parser.test.ts
@@ -111,6 +111,27 @@ describe('key parser', () => {
     expect(result!.skipped).toHaveLength(0);
   });
 
+  it('carries baseUrl through export JSON parsing (#687)', () => {
+    const exportJson = JSON.stringify({
+      version: 1,
+      exportedAt: '2026-07-06T12:00:00Z',
+      source: 'freellmapi',
+      keys: [
+        { platform: 'custom', key: 'sk-relay-123', label: 'My Relay', baseUrl: 'https://relay.example.com/v1' },
+      ],
+    });
+    const result = parseExportJson(exportJson);
+    expect(result).not.toBeNull();
+    expect(result!.keys).toEqual([
+      {
+        rawKey: 'My Relay=sk-relay-123',
+        prefix: 'CUSTOM_',
+        platform: 'custom',
+        baseUrl: 'https://relay.example.com/v1',
+      },
+    ]);
+  });
+
   it('returns null for non-export JSON', () => {
     expect(parseExportJson('{"foo":"bar"}')).toBeNull();
     expect(parseExportJson('[1,2,3]')).toBeNull();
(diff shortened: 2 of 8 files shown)


tashfeenahmed:
Thanks — this is the same CSP fix as #699, which I've just merged, so I'm closing this one as a duplicate rather than because anything is wrong with it.

For future reference: this branch also carries the #687 key export/import work, which made it a larger diff than the CSP change it was named for. #699 was CSP-only, which is why it went in.

I verified the CSP bug and the fix before merging: served on `0.0.0.0` and loaded from another device over a LAN IP, main renders a blank page with `/assets/*.css`, `/assets/*.js` and `/favicon.svg` all failing `ERR_SSL_PROTOCOL_ERROR` — exactly the three the reporter listed. With the fix the dashboard renders and the directive is still emitted behind `X-Forwarded-Proto: https`. Full server suite green on the merge.

PR #719, opened 2026-08-02
fix: harden freellmapi fallback routing

sunjcp:
Add breaker, latency, fallback logging, provider coverage guards, cloud/free model cleanup, and dashboard reliability wiring. Disable deterministic non-routable catalog rows so auto routing skips paid-only or native-Hermes-only paths.


tashfeenahmed:
Thank you for putting this much work in, and sorry for the slow reply. There is real thought in here. The platform level breaker in particular is something the router genuinely lacks: today there is a per model 429 penalty, but nothing that pulls an entire platform out of rotation when it starts failing as a whole. I would like that.

The problem is the base. This branch forked on 14 June, and main has moved a long way since:

- It recreates `server/src/db/migrations.ts`. That file was deleted on 24 June when migrations moved to a file per migration system under `server/src/db/migrations/` with a registry in `server/src/db/migrate/defaults.ts`. Merging this would undo that refactor.
- It conflicts in more than ten files, including the baseline migration, `proxy.ts`, `fallback.ts`, `catalog-sync.ts` and `index.ts`.
- The branch has 5 locale files; main has 60, with a validator that enforces key parity. The 26 new keys are added to `en.json` only, so `npm run check:i18n` would fail as soon as CI ran.
- It also adds `better-sqlite3` and a `predev` rebuild hook at the repo root, which is not needed since the server workspace already depends on it.

I do not think a rebase is realistic here. It would be closer to a rewrite than a merge, and you would be reconciling six features at once against two months of drift.

What I would rather do is take it in pieces, starting with the strongest one. Would you be up for opening a fresh branch off current main with just the circuit breaker? Roughly:

- `server/src/services/breaker.ts` plus its tests
- the call sites in the proxy path that record success and failure
- no dashboard page, no new dependencies, no migration

That is a reviewable PR I can actually land, and the persistence design you wrote up in the module comment is worth keeping as is. Once it is in, the latency monitor and the fallback event log are natural follow ups, and the Reliability page makes much more sense on top of data that already exists.

Two things to know before you start, so the same drift does not bite twice. Model availability data ships through the signed catalog rather than through migrations, so the catalog row changes in this branch would need to go a different route. And any new user facing string has to be added to all 60 locale files, which `npm run check:i18n` from `client/` will verify for you.

Closing this one as superseded rather than rejected. The breaker idea is welcome and I will review that PR quickly.

PR #722, opened 2026-08-03
fix(models): disable Cloudflare Kimi models that left the free tier

### LINKED ISSUE #713: [Model Update] Cloudflare Kimi K2.6 and K2.7 Code now require a Workers Paid plan — remove from catalog
Both models moved off the free tier on Cloudflare Workers AI, but they are still
published in the catalog (applied 2026.08.01) as `enabled: true`, so every
request fails and burns a fallback slot.

**Models affected**
- `@cf/moonshotai/kimi-k2.6`
- `@cf/moonshotai/kimi-k2.7-code`

**Live evidence (tested 2026-08-02 with a Workers AI account key)**
Both return `403` on a minimal request:

AiError: Model @cf/moonshotai/kimi-k2.6 is not available on the Workers Free plan:
This model requires a Workers Paid plan.
See https://developers.cloudflare.com/workers-ai/platform/pricing/ for details. (code 5035)

(same message for `@cf/moonshotai/kimi-k2.7-code`)

**Impact on self-hosted installs**
The monthly (free) catalog still lists them as available, so free-tier users see
them in the model list and every routed request 403s. (We are on the latest
version with the #647 disabled-model fix — the model is simply still present in
the catalog, not a routing bug.)

**Suggested fix**
- Remove both entries from the catalog (or publish them `enabled: false`).
- If the same availability applies to the Hugging Face entries
  (`moonshotai/Kimi-K2.6`), please check those too.

Happy to verify the HF entries as well if that helps.

suantea:
## What

Cloudflare moved `@cf/moonshotai/kimi-k2.6` and `@cf/moonshotai/kimi-k2.7-code` onto the **Workers Paid plan**, but the monthly catalog snapshot still lists them enabled. Every free-tier request now returns 403 (`error code 5035`) and burns a fallback slot.

## Fix

New bundled migration `20260803_000001_disable_paid_cloudflare_kimi.ts` sets both rows `enabled = 0`:

- catalog-sync's rule is **local disable wins** (`enabled = catalog_enabled ? local_enabled : 0`), so a stale catalog that still lists them enabled cannot re-enable these rows;
- rows are kept (not deleted), so an operator can still flip them back on if Cloudflare ever restores the free tier;
- `down()` re-enables both, and the migration is idempotent like the rest of the suite.

## Files

- `server/src/db/migrations/20260803_000001_disable_paid_cloudflare_kimi.ts` — new migration
- `server/src/db/migrate/defaults.ts` — register the migration
- `server/src/__tests__/db/migrate/roundtrip.test.ts` — expected migration list

## Verification

`vitest` roundtrip + idempotency suites pass (14 tests).

Closes #713

### DIFF
--- a/server/src/__tests__/db/migrate/roundtrip.test.ts
+++ b/server/src/__tests__/db/migrate/roundtrip.test.ts
@@ -23,6 +23,7 @@ const AGENT_COMPATIBILITY_FILENAME = '20260727_000001_agent_compatibility.ts';
 const TOMBSTONE_PROVENANCE_FILENAME = '20260728_000001_tombstone_provenance.ts';
 const CUSTOM_MODEL_ENDPOINT_IDENTITY_FILENAME = '20260729_000001_custom_model_endpoint_identity.ts';
 const CUSTOM_ENDPOINT_HOST_LABELS_FILENAME = '20260802_000001_custom_endpoint_host_labels.ts';
+const DISABLE_PAID_CLOUDFLARE_KIMI_FILENAME = '20260803_000001_disable_paid_cloudflare_kimi.ts';
 
 interface SchemaRow {
   type: string;
@@ -92,6 +93,7 @@ describe('migration round trip', () => {
         TOMBSTONE_PROVENANCE_FILENAME,
         CUSTOM_MODEL_ENDPOINT_IDENTITY_FILENAME,
         CUSTOM_ENDPOINT_HOST_LABELS_FILENAME,
+        DISABLE_PAID_CLOUDFLARE_KIMI_FILENAME,
       ]);
     } finally {
       db.close();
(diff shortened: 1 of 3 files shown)


tashfeenahmed:
Thank you for this, and for the careful write up. The diagnosis is correct: I checked Cloudflare's own model pages and both `@cf/moonshotai/kimi-k2.6` and `@cf/moonshotai/kimi-k2.7-code` now say "This model is not available on the Workers Free plan." Your reading of the catalog sync rule is right too. So the research here was sound and it is appreciated.

The one thing to change is where the fix lives. Model availability ships through the signed catalog rather than through freeapi migrations, so that a correction reaches existing installs instead of only fresh databases. A migration also runs only once, which means an install that had not yet synced `kimi-k2.7-code` would pick it up enabled straight afterwards, and that model is not in the bundled baseline at all, only `kimi-k2.6` is.

So this is now fixed in the catalog instead:

- Live catalog `v2026.08.03.09411bb0` has both models disabled, and premium installs pick it up on their next sync. Confirmed on a live install already.
- The free catalog was rebuilt at the same time, and both models are gone from it now rather than in 30 days. The 30 day window only delays new models arriving, not removals.

While in there I checked all 33 Cloudflare entries in the catalog, 28 chat models plus embeddings and transcription. These two are the only ones that moved to the paid plan, so nothing else needs the same treatment.

Closing this one since the fix has shipped, with thanks. The catalog fix would not have happened this quickly without your report, and the same goes for #713. Please do keep them coming.

Issue #315, opened 2026-06-13
Help wanted: review & extend dashboard translations (zh-CN, fr, es, pt-BR)


tashfeenahmed:
PR #314 adds a 5-language dashboard (English, 简体中文, Français, Español, Português-BR) using a zero-dependency JSON i18n system. **UI coverage is now complete** — every user-facing string across the dashboard is wired (289 keys per locale, parity-checked). What's left is **native-speaker review of translation quality**.

Foundation credits: #307 (@dashitongzhi), #220 (@QingJ01), #262 (@3215).

## How to contribute
Translations live in `client/src/i18n/locales/<locale>.json` — same dot-notation keys as `en.json`. To refine wording, edit the value for a key and open a PR. Keep `{placeholders}` intact. To add a new language, copy `en.json` to `<bcp47>.json`, translate, and register it in `client/src/i18n/I18nProvider.tsx` (`SUPPORTED_LOCALES` + `dictionaries`) plus the `languages` section of every locale file.

## Native-speaker review wanted
Skim your language for accuracy, tone, and naturalness, and open a PR with fixes. The fr/es/pt-BR sets and the new zh-CN keys were an initial pass:
- [ ] **简体中文 (zh-CN)**
- [ ] **Français (fr)**
- [ ] **Español (es)** — neutral / Latin-American tone preferred
- [ ] **Português (Brasil) (pt-BR)**

A few notes for reviewers:
- Some short sentences are split into `*Before` / `*After` key pairs because the UI embeds an inline `<code>` element in the middle (e.g. `keys.unifiedKeyDescBefore` + `keys.unifiedKeyDescAfter` around `api_key`). Keep the two halves grammatical around the code term in your language.
- Numeric labels use `{count}` / `{tokens}` / `{platforms}` etc. — leave the braces and names exactly as in `en.json`.

## Housekeeping
PR #314 supersedes the earlier Chinese-only i18n PRs (#220, #262, #307) by generalizing the same foundation to 5 languages with full coverage; those can be closed once #314 lands. (#244, the Chinese README, is a separate docs track.)

suantea:
作为 zh-CN 母语者做了一轮完整审校，结论：**zh-CN 翻译质量良好，无需修改**。

审校范围与方法（对比 en.json 逐键检查）：
- **键完整性**：797/797 全部齐全，无缺失、无多余
- **标点规范**：无中英文括号混用、无全角/半角混排问题
[...]

PR #723, opened 2026-08-03
fix(i18n): polish zh-CN strings for punctuation and pronoun consistency

suantea:
## What

Native-speaker review pass over the simplified-Chinese dashboard locale (issue #315), addressing the three concrete quality problems found in `zh-CN.json`:

1. **Untranslated toggle label** — `nav.switchToEnglish` was still "Switch to English"; every other locale translates it (zh-TW: 切換為英文).
2. **Half-width punctuation inside Chinese sentences** — half-width commas/colons/parens (`自动(回退链)`, `上次检查:{when}`, `,无需手动操作`) replaced with full-width forms（，：（））. Brand/technical tokens (RPM, Base URL, `model: "auto"`) are untouched.
3. **Inconsistent second-person pronoun** — mixed 你/您 across 19 keys unified to the polite 您, matching the existing tone (`您的资源池`, `接入您的应用`).

## Verification

- `node scripts/check-i18n.mjs` → **i18n validation passed for 60 locales and 749 keys**
- No key added/removed/renamed; only string values changed; placeholders `{...}` preserved exactly.

## Files

- `client/src/i18n/locales/zh-CN.json`

Refs #315


tashfeenahmed:
Merged in #740, with your commit carried across so the authorship stays yours. This PR was branched off #722, which ended up closed in favour of a catalog side fix, so rebasing it here would have dragged that migration along with it. Splitting the i18n commit out was the cleaner path.

This was careful work and it is appreciated. I checked the three claims rather than taking them on trust, and after the patch there are zero remaining 你 and zero half-width , : ; sitting against a CJK character, so it is a complete pass rather than a partial one. Leaving the technical tokens and the Latin-in-Chinese spacing alone was exactly right. You also quietly fixed something the description does not mention: analytics.savingsHintExact was switching between 你 and 您 inside a single sentence.

One change on top of yours. nav.switchToEnglish went in as 切换到英语, and I made it 切换到英文, since 语 is the spoken tongue and the sibling label is 中文, so both halves of the toggle should read as 文. Your own description cites zh-TW's 切換為英文, which is the same reasoning.

I also added two rows to the terminology table in docs/translating.md pinning the pronoun and the language label. Neither was written down anywhere, which is how the file drifted into mixed register in the first place, and that table exists precisely so the next pass does not relitigate it.

One thing I deliberately left alone: common.copyFailed uses a spaced dash, which looks wrong next to the full-width standard you are enforcing. auth.forgotPasswordDescription already uses the same spaced form in zh-CN and zh-TW though, and ja and ko use it throughout, so fixing only that one string would have created a fresh inconsistency. If you fancy a follow up, sweeping the spaced dash to —— across the Chinese locales would be a good one, and you clearly have the eye for it.

Thank you again.

Issue #434, opened 2026-07-01
docs+scripts: contributor bootstrap on Windows (PowerShell) — README is Bash-only, .exe is user-only


chirag127:
## Summary

The README's **"Local development"** section (`npm install && npm run dev`) is Bash-first: `openssl rand -hex 32`, `printf`, `cp .env.example .env`. On Windows without WSL/Git Bash, none of those work. The result is that Windows contributors either:

1. Install WSL just to bootstrap (huge overhead for a Node.js project), OR
2. Struggle through translating each command to PowerShell, OR
3. Give up and use the `.exe` installer — which is fine for **users** but blocks **contributors** from actually developing against the code

## Concrete pain points (verified 2026-07-01, Windows 11, PowerShell 7.5)

| README command | PowerShell equivalent | Notes |
|---|---|---|
| `openssl rand -hex 32` | `-join ((1..32) \| % { '{0:x2}' -f (Get-Random -Max 256) })` | `openssl` not on default Windows PATH |
| `printf "ENCRYPTION_KEY=%s\n..." > .env` | `"ENCRYPTION_KEY=..." \| Set-Content .env` | `printf` is Bash-only |
| `cp .env.example .env` | `Copy-Item .env.example .env` | `cp` alias exists but the flag syntax differs |
| `curl -fsSL .../install.sh \| bash` | (not applicable) | Docker one-liner is Linux/macOS only |

The README's Windows section says "the easiest path is the desktop `.exe` installer" — great for end users, but a contributor bootstrap gap.

## Proposal — ship a PowerShell dev-bootstrap script

Add `scripts/dev-bootstrap.ps1` (and matching `scripts/dev-bootstrap.sh` for parity) that:

1. Checks `node --version` meets minimum (see companion engines-pin issue)
2. Runs `npm install` if `node_modules` missing or lockfile changed
3. Generates a fresh `ENCRYPTION_KEY` via Node itself (portable): `node -e "console.log(require('crypto').randomBytes(32).toString('hex'))"` — README already uses this pattern in the fallback line, just standardise on it
4. Writes `.env` from `.env.example` if `.env` doesn't exist
5. Prints "run `npm run dev`" (doesn't auto-launch — respects user's shell/multiplex choice)

Idempotent. Zero external deps beyond Node itself. Works on PowerShell 5.1+ and PowerShell 7+.

### Bonus — autostart doc

[...]

PR #724, opened 2026-08-03
docs+scripts: add a contributor dev-bootstrap script

suantea:
## What

Adds `scripts/dev-bootstrap.sh` — an idempotent one-shot bootstrap for contributors building from source (issue #434). The README/docs development steps are Bash-first and Windows contributors have no scripted path; this gives POSIX shells a single command that:

1. Verifies Node meets the engines range (`>=20.18.0 <25.0.0`).
2. Runs `npm install` only when `node_modules` is missing or `package-lock.json` changed since the last install.
3. Creates `.env` from `.env.example` with a fresh `ENCRYPTION_KEY` when `.env` is missing — never touches an existing `.env`.
4. Prints the next step without auto-launching the dev servers.

README and `docs/install.md` now point contributors at it. (A `dev-bootstrap.ps1` Windows companion is planned separately.)

## Verification

- `bash -n scripts/dev-bootstrap.sh` passes.
- Manual walk-through of each branch: version check, lockfile-triggered install, `.env` generation with `node -e "...randomBytes(32)..."` (the README's own portable pattern).

## Files

- `scripts/dev-bootstrap.sh` (new)
- `README.md`, `docs/install.md` — point at the script

Refs #434

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ Contributors very welcome! See [CONTRIBUTING.md](CONTRIBUTING.md) for the dev lo
 - **Dashboard polish** — charts on the Analytics page, key rotation UX, batch import of keys from `.env`.
 - **Docs** — more examples, client library snippets for Go/Rust/etc., a deployment recipe for Docker or Fly.
 
-`npm install && npm run dev` gets you the server on :3001 and the dashboard on :5173, both with HMR. PRs should include a test, keep the existing suite green (`npm test`), and match the `.editorconfig` / tsconfig defaults already in the repo. Database migration workflow and the full contributor loop are in [CONTRIBUTING.md](./CONTRIBUTING.md).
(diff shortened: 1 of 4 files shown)


suantea:
Closing as superseded — the two scripts this PR adds are already on `main`:

- `scripts/dev-bootstrap.sh` and `scripts/dev-bootstrap.ps1` both exist on current main (they landed via the earlier contributor bootstrap work, #661/#434 etc.).
- The only remaining unique content in this branch is the README one-liner and the `docs/install.md` tip block — and that tip is now stale: it says "a `dev-bootstrap.ps1` companion is planned", but the `.ps1` already exists.

The idea was right and is now shipped; the branch just can't merge cleanly against main anymore. If the README/docs pointer is still wanted, it's a tiny standalone docs PR (and should drop the "planned" wording). Happy to re-submit it that way.

suantea:
Superseded: both scripts already on main. See comment above.

Issue #685, opened 2026-07-31
[Feature Request] Custom provider - fetch models + extra details


Aldo-f:
When adding a custom provider and hitting the 'Fetch models' button it would be nice to see more info about the models.
Like: is (most likely) free, (editable) intelligence, ...

As I would like to be remember that - although this is called FreeLLMAPI - when adding a custom endpoint there is a possibility that i can add a model that isn't free to use.

<img width="604" height="473" alt="Image" src="https://github.com/user-attachments/assets/5d6f3a58-f238-4e80-b531-4a59c9a1779d" />

PR #743, opened 2026-08-04
feat(telemetry): opt-in anonymized reliability uploader

suantea:
## What

Issue #685 design B1: self-hosted instances have no cross-instance quality signal, so a brand-new model starts blind until local traffic accumulates. This adds an **opt-in** background job that publishes a tiny, anonymized summary of what the gateway observed — the aggregate the community-reliability prior (separate PR, B3) is meant to be seeded from.

**Privacy boundary (off by default):**
- only `(platform, model_id)` + raw success/failure counts + mean latency leave the box — no keys, request bodies, client IPs, or custom base URLs;
- a single configured endpoint; nothing else ever goes out;
- a failed upload logs once and waits for the next daily tick — it never crashes the scheduler.

## Implementation

- `server/src/services/telemetry.ts` — `collectTelemetryStats` (7-day window, same source the bandit scores from) + `uploadTelemetry` (POST, 15s timeout) + `startTelemetryUpload` (boot delay + daily interval, registers nothing when opted out);
- `server/src/routes/settings.ts` — `GET/PUT /telemetry` (`optIn` + `endpoint`);
- `server/src/index.ts` — wired into the ready hook.

## Tests

14 tests in `catalog-sync-scheduler.test.ts` (reusing its scheduler harness): defaults off, persistence, aggregation + 7-day window, opt-in gate, anonymized payload shape (no key/base-url leakage), endpoint rejection, scheduler registration on/off.

### DIFF
--- a/server/src/__tests__/services/catalog-sync-scheduler.test.ts
+++ b/server/src/__tests__/services/catalog-sync-scheduler.test.ts
@@ -1,6 +1,16 @@
 import { describe, it, expect, beforeAll, afterEach, vi } from 'vitest';
-import { initDb } from '../../db/index.js';
+import { initDb, getDb } from '../../db/index.js';
 import { startCatalogSync, stopCatalogSync } from '../../services/catalog-sync.js';
+import {
+  collectTelemetryStats,
+  isTelemetryOptIn,
+  setTelemetryOptIn,
+  getTelemetryEndpoint,
+  setTelemetryEndpoint,
+  uploadTelemetry,
+  startTelemetryUpload,
+  cancelTelemetryUpload,
+} from '../../services/telemetry.js';
 import type { Scheduler } from '../../lib/scheduler.js';
 
 function makeScheduler() {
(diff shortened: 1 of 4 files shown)


tashfeenahmed:
Thanks for building this. I'm closing it for a product reason rather than a code one: this is the send half of a community reliability prior, and there is no collector anywhere to receive the uploads. Until I decide to operate that infrastructure, an outbound uploader is dead weight, and freezing the wire format at v1 before the trust model exists (install identity, submission auth, dedupe) would box us in.

For the record, the privacy posture here is clean: default off, operator supplied URL, minimal aggregate payload. If a collector design ever lands, I'd revisit this together with a few counting fixes (exclude canceled and non-chat requests, drop custom-platform rows, capture the scheduler cancel handle, require https).

One note: the PR description cites #685 design B1, but #685 is a feature request about model detail hints on custom fetch. Please keep issue references to things the issue actually asks for.

PR #745, opened 2026-08-04
14050512

akaravi:


akaravi:
merge by main

tashfeenahmed:
Hi @akaravi, thanks for putting this together, and welcome! The code itself is solid work, but I have to close this one for timing reasons rather than quality.

The core feature landed on main while this branch was in flight: client profiles (#411, merged via #758) already provide multiple named bearer tokens with enable/disable. They store tokens as SHA-256 digests, so the DB never holds plaintext for auth. This PR uses reversible encryption with a reveal endpoint and a decrypt-all scan on each request, which is a weaker model, and I can't take two competing systems on the auth path.

The per-key toggle and the `fa` locale are also already on main (the `I18nProvider` change here would regress the locale list). One heads-up: `docker/DEPLOY-karavi-s40.md` looks like your private deployment runbook with your server paths in it. You may want to remove that from the branch regardless.

Two pieces I'd genuinely like to merge, and I'd welcome a fresh small PR against current main for them:

1. The `FREEAPI_UNIFIED_KEY_PREFIX` env var, with its test and docs rows.
2. The 204/205 no-body handling in `client/src/lib/api.ts`.

Please leave out the lockfile churn and the deploy doc, and it should be a quick review. Thanks again, hope to see the follow-up.

Issue #674, opened 2026-07-30
[Feature Request] Auto Update and add models list of Custom providers


euphoria360:
Running v0.6.5 when we add a custom openAI compatible provider, there is a new option to automatically fetch and add models the provider have.
Some providers change their models availability almost daily.
It would be nice to have freellmapi to autofetch and add all of them daily.

Aldo-f:
Related to #292
Things like that would probably be a [paid-feature](https://github.com/tashfeenahmed/freellmapi/issues/292#issuecomment-4691822037)...

euphoria360:
> Related to [#292](https://github.com/tashfeenahmed/freellmapi/issues/292) Things like that would probably be a [paid-feature](https://github.com/tashfeenahmed/freellmapi/issues/292#issuecomment-4691822037)...


it is fine to have the one month paywall limit. as far as i understood, free mode of freellmapi, has a paywall to sync new **models** after 1 month of their release. there is no limit stated about the providers:

<img width="392" height="412" alt="Image" src="https://github.com/user-attachments/assets/bb536c40-b211-4a93-994d-3487c0c8baf7" />


It would be nice if it could sync custom provider models daily and then check their models in its model database and only add models that are more than 1 month old.

Aldo-f:
> It would be nice if it could sync custom provider models daily and then check their models in its model database and only add models that are more than 1 month old.

This is the case, see this PR: #649 _(I haven't tested it)_
So i think this FR can be closed @euphoria360 ?

euphoria360:
if freellamapi adds new models for custom providers automatically already, then yes. thank you.

tashfeenahmed:
Closing as a duplicate of #292, which tracks the scheduled auto-refresh. What exists today: each custom endpoint has a one-click "Fetch models" button that pulls /v1/models and adds anything new. Making that run automatically on a schedule is the remaining piece, tracked in #292.

PR #746, opened 2026-08-04
feat(custom): auto-sync custom provider model lists on a schedule

suantea:
## What

Custom providers change their model availability almost daily, but today the model list is only refreshed when the operator manually hits "Fetch models". This adds a scheduled pass that walks every configured custom endpoint and registers whatever NEW models it serves.

- `services/custom-model-register.ts` — the POST /custom transaction body extracted verbatim into a shared function, so manual and scheduled registration share one code path and can never drift;
- `services/custom-model-sync.ts` — `runCustomModelSync(db)` enumerates custom endpoints (`listCustomEndpoints`, added to `custom-endpoint.ts`), discovers each one's current model list and registers the fresh ids. `startCustomModelSync(db, scheduler)` wires it into the existing `NodeScheduler` on server start;
- Interval: `CUSTOM_MODEL_SYNC_INTERVAL_MS` env (0 = disable, default 24h — daily, per #674).

Deliberate bounds:
- **Add only, never delete** — a model that disappears upstream keeps its row; an automated pass must not remove something the operator tuned by hand.
- **Failure isolation** — one dead relay is logged and skipped, the rest still sync.
- **No catalog paywall interplay** — these models come from the operator's own endpoint with their own key, not from the published catalog (covers #674's discussion; the free tier's 1-month catalog delay is untouched).
- The bundled free catalog is never touched, which is exactly what #663's "don't auto-refresh the built-in API" asks for.

## Tests

- `custom-model-sync.test.ts` (5 cases): zero endpoints, new-model registration, skip-already-registered, failing-endpoint isolation, interval env parsing (default/0/override).
- Server `tsc` passes. Full suite: 1850/1853 green — the 3 failures (`db/hardening` ×2 on Windows, `csp-inline-bootstrap` hash drift) reproduce on a clean `main`, pre-existing and unrelated.

## Files

- `server/src/routes/keys.ts` (POST /custom now delegates to the shared register)
- `server/src/services/custom-endpoint.ts` (+`listCustomEndpoints`)
- `server/src/services/custom-model-register.ts` (new)
- `server/src/services/custom-model-sync.ts` (new)
- `server/src/index.ts` (schedule the pass on start)
- `server/src/__tests__/services/custom-model-sync.test.ts` (new)

Refs #674, #663, #656


Aldo-f:
What happens when our custom provider also has a paid model?
Will this then be used?

I was just thinking about how this repo only accepts free models.
Is it allowed to include paid models from a custom provider?

Auto syncing the custom provider will automatically add any paid models it find.


(I'm only mentioning this because I dont think we discussed this use case. Even when we click the sync button there is a change that a paid model will be included)

suantea:
**Great question, and thanks for raising it — we honestly haven't discussed this use case yet.**

**With the current PR implementation, auto-sync and manual sync share the exact same registration path, so if a custom provider's model list includes paid models, they would indeed be registered too. That conflicts with this repo's "free models only" policy, and it's a gap we need to close.**

**My idea is to classify each model as free or paid during sync — e.g., extract pricing from the provider's own deployment docs / model pages (AI-assisted) and tag every model as `free` or `paid`. Paid models would then be skipped by default, or at least flagged for human review. This automates most of what is manual verification today, leaves humans to handle only edge cases, and guarantees the auto-sync never violates the repo's policy.**

**If, in the future, the maintainers decide paid models should be allowed (e.g., as an opt-in feature that each operator can enable or disable), that can be discussed separately as an option — but as the default behavior, I believe we should strictly follow the "free only" policy.**

**On a side note: I'm genuinely happy to see how fast this project is growing, and I hope to contribute more as a participant in the future. If you're open to this direction, I'd be glad to help land the improvement.**

suantea:
Rebased onto current `main` (v0.6.9) and resolved the conflicts:

- **`keys.ts`**: kept the newer `registerCustomChatModels` path (still used by the bulk importer) alongside the PR's extraction — `POST /custom` now delegates to the shared `registerCustomModels` exactly as before, and the scheduled sync shares that same code path. The merge also carried over the newer endpoint-pooling (`resolveCustomEndpointKey`, #619) and model-identity (#651) behavior that landed on main since this branch was created.
- Restored the `ensureModelInProfiles` import that the merge had dropped.

Verified: server `tsc -b` clean; custom-model-sync + keys + custom-provider + custom-endpoint-identity suites 81/81 green. CI re-running on the rebased branch.

PR #759, opened 2026-08-06
style(routing): render the exploration toggle as a strategy-matching button

suantea:
The exploration toggle (#731) rendered as a plain checkbox footnote below the strategy selector, visually unrelated to the segmented strategy buttons it belongs to. Now rendered as a button with the same px-3 py-1.5 text-xs rounded-lg transition-colors treatment (active = bg-foreground text-background font-medium), sitting on the same row as the strategy group. role="switch" + aria-checked preserve the accessibility contract of the checkbox it replaces; tooltip stays. client tsc -b passes. Refs #731

### DIFF
--- a/client/src/pages/FallbackPage.tsx
+++ b/client/src/pages/FallbackPage.tsx
@@ -249,54 +249,60 @@ export default function FallbackPage() {
             )}
           </div>
 
-          <div className="inline-flex flex-wrap items-center gap-1 rounded-xl border p-1">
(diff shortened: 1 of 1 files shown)


tashfeenahmed:
Thanks for the polish, but the footnote placement is intentional. Exploration is a niche knob and I want the routing card to stay light. Styled as a pill next to the strategy buttons it also reads like a fifth selected strategy, which is more confusing than the footnote. Keeping the current placement, so closing this.

PR #762, opened 2026-08-06
fix(analytics): use viewer timezone for timeline chart buckets

### LINKED ISSUE #761: Analytics timeline chart x-axis uses UTC, not the viewer's local timezone
## Problem

On the Analytics page, the "Requests over time" and "Tokens over time" charts label their x-axis in UTC. For viewers east of UTC the axis appears shifted — e.g. in UTC+8 a request made at 09:00 local shows up under 01:00. For day-granularity ranges (7d/30d/90d) the daily buckets are UTC days, not local days, so "today" on the chart doesn't match the viewer's wall clock.

## Root cause

`/api/analytics/timeline` buckets rows with SQLite `strftime('%Y-%m-%dT%H:00:00', hour)` / `strftime('%Y-%m-%d', hour)` over `request_hourly.hour`, which is stored in UTC (`server/src/routes/analytics.ts`). The client renders the returned `timestamp` string verbatim as the x-axis label (`XAxis dataKey="timestamp"` in `AnalyticsPage.tsx`) with no localization.

Note this is a different layer from #170 (shipped in #176): that fix localized point timestamps (Recent Errors, request detail, Keys page) via `formatSqliteUtcToLocalTime`, but the timeline chart never went through that path. A client-side label conversion alone would also still leave day buckets aggregated on UTC day boundaries, so the bucketing itself has to be timezone-aware.

## Proposed fix

Let the client pass its UTC offset (e.g. `tzOffset=480`, minutes) to `/api/analytics/timeline`; the server applies it as a bound `strftime` modifier (`'+480 minutes'`) so both hour and day buckets follow the viewer's wall clock. Validate the parameter (integer, -720..840) and fall back to UTC bucketing when absent or invalid, keeping current behavior for API consumers that don't send it.

I have this implemented locally with tests (day-boundary bucketing, hour-label shift, invalid-input fallback) and will open a PR referencing this issue.

hb-0:
Fixes #761

## Problem

The Analytics page's "Requests over time" and "Tokens over time" charts rendered their x-axis in UTC. For viewers east of UTC the axis was shifted, and for day-granularity ranges (7d/30d/90d) the daily buckets followed UTC day boundaries instead of the viewer's wall clock.

## Root cause

`/api/analytics/timeline` bucketed rows with SQLite `strftime(..., hour)` over `request_hourly.hour`, which is stored in UTC. The client rendered the returned `timestamp` string verbatim as the x-axis label with no localization.

Note: #170 (shipped in #176) fixed point-in-time displays (Recent Errors, request detail, Keys page) via `formatSqliteUtcToLocalTime`, but the timeline chart was not covered, and a client-side label conversion alone would still leave day buckets aggregated on UTC boundaries.

## Change

- `server/src/routes/analytics.ts`: accept a `tzOffset` query parameter (minutes from UTC, e.g. `480` for UTC+8). Validate it as an integer in `[-720, 840]` and apply it as a bound `strftime` modifier (`'+480 minutes'`) so both hour and day buckets follow the viewer's wall clock. Invalid/missing values fall back to UTC bucketing.
- - `client/src/pages/AnalyticsPage.tsx`: compute `tzOffset = -new Date().getTimezoneOffset()` and pass it to `/api/analytics/timeline`, including it in the React Query key.
- - `server/src/__tests__/routes/analytics.test.ts`: add tests for day-boundary bucketing, hour-label shift, and invalid-input fallback.
## Verification

- `npm test` server suite: 187 files, 1953 tests passed.
- - `npm run build` / `tsc` clean on both server and client.
- - Deployed to my own server and confirmed the `/analytics` x-axis now shows local time instead of UTC.


tashfeenahmed:
Thanks for this, and for filing #761 first. Your server side day bucketing was the part #804 could not cover, so I merged your branch with a small follow up commit that stops the tick formatter from shifting the already local timestamps a second time. Welcome aboard, you are on the contributors list now.

Issue #767, opened 2026-08-06
[Model Request] Add 5 free Ollama Cloud models (gpt-oss:20b, minimax-m3, nemotron-3 family)


mohammadsaleh0007:
Verified live against the Free tier of Ollama Cloud (https://ollama.com/v1,
platform `ollama`, key prefix `OLLAMA_CLOUD_`): all 5 return 200 on
`POST /v1/chat/completions` with a Free account (tested 2026-08-06).

| Catalog model_id | Ollama library |
|---|---|
| `gpt-oss:20b` | https://ollama.com/library/gpt-oss |
| `minimax-m3` | https://ollama.com/library/minimax-m3 |
| `nemotron-3-nano:30b` | https://ollama.com/library/nemotron-3-nano |
| `nemotron-3-super` | https://ollama.com/library/nemotron-3-super |
| `nemotron-3-ultra` | https://ollama.com/library/nemotron-3-ultra |

Notes:
- These are the ONLY currently-free Cloud models beyond the existing
  gemma4:31b / gpt-oss:120b. Tested the full /v1/models list (18 models);
  the rest (deepseek-v4-*, glm-5.x, kimi-k2.x/k3, minimax-m2.7,
  mistral-large-3:675b, qwen3.5:397b) all return 403
  "this model requires a subscription" on Free.
- model_id uses the exact ID returned by /v1/models (name:tag convention,
  matching existing rows like gpt-oss:120b).
- Pricing rows would be needed in model-pricing.ts (defaults fine).

tashfeenahmed:
These are all live in the catalog now (gpt-oss:20b, minimax-m3, and the nemotron-3 nano/super/ultra), with pricing rows added in #768. They ship through the hosted catalog, so they'll appear after the next sync, or you can trigger one manually from the dashboard. Thanks for probing these and writing them up so clearly.

PR #768, opened 2026-08-07
feat(models): add 4 free Ollama Cloud models (#767)

suantea:
Adds the verified-free Ollama Cloud models requested in #767 (tested live 2026-08-06 against the Free tier at https://ollama.com/v1):

- `minimax-m3`, `nemotron-3-nano:30b`, `nemotron-3-super`, `nemotron-3-ultra` added to the catalog
- `gpt-oss:20b` was already present
- `model_id` uses the exact `/v1/models` id (name:tag convention, matching rows like `gpt-oss:120b`)
- pricing rows mirror the same models paid variants on OpenCode Zen

Server tsc passes. Refs #767


tashfeenahmed:
Thanks! Good news: all four of these are already in the hosted catalog, which is where model data ships from these days. The baseline migration is frozen and existing installs never re-run it, so rows added there only reach fresh installs (same note as on #772). Premium installs have these models already, and the free tier picks them up through the monthly catalog as each row clears the 30 day window: minimax-m3 and nemotron-3-super on Aug 17, nemotron-3-nano:30b and nemotron-3-ultra on Aug 27. One heads up on the data itself: 131072 is not the right context for these models, they range from 256K to 1M, so worth checking the model pages when authoring rows. Your model-pricing.ts additions are the part the repo actually needs, so I trimmed the PR down to those four rows and merged that. Thanks for following up on #767.

PR #769, opened 2026-08-07
feat(scripts): add a semi-automated free-model scanner

suantea:
Follow-up to #767 (Model Request). A semi-automated scanner that probes the `/v1/models` endpoints of the common free LLM providers (Groq, OpenRouter, Ollama Cloud, NVIDIA NIM, Mistral) with optional per-provider keys (`SCAN_<PROVIDER>_KEY`) and prints a candidate model list for manual review against the catalog.

Deliberately semi-automated: free-tier availability churns (#722 — Cloudflare Kimi moved to paid; OpenRouter `:free` rotates), so the output is a candidate list for a human to confirm, never an auto-write into the catalog.

Verified: script runs and degrades gracefully without keys. Refs #767

### DIFF
--- a/scripts/scan-free-models.ts
+++ b/scripts/scan-free-models.ts
@@ -1,15 +1,21 @@
 #!/usr/bin/env node
-// scan-free-models.ts — 半自动免费模型扫描器
+// scan-free-models.ts — semi-automated free-model scanner
 //
-// 探测各免费 LLM provider 的 OpenAI 兼容 /v1/models 端点,输出候选模型清单,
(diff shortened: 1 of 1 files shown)


tashfeenahmed:
Thanks for putting this together, the idea is sound and the free tier churn you point at is real. But model discovery is handled by the hosted catalog pipeline these days, which already tracks provider listings against the live catalog rather than the frozen baseline migration, so a repo local scanner would duplicate that and drift out of sync (as it already does here by reading the January baseline). I would rather not take on the maintenance surface for a second path, so closing this. Appreciate the effort, and the #722 style fixes that come out of watching free tiers are always welcome.

Issue #779, opened 2026-08-08
3 pre-existing test failures: CSP bootstrap hash drift + DB file-permission tests fail on Windows


suantea:
## Summary

Three tests fail on a clean `origin/main` checkout (verified with a stash on top of v0.6.9, so unrelated to any pending PR):

### 1. `csp-inline-bootstrap.test.ts` — "the hash in app.ts matches the inline script in client/index.html" (real drift, all platforms)

The guard this suite exists for has tripped: `INLINE_BOOTSTRAP_SHA` in `server/src/app.ts` is

```
'sha256-4Mz/yZAENQGlTAAEe1WqXruXCripvlvl0s+Q9S1VS4A='
```

but the inline `<script>` in `client/index.html` computes to

```
'sha256-NhQ9t1luOLg/gJxVj+cgjX5vxl4o7owwcLBpVHf+wD8='
```

So the inline theme/direction bootstrap script was edited without syncing the CSP hash. The header is served with the stale hash → the bootstrap is blocked by CSP in the browser (`script-src 'self'` + stale hash), which reintroduces the dark-mode white-flash / theme bootstrap failure that #682 fixed. The other two tests in the file pass (including "matches the built client" — likely because `client/dist` is absent in a fresh checkout).

Fix: regenerate the hash from the current `client/index.html` inline script and update `INLINE_BOOTSTRAP_SHA`.

### 2. `db/hardening.test.ts` — "restricts the database file to the owner" and "restricts the WAL sidecars once they exist" (Windows)

`expect(mode & 0o077).toBe(0)` fails with `54` (`0o66`) on Windows:

```
AssertionError: expected 54 to be +0 // Object.is equality
  ❯ src/__tests__/db/hardening.test.ts:55:26
```

Windows does not enforce POSIX permission bits — `fs.chmod`/default file modes leave group/other read+write bits set, so these assertions can never pass there. Suggest skipping the chmod assertion on `process.platform === 'win32'` (or documenting that the OS-level restriction is only enforceable on POSIX).

## Environment

- Windows 10/11, Node via Git Bash
- `vitest run` (server) — full suite: 1965/1968 pass, these 3 fail

suantea:
Verified fixed on current main — closing.

All three failures from this report are resolved:

[...]

PR #780, opened 2026-08-09
fix(tests): skip owner-only DB mode assertions on Windows (#779)

suantea:
fix(tests): skip owner-only DB mode assertions on Windows (#779)

Windows does not enforce POSIX permission bits — fs.chmod and default
file modes leave group/other bits set — so `expect(mode & 0o077).toBe(0)`
in db/hardening.test.ts can never pass there (2 of the 3 pre-existing
failures in #779). Guard both assertions with `it.skipIf(win32)`: the
owner-only restriction is only enforceable/testable on POSIX.

Note on the third failure: the CSP inline-bootstrap hash already matches
on current main (`csp-inline-bootstrap.test.ts` passes 4/4), so no
change was needed there.

Refs #779

### DIFF
--- a/server/src/__tests__/db/hardening.test.ts
+++ b/server/src/__tests__/db/hardening.test.ts
@@ -46,7 +46,13 @@ describe('database runtime hardening', () => {
     expect(jmValue.toLowerCase()).toBe('wal');
   });
 
-  it('restricts the database file to the owner', () => {
+  // Windows does not enforce POSIX permission bits: fs.chmod and default file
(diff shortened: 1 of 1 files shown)


tashfeenahmed:
Superseded by #791, which is now merged. Instead of skipping the owner-only assertions on Windows, #791 makes the restriction real there via icacls and asserts the actual ACL, so the tests pass on Windows by verifying the guarantee rather than waiving it. Thanks for flagging #779 with a workable interim fix.

PR #782, opened 2026-08-09
feat(dashboard): automatic update check with release-notes dialog

suantea:
## What's new

Automatic update detection (previously only a manual check in Settings):

- Checks the GitHub latest release on load, throttled to once per 24h, with the result cached in localStorage (no outbound request on every page load)
- When a newer version exists, a corner pill appears bottom-left; clicking it opens a dialog with the release notes (Markdown-rendered), a link to the release page, and a per-version dismiss
- i18n keys added to all 60 locale files (`npm run check:i18n` passes; en is the fallback)

## Files touched

- `client/src/components/update-reminder.tsx` (new)
- `client/src/App.tsx` (mount)
- `client/src/i18n/locales/*.json` (update.* keys in all 60 locales)


tashfeenahmed:
Merged with two changes to the transport, since the component itself was solid. First, the browser fetch to api.github.com could never work outside the Electron shell: the dashboard ships connect-src 'self', so the check now goes through GET /api/update/release on the server, which reuses the existing #635 update checker's fetch conventions and caches for 6 hours. Second, an automatic phone-home needs consent in a self-hosted app, so the check is now opt-in via a toggle next to the version row (settings key update_check_enabled, default off). Also fixed along the way: the per-version dismissal never persisted, an opt-out could show a stale pill for up to a day, and two lint errors. Thanks for the feature, the pill and release-notes dialog are unchanged.

PR #818, opened 2026-08-10
feat(custom): expand upstream model aliases into catalog entries (#790)

### LINKED ISSUE #790: Allow merging model aliases of custom providers into the primary model id
I haven't found the relevant option. Aliases merging would allow requests for the primary model to be served via these aliases.

Image

suantea:
Closes #790

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ server/data/
 .env.local
 .DS_Store
 
+# TypeScript incremental build cache
+*.tsbuildinfo
+
 # Personal deployment scripts (contain keys/credentials — kept local)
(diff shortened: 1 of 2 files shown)


tashfeenahmed:
Thanks, but closing this one for now. I checked the live model rosters from every provider currently in the pool (325 model entries from the three most recent catalog sweeps) and none of them ship an aliases, alias, id_aliases or ids field, so there is nothing for this expansion to act on today. Treating a generic ids field as an alias list is also risky: on a relay that uses ids for something else it would fabricate catalog entries. Happy to revisit with a concrete provider and payload where aliases actually appear.

Issue #789, opened 2026-08-09
Over‑crowded tooltip text for providers


yfdyh000:
The tooltips containing providers name or scores for the model‑list use commas or dot separators instead of line breaks to separate providers. Custom labels or custom‑provider base url clutter the list and render it nearly unreadable.

<img width="337" height="155" alt="Image" src="https://github.com/user-attachments/assets/b262a89f-4f92-4624-a180-8f47522dc3eb" />

<img width="245" height="196" alt="Image" src="https://github.com/user-attachments/assets/cb7f2596-532d-42b2-a53c-4072de56e09b" />

PR #835, opened 2026-08-11
fix(agents/models): line-break provider tooltips instead of comma-cluttering them (#789)

suantea:
## What
#789：provider 过多的 tooltip 文本挤成一长串逗号连接，难以阅读。

## Change
- tooltip 容器支持 `whitespace-pre-line`，多 provider 用换行（`\n`）分隔而不是逗号
- model-combobox / model-table 的 servedBy 与 score breakdown tooltip 改为逐行展示
- 附带：AgentIcon 无品牌图标时使用 lettermark 兜底（atomcode/mimo 用品牌色）

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ server/data/
 .env.local
 .DS_Store
 
+# TypeScript incremental build cache
+*.tsbuildinfo
+
 # Personal deployment scripts (contain keys/credentials — kept local)
 deploy-pi.sh
 update-hermes.sh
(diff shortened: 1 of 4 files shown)


tashfeenahmed:
Closing this one as a duplicate of #836, which I just merged.

The two PRs make the same three edits to the same three files. The only difference here is the extra .gitignore block, and `*.tsbuildinfo` is already ignored in main (.gitignore lines 18-19), so adding it again would just duplicate those lines.

Thanks for the fix, it landed via #836.

Issue #814, opened 2026-08-10
Expose curated multimodal OpenCode models


aotoganngi-ctrl:
## Parent

Part of #811.

## What to build

Expose the selected coding, visual, and short-video models as individual OpenCode choices while retaining the complete enabled catalog behind capability-first auto routing.

## Acceptance criteria

- [ ] Selected models appear as named OpenCode choices with concise task descriptions.
- [ ] Gemini 3.5 Flash and Gemini 3 Flash Preview are individually available for short-video analysis.
- [ ] Auto routing remains capability-first and can use pinned or unpinned enabled models.

## Blocked by

Multimodal gateway verification.

tashfeenahmed:
Closing with #811, see the note there.

suantea:
## 中文

认领此 issue（#843 多模态切片的 curated 模型部分）并开始开发。计划：

- 新增迁移 `20260811_000001_opencode_multimodal_curation`：为 opencode 平台添加 curated 多模态模型条目——**Qwen VL** + **Gemini 3.5 Flash** + **Gemini 3 Flash Preview**（短视频分析），固定 model_id 保证 roundtrip 稳定；
- 自动路由保持能力优先（capability-first），curated 模型既可被 auto 路由使用，也可作为独立 OpenCode 选择；
- 验证：server `tsc -b` + 迁移 roundtrip 测试。

## English

Claiming this issue (the curated-models part of the #843 multimodal slice) and starting development. Plan:

- New migration `20260811_000001_opencode_multimodal_curation`: add curated multimodal model rows for the opencode platform — **Qwen VL** + **Gemini 3.5 Flash** + **Gemini 3 Flash Preview** (short-video analysis), fixed model ids for roundtrip stability;
- Auto routing stays capability-first; curated models are usable by auto routing and as individual OpenCode choices;
- Verify with server `tsc -b` + migration roundtrip tests.

Issue #813, opened 2026-08-10
Test and deploy multimodal gateway


aotoganngi-ctrl:
## Parent

Part of #811.

## What to build

Prove the multimodal gateway behavior through its public chat-completions API and run it as a loopback-only local service.

## Acceptance criteria

- [ ] Tests cover text pass-through, visual pass-through, non-visual fallback, fallback failure, and image removal.
- [ ] The deployed service listens only on loopback.
- [ ] The local health check succeeds.

## Blocked by

Multimodal gateway implementation.

tashfeenahmed:
Closing with #811, see the note there.

Issue #812, opened 2026-08-10
Implement capability-aware multimodal gateway


aotoganngi-ctrl:
## Parent

Part of #811.

## What to build

Provide a local OpenAI-compatible multimodal gateway that forwards text requests and makes visual routing capability-aware: direct pass-through for visual-capable target models, and Qwen VL pre-analysis followed by the originally requested non-visual model for image requests.

## Acceptance criteria

- [ ] Text requests are forwarded unchanged.
- [ ] Visual-capable targets receive image requests unchanged.
- [ ] Non-visual image targets receive Qwen VL analysis before the target call.
- [ ] The target call does not receive image blocks after fallback analysis.

## Blocked by

None.

tashfeenahmed:
Closing with #811, see the note there.

Issue #811, opened 2026-08-10
Add capability-aware visual fallback and curated OpenCode models


aotoganngi-ctrl:
## Problem Statement

OpenCode users need a curated set of pinned coding and multimodal models while retaining one capability-first auto route for the remaining enabled catalog. Image requests sent to a non-visual target need a visual interpretation before the requested coding model can act. Short-video analysis needs explicitly selectable Gemini models.

## Solution

Add a local multimodal gateway ahead of FreeLLMAPI. It preserves text-only and visual-capable requests, performs a Qwen VL analysis only for image requests targeting non-visual models, and forwards that analysis to the originally selected model. Expose the selected model set and two Gemini video-analysis entries in the OpenCode GitHub provider; keep all enabled models eligible for the capability-first auto route.

## User Stories

1. As an OpenCode user, I want pinned models shown as individual choices so that I can select a known capability.
2. As an OpenCode user, I want auto routing to prefer stronger models so that routine work uses the best available capacity.
3. As an OpenCode user, I want unpinned models retained in auto routing so that free capacity is not wasted.
4. As an OpenCode user, I want image-capable targets to receive images directly so that no unnecessary model call is made.
5. As an OpenCode user, I want non-visual targets to receive a visual analysis so that they can implement changes based on screenshots.
6. As an OpenCode user, I want the visual analysis to preserve the original task context so that coding output remains relevant.
7. As an OpenCode user, I want text-only requests to bypass visual processing so that latency and quota use remain unchanged.
8. As an OpenCode user, I want a clear failure when the visual preprocessor fails so that I can retry or choose a visual target.
9. As an OpenCode user, I want Gemini video models listed separately so that I can analyze short clips deliberately.
10. As an OpenCode user, I want video constraints documented so that I do not send unsupported long videos.
11. As an operator, I want provider credentials to remain outside source control and public issues.
12. As an operator, I want model routing to remain loopback-only so that local credentials are not exposed on the LAN.

## Implementation Decisions

- Use a local OpenAI-compatible gateway between OpenCode and FreeLLMAPI.
- Treat visual capability as a model property; Qwen2.5-VL-72B-Instruct is the visual fallback.
- For a non-visual image target, call the visual fallback first, remove image blocks from the second request, and append its analysis as context.
- Do not introduce video frame extraction; video support is limited to direct short MP4 handling by the two Gemini entries and their documented media-size constraints.
- Keep capability-first auto routing and the existing enabled catalog.
[...]

PR #843, opened 2026-08-11
feat: AI model intelligence, multimodal fallback, multi-user, and modularity (#760 #811 #812 #813 #814 #763 #267 #824)

suantea:
Batch of independent sub-features, each reviewable separately:

## #760 — AI model intelligence (P1+P2)
- **Benchmarks**: `/api/analytics/by-model` now returns a benchmark surface — `avgTtfbMs` + `min/max latency` spread — so the UI can answer "is this model slow today" without digging into traces
- **Dynamic ranking**: `timeOfDayWeights()` shifts speed→reliability weight during local peak hours (18:00–06:00) when free relays are congested; off-peak behaviour is unchanged

## #811/#812/#813/#814 — multimodal gateway
- Migration `20260811_000001_opencode_multimodal_curation`: curated opencode multimodal models — Qwen VL + two Gemini video-analysis entries (fixed ids for roundtrip stability)
- `lib/vision-fallback.ts`: Qwen VL pre-analysis replaces image blocks with text before a non-visual target call (image removal — the target never receives image blocks)
- `proxy.ts`: capability-aware fallback runs before the no-vision rejection; `vision-fallback.test.ts` covers pass-through / fallback / failure / removal (7 cases)

## #763 — modularity
`services/modules.ts`: a dependency-free pluggable-module registry (register / enable / disable / isEnabled), with compression registered as the flagship module.

## #267 — multi-user (minimal slice)
`POST /api/auth/users` (requireAuth) lets an authenticated user create additional dashboard accounts — each signs in with their own session. Per-user API keys/workspaces are the next increment.

## #824 — docs
FAQ section on password reset, forgot-password reset codes (server logs), logs location, and the desktop no-password build.

## Verification
- server `tsc` clean
- vision-fallback 7/7; roundtrip + registry-drift + scoring 40/40

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -271,3 +271,23 @@ and `config.json` (window/theme/port/LAN preferences). Copy both to move an
 install. For the server (non-desktop) deployment, the equivalent state is the
 `.env` file and the SQLite DB at `server/data/freeapi.db` (or wherever
 `FREEAPI_DB_PATH` points).
+
+## FAQ: password reset and logs
+
+**Where are the logs?**
+The router writes to stdout of the process that runs it — the terminal you
+launched `freellmapi` / the server in, or the desktop app's own console. The
+one-time codes described below appear there.
+
+**I forgot my password — how do I reset it?**
+On the login screen, click **Forgot password?** to generate a reset code. The
+code is printed to the **server logs** (see above) — open them, copy the code,
+paste it into the form, and set a new password. The dashboard account is the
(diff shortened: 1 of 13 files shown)


tashfeenahmed:
Closing this one, since you've already split it into #909, #910, #911 and #912. Those cover the same work and it's conflicting with main now anyway.

Thanks for breaking it up, it's much easier to review the pieces separately. I'll go through them individually and reply on each.

PR #845, opened 2026-08-11
fix(openai-compat): support real token usage in streaming responses

xyblue135:
## Summary

- accept `stream_options.include_usage` on `/v1/chat/completions`
- forward the option to OpenAI-compatible streaming providers
- use upstream-reported usage for analytics and rate-limit accounting
- keep the existing token estimate as a fallback when upstream usage is absent

## Problem

OpenAI-compatible streaming requests currently drop `stream_options.include_usage`. Even when an upstream usage frame is available, FreeLLMAPI records its local token estimate instead of the provider-reported values.

For the reproduced request, the upstream reported 3263 prompt tokens and 15 completion tokens, while the FreeLLMAPI dashboard recorded 1 input token and 23 output tokens.

## Test plan

- `npm run test -w server -- src/__tests__/routes/proxy-stream-integrity.test.ts src/__tests__/providers/openai-compat.test.ts` (71 tests passed)
- full server test suite passed
- `npm run test -w client --if-present` (69 tests passed; i18n validation passed)
- `npm run build` passed for server, CLI, and client

### DIFF
--- a/server/src/__tests__/routes/proxy-stream-integrity.test.ts
+++ b/server/src/__tests__/routes/proxy-stream-integrity.test.ts
@@ -49,7 +49,7 @@ const TOOLS = [{ type: 'function', function: { name: 'Read', description: 'read 
 function mockUpstream(script: Array<{ body: string; status?: number }>) {
   const origFetch = global.fetch;
   let call = 0;
-  const seen: Array<{ model: string }> = [];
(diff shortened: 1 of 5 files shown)


tashfeenahmed:
Thanks for this, the diagnosis was exactly right: stream_options never made it upstream, so streamed responses always fell back to the character estimate. Clean fix with a solid end to end test. Welcome aboard, you are on the contributors list now.

PR #844, opened 2026-08-11
fix(responses): align /v1/responses shim with the official Responses API spec

NirvanaCh7:
Fixes #842 (partially) — addresses the 5 highest-severity deviations. Items 6 and 7 are out of scope here.

## Fixed

- **1 — streaming `output_index` not unique** (`06b1f8b`): dense, unique indices across tool-call and text items in delta/done events, so the SDK's `snapshot.output[output_index]` no longer collides when a model streams tool_calls before the closing text (GLM/Qwen).
- **2 — `inputItemSchema` rejects standard item types** (`7b58251`): accept the remaining official `ResponseInputItemParam` kinds (`computer_call`, `computer_call_output`, `reasoning`, `local_shell_call`) so agent harnesses stop 400'ing at validation; `toChatMessages` skips non-message kinds instead of mis-parsing them.
- **3 — computer use cannot survive translation** (`7b58251`): detect `computer`/`computer_use_preview` tools and `computer_call` items and fail up front with a clear `422 no_computer_use_model` rather than silently dropping the calls and breaking the harness loop. `responsesInputHasImage` also detects `computer_screenshot` parts under `computer_call_output.output`.
- **4 — consecutive / empty assistant messages not normalized** (`9bf082b`): merge a `message` item with the adjacent `function_call` items into a single assistant turn with `content` + `tool_calls` (the shape chat providers expect), folding tool_call ids so tool results pair correctly; drop empty assistant items with no tool_calls.
- **5 — `developer` → `system` can land mid-conversation** (`e071f31`): hoist all system/developer messages into a single leading block (instructions first, then in relative order), which Gemini/Claude/Mistral require.

## Not fixed (left for follow-up)

- **6 — `reasoning_content` dropped**, risking false empty-completion failover for thought-only outputs.
- **7 — streaming usage undercounts tool-call tokens**: text-only `totalOutputTokens`, hardcoded `cached_tokens` / `reasoning_tokens` of 0.
---
## Here are some additional patches. You can choose whether to merge them.
[...]

PR #846, opened 2026-08-11
fix(responses): dense unique output_index + accept official input item types (#842)

### LINKED ISSUE #842: [Bug]: /v1/responses shim diverges from the official Responses API spec
The `/v1/responses` shim translates OpenAI Responses API requests into chat-completions and forwards them to upstream providers. While testing this surface, several deviations from the official Responses API were found, in severity order.

### 1. Streaming `output_index` is not unique when tool_calls precede text

The Responses streaming SDK indexes `snapshot.output` by `output_index` and expects dense, unique indices. The shim hardcodes the text item's `output_index` to `0` in the delta/done events, and recomputes tool-call indices from `toolAcc.size`. When a model streams `tool_calls` *before* any closing text (common on GLM/Qwen), the `function_call` item owns index `0` and the later text item/deltas collide with it, so the SDK crashes on `snapshot.output[output_index]`.

### 2. `inputItemSchema` rejects standard item types

Validation only accepts `message` / `function_call` / `function_call_output`. The official `ResponseInputItemParam` union also includes `computer_call`, `computer_call_output`, `reasoning`, `local_shell_call`, etc. Clients that round-trip these items (e.g. agent harnesses) get a 400 at the schema before routing, even though the fields the shim consumes are present.

### 3. Computer use cannot survive the chat-completions translation

A `computer` / `computer_use_preview` tool plus its `computer_call`/`computer_call_output` items have no chat-completions equivalent (no computer tool, no screenshot context). Dropping them silently would break the harness loop; currently they are neither rejected nor mapped, so behavior is undefined.

### 4. Consecutive / empty assistant messages are not normalized

Multi-turn input often pairs a `message` item with an adjacent `function_call` item. `toChatMessages` emits two consecutive `role: 'assistant'` messages, which Gemini maps to consecutive model turns and strict upstreams (Mistral/Cohere) reject with 400. The `/v1/chat/completions` route normalizes this; the responses translation does not.

### 5. `developer` → `system` can land mid-conversation

`toChatMessages` keeps original ordering and maps `developer` → `system`, so a system message can appear after user turns, which strict upstreams reject.

### 6. `reasoning_content` is dropped, risking false empty-completion failover

Streaming `delta.reasoning_content` and non-streaming `message.reasoning_content` are ignored. A reasoning model that emits thought-only output with no body text is treated as an empty completion, and the healthy model is failed over / benched.

### 7. Streaming usage undercounts tool-call tokens

`totalOutputTokens` accumulates only text deltas; tool-call name + arguments are not counted, and `cached_tokens` / `reasoning_tokens` are hardcoded to 0.

suantea:
Closes #842

Two deviations from the official Responses API spec:

## 1. Streaming `output_index` not unique when tool_calls precede text

The shim hardcoded the text item's `output_index` to `0` and recomputed tool-call indices from `toolAcc.size`. On GLM/Qwen streams (tool_calls *before* any closing text — the common ordering), the `function_call` item owns index `0` and the later text deltas collide with it, so the SDK crashes on `snapshot.output[output_index]`.

**Fix**: one monotonic counter (`nextOutputIndex++`) shared by text items, streamed tool calls and rescued inline calls — indices stay dense and unique across every output item.

## 2. `inputItemSchema` rejected standard item types

Only `message` / `function_call` / `function_call_output` were accepted; the official `ResponseInputItemParam` union also includes `computer_call`, `computer_call_output`, `reasoning`, `local_shell_call`, `web_search_call`. Agent harnesses round-tripping these items got a 400 before routing.

**Fix**: accept them loosely (type-checked, `.passthrough()`) and ignore them at conversion — the shim only consumes message / function_call / function_call_output.

## Verification

- server `tsc` clean
- responses suite 14/14 pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ server/data/
 .env.local
 .DS_Store
 
+# TypeScript incremental build cache
+*.tsbuildinfo
+
 # Personal deployment scripts (contain keys/credentials — kept local)
 deploy-pi.sh
 update-hermes.sh
--- a/server/src/routes/responses.ts
+++ b/server/src/routes/responses.ts
@@ -100,10 +100,21 @@ const functionCallOutputItemSchema = z.object({
   output: z.union([z.string(), z.array(contentPartSchema), z.record(z.string(), z.unknown())]),
 });
 
+// #842: the official ResponseInputItemParam union also includes
+// computer_call / computer_call_output / reasoning / local_shell_call /
+// web_search_call. Agent harnesses round-trip these items and a strict union
+// rejected the whole request with a 400 before routing. We accept them loosely
+// (only the type is checked) and ignore them at conversion — the shim only
+// consumes message / function_call / function_call_output.
+const otherItemSchema = z.object({
+  type: z.enum(['computer_call', 'computer_call_output', 'reasoning', 'local_shell_call', 'web_search_call']),
+}).passthrough();
+
 const inputItemSchema = z.union([
   functionCallItemSchema,
   functionCallOutputItemSchema,
   messageItemSchema,
+  otherItemSchema,
 ]);
 
 // Accept ANY tool type, not just 'function'. Codex (Responses API) sends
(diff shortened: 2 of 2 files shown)


tashfeenahmed:
Thanks, the output_index diagnosis was correct and confirmed. Closing in favor of #844 though: it was the issue author's own fix, opened first, covers the same two items plus three more, and ships tests for all of them. One part of this PR made it in anyway, main now accepts web_search_call items, which #844 originally missed.

PR #847, opened 2026-08-11
fix(router): exploration never promotes a model that can't serve the request

NirvanaCh7:
## Summary

The "explore unseen models" routing toggle picked a random unmeasured model purely by sample count and pushed it to the front of the routing chain. For requests with request-level capability gates (vision / tools / structured output) it could promote a model the main routing loop would immediately skip — e.g. an image request randomly probing a text-only model, wasting a fallback hop and (in the worst case) surfacing the wrong model to the client.

## Change

`routeRequest` exploration candidates are now filtered by the same gates as the main loop:

- models ruled out earlier in the request (`skipModels`)
- text-only models when the request requires vision
- models without structured tool-call support when the request uses tools
- platforms that drop `response_format` when structured output is requested
- models whose context window or per-minute token limit can't hold the request

So an exploration pick is always a model that can actually serve the request; the loop still falls through to the scored order when the explore pick fails.

## Tests

Added a regression test in `router-bandit.test.ts` asserting an image request (requireVision) never routes to a text-only model even with exploration enabled. Full `router-bandit`, `router-key-bandit` and `fallback` suites pass; `tsc` clean.

### DIFF
--- a/server/src/__tests__/services/router-bandit.test.ts
+++ b/server/src/__tests__/services/router-bandit.test.ts
@@ -131,19 +131,31 @@ describe('bandit router', () => {
     expect(counts['y'] ?? 0).toBeGreaterThan(0);
   });
 
-  it('vision requests never route to a text-only model, even with exploration on', () => {
+  it('exploration draws are never wasted on a model that cannot serve the request', () => {
+    // vision-a is measured and wins every ordinary bandit draw. text-b and
+    // vision-c are both unmeasured, but only vision-c can serve a vision
+    // request. The pre-fix router put BOTH in the explore pool, so ~half the
+    // explore draws promoted text-b, which the main loop's vision gate then
+    // skipped — a wasted draw. With the pool filtered, vision-c receives
+    // every explore draw (~10% of requests); unfixed it got only ~5%, far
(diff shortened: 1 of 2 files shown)


NirvanaCh7:
358710ac94852da956e97e43fdc9fd4c8518ba08

NirvanaCh7:
You‘ve lost a commit in my repo https://github.com/NirvanaCh7/freellmapi/commit/5d4eb03bf91d572acc6fdb5e3166feef618100d0 
which hasn’t been merged. 
@tashfeenahmed

tashfeenahmed:
Merged, thanks. Two notes. First, an honest history note: through a push mix up on my side, your router filter itself actually reached main inside the squash labeled #844, and the responses work landed via #850, so this PR's final diff is the follow up hardening: the regression test now discriminates (the original one also passed on unfixed main because the loop gates already prevented wrong routing, so it pinned nothing) plus the skipPlatforms gate for a complete mirror of the loop. Second, the fix itself was correct and composes cleanly with the weight override exclusion from #747. The vision input support you offered to split from #844 is welcome as its own PR.

PR #851, opened 2026-08-11
fix(responses): align /v1/responses shim with the official Responses API spec

### LINKED ISSUE #842: [Bug]: /v1/responses shim diverges from the official Responses API spec

NirvanaCh7:
The origin #844 without vision support
Fixes #842 (partially) — addresses the 5 highest-severity deviations. Items 6 and 7 are out of scope here.

## Fixed

- **1 — streaming `output_index` not unique** (`29e2866`): dense, unique indices across tool-call and text items in delta/done events, so the SDK's `snapshot.output[output_index]` no longer collides when a model streams tool_calls before the closing text (GLM/Qwen).
- **2 — `inputItemSchema` rejects standard item types** (`f79091f`): accept the remaining official `ResponseInputItemParam` kinds (`computer_call`, `computer_call_output`, `reasoning`, `local_shell_call`) so agent harnesses stop 400'ing at validation; `toChatMessages` skips non-message kinds instead of mis-parsing them.
- **3 — computer use cannot survive translation** (`f79091f`): detect `computer`/`computer_use_preview` tools and `computer_call` items and fail up front with a clear `422 no_computer_use_model` rather than silently dropping the calls and breaking the harness loop. `responsesInputHasImage` also detects `computer_screenshot` parts under `computer_call_output.output`.
- **4 — consecutive / empty assistant messages not normalized** (`482b70a`): merge a `message` item with the adjacent `function_call` items into a single assistant turn with `content` + `tool_calls` (the shape chat providers expect), folding tool_call ids so tool results pair correctly; drop empty assistant items with no tool_calls.
- **5 — `developer` → `system` can land mid-conversation** (`e67620f`): hoist all system/developer messages into a single leading block (instructions first, then in relative order), which Gemini/Claude/Mistral require.

## Not fixed (left for follow-up)

- **6 — `reasoning_content` dropped**, risking false empty-completion failover for thought-only outputs.
- **7 — streaming usage undercounts tool-call tokens**: text-only `totalOutputTokens`, hardcoded `cached_tokens` / `reasoning_tokens` of 0.

### DIFF
--- a/server/src/__tests__/routes/responses-translate.test.ts
+++ b/server/src/__tests__/routes/responses-translate.test.ts
@@ -52,6 +52,52 @@ describe('Responses → chat translation (#96)', () => {
     expect(msgs[0]).toEqual({ role: 'tool', tool_call_id: 'call_1', content: 'sunny' });
   });
 
+  it('merges an assistant message item with its following function_call items into one turn', () => {
+    const msgs = toChatMessages({
+      input: [
+        { type: 'message', role: 'assistant', content: 'Let me check.' },
+        { type: 'function_call', call_id: 'call_1', name: 'get_weather', arguments: '{"city":"SF"}' },
+        { type: 'function_call', call_id: 'call_2', name: 'get_time', arguments: '{}' },
+      ],
+    } as any);
+    expect(msgs).toEqual([
+      {
+        role: 'assistant',
+        content: 'Let me check.',
+        tool_calls: [
+          { id: 'call_1', type: 'function', function: { name: 'get_weather', arguments: '{"city":"SF"}' } },
+          { id: 'call_2', type: 'function', function: { name: 'get_time', arguments: '{}' } },
+        ],
+      },
+    ]);
+  });
+
+  it('merges an empty assistant item with its following function_call items (content null)', () => {
+    const msgs = toChatMessages({
+      input: [
+        { type: 'message', role: 'assistant', content: [] },
+        { type: 'function_call', call_id: 'call_1', name: 'get_weather', arguments: '{}' },
+      ],
+    } as any);
+    expect(msgs).toEqual([
+      {
+        role: 'assistant',
+        content: null,
+        tool_calls: [{ id: 'call_1', type: 'function', function: { name: 'get_weather', arguments: '{}' } }],
+      },
+    ]);
+  });
+
+  it('drops empty assistant message items (no tool_calls)', () => {
+    const msgs = toChatMessages({
+      input: [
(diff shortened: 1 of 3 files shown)


NirvanaCh7:
e1ecd0f522ab03f57f676ddbfdbc5fd1e6e71d2f

PR #852, opened 2026-08-11
feat(vision): end-to-end image input — Responses translation, Gemini signature fallback, body limits, inbound normalization

NirvanaCh7:
## Summary

End-to-end image input support: Codex (and any Responses/chat client) can now send images through this proxy — translated correctly, routed to vision-capable models, accepted at realistic payload sizes, and shrunk before hitting upstream limits. Driven by a real failing session: a Codex turn with two duplicated 1080p screenshots produced an 11.85MB request that died in four different ways this branch fixes, one commit each.

### 1. `fix(responses)`: translate image input and route it to vision-capable models

**What:** `/v1/responses` previously flattened every content part to text and hard-422'd any image input. Now `input_image` / `image_url` / `computer_screenshot` parts become `image_url` content blocks, so the shared chat pipeline (adapters, compression, budget, routing) handles them exactly like `/v1/chat/completions`: `requireVision` routing skips text-only models — including a pinned/sticky one — and falls back to a vision-capable peer; a request is rejected up front with `no_vision_model` only when no vision model is enabled. An alignment pass against the official Responses input spec folds `refusal` parts into text, carries the `detail` hint through, and rejects `file_id`-only images (no Files backend here) with a clear 422 instead of dropping them and answering blind.

**Why:** Codex requires `wire_api="responses"`, so its users had no image path at all — this was the root gap every later commit builds on.

### 2. `fix(google)`: fall back to Google's documented dummy thought_signature

**What:** Gemini 3 strictly validates `thoughtSignature` on `functionCall` parts in replayed history and 400s (`Function call is missing a thought_signature`) when it's absent. OpenAI-format clients can't carry the field, so this provider caches signatures keyed by call id and name+args; when both the client-provided and cached signatures are missing (restart, 30-min TTL expiry, history first produced by another provider), the part now gets `context_engineering_is_the_way_to_go` — one of exactly two sentinel strings Google officially documents for calls the API didn't generate.

**Why:** Signature-less multi-turn tool conversations through Gemini 3 were hard 400s. A fabricated value (e.g. a hash) is not accepted — the signature is an encrypted blob the server validates — so the documented sentinel (server-side "skip validation", at some reasoning-quality cost) is the only correct fallback.

### 3. `fix(server)`: raise the inference JSON body limit for vision payloads

**What:** The inference surfaces (`/v1`, `/v1beta`, `/mcp`, Ollama `/api/*`) get their own JSON ceiling — `REQUEST_BODY_LIMIT_MB`, default 50MB — while the dashboard/admin surface keeps 10mb. Over-limit bodies are normalized to an OpenAI-style 413 (`invalid_request_error` / `request_too_large`, with size, limit, and the env knob in the message) and logged as an error row in request analytics.

**Why:** Vision turns replay the ENTIRE history with base64 images (~33% inflation) — the motivating session hit 11.85MB — and the old global 10mb ceiling rejected that **before auth/routing**: an opaque `PayloadTooLargeError` to the client, no fallback attempt, no analytics row, nothing in the dashboard. The ceiling only gates parsing, so it must sit above the largest raw payload a client may send.

### 4. `feat(server)`: normalize inbound images before routing

**What:** Data-URL images over `IMAGE_NORMALIZE_THRESHOLD_KB` (default 1MB raw) are decoded with sharp, downscaled to fit `IMAGE_NORMALIZE_MAX_DIMENSION` (default 2048 — aligned with OpenAI's internal ceiling), and re-encoded — JPEG q90 by default (clamped 1-100), or PNG only when the alpha channel actually carries transparency (`stats().isOpaque`). Also normalizes formats most upstreams reject (bmp/gif/tiff/avif/webp → jpeg/png; gif keeps its first frame); anything undecodable (notably HEIC) or corrupt rides through untouched. Runs on every chat-shaped surface — `/v1/chat/completions`, `/v1/responses`, `/v1/messages`, and the inbound-chat funnel behind the Ollama/Gemini wire routes — before estimation, budgeting and routing. `IMAGE_NORMALIZE=off` disables it.

**Why:** Raising the body limit lands the requests, but the bytes still travel: every upstream resizes internally anyway, so pixels beyond the cap are pure transport waste that counts against payload limits (Gemini ~20MB/request, Anthropic 5MB/image) and our own buffering — and history replay stacks the same screenshots turn after turn (Codex sends pasted images verbatim; verified against a real session). Measured on the motivating screenshots: 1.95MB → 0.36MB (-81%) and 3.78MB → 0.44MB (-88%); the alpha check matters because macOS screenshots are RGBA-but-opaque, and honoring the bare `hasAlpha` flag kept them on PNG at 1.9MB.

## Verification

- Full server suite green: 207 files / 2271 tests passed.
- New tests cover: image translation + `requireVision` routing + both 422 pre-checks, the sentinel fallback (uniquely keyed to bypass the module cache), body-limit acceptance of an 11.5MB payload + normalized/logging 413, and the normalizer's resize / opaque-flatten / real-alpha / format-conversion / recompress / quality-clamp / disabled / corrupt paths.
- All knobs documented in `.env.example`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -39,12 +39,12 @@ PORT=3001
 # ADMIN_RATE_LIMIT_RPM=600
 
 # JSON body limit in MB for the inference surfaces (/v1, /v1beta, /mcp,
-# Ollama /api/*) — default: 50. Vision requests embed base64 images in the
+# Ollama /api/*) — default: 25. Vision requests embed base64 images in the
 # body and replay them with every turn, so large sessions can clear 10MB;
 # this ceiling only gates parsing — inbound image normalization (below)
 # shrinks the payload afterwards. Raise it if bigger payloads are rejected
 # with 413 'request_too_large'.
-# REQUEST_BODY_LIMIT_MB=50
+# REQUEST_BODY_LIMIT_MB=25
 
 # Inbound image normalization: data-URL images over the threshold are
 # downscaled to the long-edge cap and re-encoded (JPEG, or PNG only when the
--- a/.env.example
+++ b/.env.example
@@ -41,8 +41,23 @@ PORT=3001
 # JSON body limit in MB for the inference surfaces (/v1, /v1beta, /mcp,
 # Ollama /api/*) — default: 50. Vision requests embed base64 images in the
 # body and replay them with every turn, so large sessions can clear 10MB;
-# raise this if bigger payloads are rejected with 413 'request_too_large'.
+# this ceiling only gates parsing — inbound image normalization (below)
+# shrinks the payload afterwards. Raise it if bigger payloads are rejected
+# with 413 'request_too_large'.
 # REQUEST_BODY_LIMIT_MB=50
+
+# Inbound image normalization: data-URL images over the threshold are
+# downscaled to the long-edge cap and re-encoded (JPEG, or PNG only when the
+# alpha channel carries real transparency) before routing, shrinking replayed
+# screenshots ~6-10x. Also normalizes exotic formats upstreams reject
+# (bmp/gif/tiff/avif/webp → jpeg/png); gif keeps its first frame only. Every
+# upstream resizes internally anyway (OpenAI to 2048px, Anthropic to 1568px),
+# so pixels beyond the cap are transport waste. Set IMAGE_NORMALIZE=off to
+# disable entirely.
+# IMAGE_NORMALIZE=on
+# IMAGE_NORMALIZE_MAX_DIMENSION=2048
+# IMAGE_NORMALIZE_THRESHOLD_KB=1024
+# IMAGE_NORMALIZE_QUALITY=90
 
 # Outbound proxy for provider requests. Normally set in the dashboard
 # (Keys → Outbound proxy); these env vars are for headless installs.
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,12 @@ PORT=3001
 # tighter cap. Set to 0 to disable.
 # ADMIN_RATE_LIMIT_RPM=600
 
+# JSON body limit in MB for the inference surfaces (/v1, /v1beta, /mcp,
+# Ollama /api/*) — default: 50. Vision requests embed base64 images in the
+# body and replay them with every turn, so large sessions can clear 10MB;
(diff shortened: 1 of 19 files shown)


tashfeenahmed:
Merged — thanks, this is a solid piece of work, and the four-commits-one-session framing made it easy to review each part on its own terms.

I stacked one hardening commit on top before merging, all three items behind the same concern (a native dep and a memory ceiling on the small hosts this runs on):

1. **`sharp` now loads lazily.** A static `import sharp from 'sharp'` makes a missing or mismatched prebuild an *import-time* crash, so the whole gateway would fail to boot over an optional image optimization. Concretely: `desktop/electron-builder.yml` only `asarUnpack`s `better-sqlite3`'s `.node` files, so sharp's native binary and its libvips `.dylib`s would ship *inside* the asar where they can't be dlopen'd — the desktop app would have died on startup. It now resolves on first use and degrades to pass-through (images ride through at original size, exactly as before the module existed), warning once per process. `image-normalize-no-sharp.test.ts` covers that path.

   Follow-up if you want normalization active in the desktop build too: an `asarUnpack` entry for sharp plus adding it to `rebuild:native`. Not a blocker — degraded is the right default there.

2. **`REQUEST_BODY_LIMIT_MB` defaults to 25 instead of 50.** `express.json` buffers the whole body in memory and the limit is per request, so what actually has to fit in RAM is a handful of concurrent maximal bodies — 50MB isn't survivable on the Pi / small-VPS installs this targets. 25MB still clears your observed 11.85MB session with room to spare, and your normalizer shrinks payloads 6-10x right after the parse anyway. Raising it stays a one-liner for anyone on a bigger box.

3. **The Gemini skip-validation sentinel now logs once per process.** It's a silent quality trade, and a steady stream of them means the signature cache is missing (restart loop, TTL too short, history minted elsewhere) rather than a one-off replay — without a log there's nothing to correlate degraded tool-calling against.

Rebased over #854/#855/#856, which landed just ahead of this and touch the same streaming blocks in `proxy.ts` / `responses.ts` / `anthropic.ts` — clean, no conflicts. Full suite green: 209 files, 2287 passed, 5 skipped.

The alpha-vs-`isOpaque` detail in particular was worth catching; that's the kind of thing that only shows up against a real payload.

PR #853, opened 2026-08-11
feat(providers): add OrcaRouter as an OpenAI-compatible gateway provider

XiaoHuo888-hue:
## Summary

This registers [OrcaRouter](https://www.orcarouter.ai) the same way the existing OpenRouter provider is wired, so the model picker, config reference and docs stay consistent. OrcaRouter is an OpenAI-compatible gateway: one `ORCAROUTER_API_KEY` (keys start with `sk-orca-`) unlocks 150+ models from OpenAI, Anthropic, Google, DeepSeek, Qwen, MiniMax and xAI behind a single `https://api.orcarouter.ai/v1` endpoint. It also runs gateway-level, zero-trust security for AI agents on the same endpoint - screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

I'm an engineer on the OrcaRouter team.

## What this changes

- `shared/types.ts`: adds `orcarouter` to the `Platform` union
- `server/src/providers/index.ts`: registers the OrcaRouter `OpenAICompatProvider` (base URL `https://api.orcarouter.ai/v1`, OpenRouter-style attribution headers)
- `server/src/lib/key-parser.ts`: detects `ORCAROUTER_` prefixes and `orcarouter` / `orca-router` auth.json names
- `server/src/routes/keys.ts`: accepts `orcarouter` in the add-key platform list
- `server/src/services/provider-quota.ts`: pools the OrcaRouter free-tier models (`orcarouter/fusion*`, `*-free`) under one capacity pool, mirroring the observed shared 429 behavior
- `client/src/components/keys/shared.tsx`: Keys page dropdown entry with a "Get API key" link
- `client/src/lib/routing.ts`: platform color for OrcaRouter
- tests: key-parser, provider-instance, and quota-pool coverage

Model rows are not seeded here or in migrations. They ship through the hosted catalog and are gated on `hasProvider` by catalog-sync, the same mechanism the recent AnyAPI provider (#772) uses.

## Verification

- Unit tests: full server suite passes (2246 tests, 0 failures)
- Typecheck: `tsc` clean in the server, client, and cli workspaces
- Live API: `orcarouter/fusion`, `orcarouter/fusion-flash`, and `orcarouter/fusion-mini` all return 200 against `https://api.orcarouter.ai/v1/chat/completions`
- Ran locally: booted the gateway with an OrcaRouter key via declarative config, then `POST /v1/chat/completions` with `model: orcarouter/fusion-flash` returned a real completion; `/v1/models` lists the model; a nonexistent model id returns a clear 400


tashfeenahmed:
Thanks for this, and for disclosing your affiliation up front. The code is clean: a faithful mirror of the OpenRouter registration, every file on the new-platform checklist, tests in the right places. I'm closing it on the project's inclusion rule, not on the diff.

**The rule:** FreeLLMAPI only lists providers that grant recurring free credits, meaning an allowance that refills on its own and that someone who has never paid can actually use. OrcaRouter's Hacker tier is free *routing*. Your pricing page puts it plainly: "You pay providers directly at their published rates." That's an honest proposition, just not the thing this catalog indexes.

**What I measured.** Your `/routing/free-models` docs do describe a recurring free tier, so I tested it rather than ruling it out on the pricing page. Real key, $0 wallet, never topped up (the profile of essentially every FreeLLMAPI user):

**0 successful calls out of 9 attempts across 6 model ids.**

`deepseek/deepseek-v4-flash-free`, `deepseek/deepseek-v4-pro-free` and `orcarouter/free` each returned `429 free_rate_limited` on the first request of the session, `Retry-After` pinned to the next 00:00 UTC (~16h). Six retries over two minutes decremented in step with wall clock, always the same instant, so it's the daily bucket rather than a transient blip. Request id `202608120746415219452908268d9d66AemoYGq`. That matches what your docs say the lowest tier gets: "a deliberately small daily allowance," tiered by lifetime spend, numbers unpublished. For someone who will never top up, it reads as zero.

**One correction.** A $0 wallet separates billed from free cleanly: billed ids 403 `insufficient_user_quota`, free ids 429.

| id | result on a $0 wallet |
|---|---|
| `orcarouter/fusion` | 403, billed |
| `orcarouter/fusion-flash` | 403, billed |
| `orcarouter/fusion-mini` | 403, billed |
| `orcarouter/auto` | 403, billed |
| `orcarouter/free` | 429, free tier |
| `deepseek/deepseek-v4-*-free` | 429, free tier |

So `fusion*` is billed. The PR calls it "the free chat models, live 200 on 2026-08-12" and pools it under `orcarouter::free` in `provider-quota.ts`. You likely tested from a funded workspace, where it does return 200.

**Also:** the two `-free` ids are DeepSeek V4 Flash and Pro, which the catalog already carries from seven and four other providers respectively, several with published numeric limits and one at 1M context. No new capability for users, just another route.

Not a permanent no. If the free line ever carries a model the catalog lacks, and serves a never-topped-up workspace with a limit that can be written as a number, reopen and I'll retest exactly as I did here.

PR #858, opened 2026-08-11
chore(server): assert the pure lib modules never gain a value import

OhOkThisIsFine:
Test-only. No source change.

## Why

Several `lib/` modules are pure by design — functions over their arguments, no I/O, no DB, no config. `error-classify.ts` states it in its own header, along with the reason:

> Pure functions over an error's message/status — no I/O — so they live in a neutral lib module that any of those can import without forming an import cycle (fusion ↔ proxy in particular).

That is a real invariant with a real failure mode behind it, and nothing enforces it. Adding `import { getDb } from '../db/index.js'` to one of these modules compiles cleanly, passes every existing test, and quietly converts a leaf into a hub. The damage shows up later and somewhere else: a reintroduced import cycle, a DB read on a path documented as never-throwing, or a unit test that suddenly needs a database to run.

Same rationale as the existing `registry-drift.test.ts` — make the quiet no-op a failing test.

## What it checks

No **value** imports in: `budget.ts`, `error-classify.ts`, `header-value.ts`, `structured-output.ts`, `tool-args.ts`, `tool-call-rescue.ts`, `client-classifier.ts`, `content.ts`, `think-tags.ts`.

Type imports stay legal, including the `import { type A, type B } from …` form. They are erased before they can create a cycle or pull in a runtime dependency — which is why the three modules that already use type-only imports are *on* the list rather than excluded from it. Bare side-effect imports (`import './x.js'`) and `require(...)` count as value dependencies and are caught.

A second case asserts every listed file still exists, so a rename cannot leave the list silently guarding nothing.

## Verified to fail

A guard that cannot fail is not a guard. Adding `import { getDb } from '../db/index.js'` to `budget.ts` fails the run:

```
 ✗ budget.ts has no value imports
 Test Files  1 failed (1)
```

and removing it passes again. The detector also has its own cases covering type-only, mixed-named, side-effect and `require` forms.

Full server suite green: 206 files, 2256 passed, 8 skipped.

### DIFF
--- a/server/src/__tests__/lib/module-purity.test.ts
+++ b/server/src/__tests__/lib/module-purity.test.ts
@@ -37,18 +37,46 @@ const PURE_MODULES = [
   'think-tags.ts',
 ];
 
+/** True when every specifier inside `{...}` is type-only, so the whole
+ *  statement is erased. */
+function allSpecifiersAreTypes(specifiers: string): boolean {
+  return specifiers.split(',').every(s => s.trim() === '' || /^type\s/.test(s.trim()));
+}
+
 /** Import statements that survive compilation, i.e. everything except
  *  `import type ...`. Also catches bare side-effect imports (`import './x.js'`)
- *  and `require(...)`, which are value dependencies just as much. */
+ *  and `require(...)`, which are value dependencies just as much.
+ *
+ *  Three forms beyond a plain static import are load-bearing here, because each
+ *  creates the same runtime dependency while looking nothing like `^import`:
+ *  a dynamic `await import('./x.js')`, and re-exports of the form
+ *  `export { x } from './x.js'` / `export * from './x.js'` — a re-export is an
+ *  import that also republishes, so it forms a cycle just as readily. */
 function valueImportsIn(source: string): string[] {
   const found: string[] = [];
   for (const raw of source.split('\n')) {
     const line = raw.trim();
     if (line.startsWith('//') || line.startsWith('*')) continue;
+
+    // Dynamic import — `const m = await import('./x.js')` never starts with
+    // `import`, so the anchored rules below cannot see it. `import.meta.url`
+    // is not a call and does not match.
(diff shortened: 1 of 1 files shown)


suantea:
Good idea — the same "make the quiet no-op a failing test" pattern as registry-drift.test.ts, applied to the purity invariant. A few notes:

1. **Scope check**: the list is hardcoded here, so a NEW pure lib module added later won't be covered until someone remembers to add it. Consider failing the test (or at least warning) when a `lib/*.ts` module isn't on either list — that makes the guard self-maintaining instead of drifting. (If you deliberately don't want that, a comment saying "new pure modules must be added here" is the minimum.)
2. **TYPE-only imports excluded correctly** — good call, `import type` is erased and can't cycle.
3. **CI coverage**: since this is test-only, confirm it runs in the same vitest invocation as the rest of the suite (it appears to be a normal test file — just checking it's not silently skipped by a pattern).

Nice, low-risk change otherwise.

OhOkThisIsFine:
Thanks — `de08239` makes it self-maintaining, and you were right that it would drift.

An import-free `lib/*.ts` module must now be on one of two lists or the test fails. Purity is computed from the source, so only modules that *are* import-free need classifying — adding one with dependencies costs nothing.

Applying it immediately surfaced **13 modules already pure in fact and unguarded** (`config.ts`, `key-parser.ts`, `served-model.ts`, `wake-detect.ts`, …) — the drift you predicted, already there. I put them on `NOT_GUARDED` rather than `PURE_MODULES`: nothing depends on them staying import-free, and freezing a module nobody has argued should be frozen buys a false failure the first time one legitimately needs a dependency. Promoting any is a one-line move between lists.

Verified the guard can actually fail — a temporary import-free `lib/` module is reported by name with both options spelled out.

**CI**: confirmed, not skipped. `server/vitest.config.ts` uses `include: ['src/__tests__/**/*.test.ts']` and the file sits at `src/__tests__/lib/module-purity.test.ts`.

PR #861, opened 2026-08-12
fix(cli): resolve --model against the unfiltered catalog before launching

OhOkThisIsFine:
## The ambiguity

`catalog()` fetches `/v1/models?available=true`, so a model absent from that roster has **two** possible meanings: it does not exist, or it exists and is out of quota right now. The CLI treated them alike — and picked the wrong half. A pinned model reads as a typo the moment its provider rate-limits.

`resolveLaunchModel()` (new `cli/src/models.ts`) validates against the **unfiltered** roster and makes "registered but not currently available" its own outcome: the launch proceeds with a warning on stderr, because the router may recover the model mid-session, but it never proceeds silently.

The unfiltered fetch is best-effort — any failure degrades to the filtered roster rather than blocking a launch, so an older gateway that ignores the parameter still works.

## Two defects fall out of the same seam

**`--model` never reached the setup generators.** It was parsed into `CliOptions` and then dropped, so `setup-claude --model X` wrote whatever `primaryModel()` preferred and said nothing about ignoring the flag. It now threads through `GenerateContext` as `requestedModelId` and overrides the per-generator heuristic. A requested id missing from the available roster is still pinned — it was validated against the full catalog upstream, so it is out-of-quota, not a typo, and substituting a different model would be the wrong repair.

**`?? 128_000` was invented for any model publishing no window** and pinned as `CLAUDE_CODE_AUTO_COMPACT_WINDOW`. A wrong window is worse than none: too high and the client compacts after the gateway has already rejected the request; too low and it compacts a conversation that still fits. The variable is now absent when the catalog states nothing, and the client applies its own default.

## One thing deliberately not unified

The launcher's default stays "first servable concrete model", **not** the setup generators' preference for `auto`. A generated config is long-lived and should let the router choose per request; a launch pins one id into `ANTHROPIC_MODEL` for the whole session, where naming a concrete model is what makes that session reproducible. `models.ts` documents why they differ — I unified them by accident first, and two existing tests caught it.

## Testing

New `cli/src/models.test.ts` (10 cases) covers the unavailable-vs-unknown split, the `auto` special case, the degraded single-roster path, and the undefined-window contract. Two cases added to `tools.test.ts` for `--model` reaching the generators.

Windows-local: **41 passed, 15 failed**. Those 15 are the pre-existing POSIX-separator assertions in `tools.test.ts` (`expected '\home\tester\...' to be '/home/tester/...'`), identical on unmodified `main` on this machine and green on Linux CI. Verified by diffing failing-test names against a stashed baseline.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -38,6 +38,8 @@
     "dist/index.d.ts",
     "dist/config-files.js",
     "dist/config-files.d.ts",
+    "dist/models.js",
+    "dist/models.d.ts",
     "dist/tools.js",
     "dist/tools.d.ts",
     "dist/types.js",
--- a/cli/src/index.test.ts
+++ b/cli/src/index.test.ts
@@ -2,7 +2,8 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 import { afterEach, describe, expect, it } from 'vitest';
-import { claudeLaunchEnv, codexArgs, parseArgs } from './index.js';
+import { claudeLaunchEnv, codexArgs, parseArgs, resolvePinnedModel } from './index.js';
+import { UnknownModelError } from './models.js';
 import type { CatalogModel } from './types.js';
 
 describe('CLI arguments and launchers', () => {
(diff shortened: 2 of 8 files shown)


suantea:
Good fix — validating `--model` against the unfiltered roster is the right way to separate "no such model" from "exists but out of quota", and the warn-not-fail behavior (router may recover mid-session) matches how the router actually works. A few observations:

1. **`catalogs()` failure mode**: the unfiltered fetch is best-effort with a degraded fallback to the available-only roster — good. Worth confirming the degraded path still *warns* (not silently proceeds), since silently degrading to the filtered roster would reintroduce exactly the ambiguity this PR removes.
2. **`resolveLaunchModel` vs `warnIfUnavailable` split**: two passes over the same rosters (one to resolve the id, one to warn). Fine, but a single function returning both the resolved id and an availability flag would avoid the second call and keep the two verdicts in lockstep.
3. **Tests**: the diff is 495 lines — I assume `cli/src/models.test.ts` covers "registered but not available → proceeds with warning" and "not registered → error". If not, that's the case that locks the behavior in.

Looks solid otherwise — this one plugs a real UX trap for pinned models.

OhOkThisIsFine:
Good catches — fixed in `c93a24c`.

**1. Degraded `catalogs()`**: you were right, it did *not* warn. The `catch` silently returned the filtered roster as `full`, which reinstates exactly the ambiguity the second fetch removes. It now records `degradedReason` and reports it on stderr — but only when a `--model` is pinned, since without one the launcher picks from the filtered roster and there is no ambiguity to lose.

**2. Two passes**: agreed. `warnIfUnavailable` is gone; one `resolvePinnedModel` call per site now feeds both the pinned id and the warning, so they can't disagree. One deliberate exception: `launchClaude` still lets `claudeLaunchEnv` resolve internally — it's a pure call over the same rosters and can't reach a different verdict, and keeping the env builder side-effect-free seemed worth more than eliding it.

**3. Tests**: already covered — `models.test.ts:23-28` for registered-but-unavailable, `:30-33` for absent-from-both-rosters throwing. Added tests for the warning paths around them.

Also fixed while in here: `cli/package.json`'s `files` array was missing `dist/models.js`, so the published tarball would have failed at import.

Issue #433, opened 2026-07-01
feat: /livez + /readyz + /v1/providers — meta-gateway interop (OmniRoute #5687, LibreChat, LiteLLM Proxy)


chirag127:
## Summary

FreeLLMAPI is a great candidate to sit **behind** a bigger meta-gateway (OmniRoute, LibreChat, LiteLLM Proxy, my own gateway) as one of many upstream lanes. Today that integration is friction-heavy because:

1. **No stable `/health` (or `/livez` + `/readyz`) endpoint** documented — orchestrators need it for auto-failover / load-balancer probing. `README.md` mentions "Health checks — Periodic probes" but that's the internal upstream-probe, not an exposed liveness endpoint.
2. **No `/v1/providers` (or equivalent) metadata endpoint** to enumerate which of the 16 upstreams are configured, healthy, and how much budget is left. A meta-gateway wants to skip a freellmapi instance whose 16 upstream tiers are all rate-limited today.
3. **`GET /v1/models`** returns model IDs but not the free-tier budget signal (RPM/RPD/TPM/TPD remaining) that would let a meta-gateway prefer this freellmapi vs. another aggregator.

## Why this matters (concrete integrator)

I've filed [OmniRoute #5687](https://github.com/diegosouzapw/OmniRoute/issues/5687) proposing OmniRoute add freellmapi as a no-key fallback provider (14K stars, ~1.7B tok/mo across 16 tiers). For OmniRoute to route intelligently, it needs to know:

- Is this freellmapi instance up? → `/health` or `/livez`
- Are upstream tiers hot or all cooling down? → `/v1/providers` or aggregate signal
- Which models can this instance actually serve *right now*? → `GET /v1/models?ready=true` filter, OR budget-aware model list

Without these, the meta-gateway either:
- Sends every request blindly and eats N failovers per call (bad UX), OR
- Excludes freellmapi from the pool entirely (loses the ~1.7B tok/mo)

## Proposal

Three small endpoints, all read-only, no auth escalation:

### 1. `GET /livez`

Returns 200 + `{"status": "ok", "version": "x.y.z", "uptime_s": N}` if server is up.
Returns 503 if DB is unreachable or `ENCRYPTION_KEY` isn't loaded.

### 2. `GET /readyz`

Returns 200 if at least one upstream provider is `healthy`.
Returns 503 with `{"reason": "all_upstreams_rate_limited"}` otherwise.
Meta-gateways use this to decide "route to this instance or skip".

[...]

PR #862, opened 2026-08-12
feat(cli): freellmapi doctor — prove a client's requests reach this gateway

OhOkThisIsFine:
> **Stacked on #861.** This branch contains that commit plus one. Review the second commit (`849b460`) only; the diff collapses to just that once #861 merges. Happy to rebase if you'd rather take them in the other order.

## Why the server cannot answer this

Users believe they are routed through the gateway when they are not, and nothing tells them — the failure is a request that *never arrives*, so there is no server-side log to inspect. The verdict has to be computed in a process that **inherits the session environment**, resolving each tool's endpoint the way that tool resolves it.

## Verdicts

| verdict | meaning |
|---|---|
| `routed` | reaches **this** gateway |
| `elsewhere` | reaches something else, or something that is not this gateway |
| `shadowed` | a higher-precedence layer overrides the one that was configured |
| `unreachable` | configured endpoint answers nothing |
| `unknown` | nothing configures the tool, or doctor does not model it |

Exit code is 0 only when every tool is routed, so it works as a script precondition.

**`elsewhere` was added after the first live run.** With only the original four verdicts the command printed `OK claude: routed` for a Claude Code pinned to `api.anthropic.com` — true in the narrow sense, and exactly the confident-wrong answer to the question actually being asked. `routed` now means "reaches this gateway", and a regression test pins it.

## Precedence, per tool

- **claude** — process env `ANTHROPIC_BASE_URL` **over** the `~/.claude/settings.json` `env` block. That direction is the motivating failure: a launcher that exports the variable silently beats a settings file the user edited, while the block's *other* keys still apply, which makes it nearly invisible by reading config alone.
- **codex** — the *selected* provider from `model_provider`, then that entry's `base_url` in either the `[model_providers.NAME]` or dotted spelling, honoring `CODEX_HOME`. Reading any `base_url` in the file would report a provider the user is not selecting.

## Identity is deliberately weak, and says so

`/livez` carries no product marker — it answers `{status, version, uptime_s}`, which another service could serve. Matching that shape is treated as *evidence, not proof*, and an unidentified responder is reported as `elsewhere` rather than assumed to be us. Strengthening this would need a marker the server does not currently emit; I'd rather add that in a separate PR than have `doctor` overclaim.

## Sample output

```
FAIL claude: elsewhere
     claude reaches https://api.anthropic.com, NOT the http://127.0.0.1:3001 this command was pointed at.
     -> ANTHROPIC_BASE_URL (process environment) = https://api.anthropic.com
OK   codex: routed
     codex reaches http://127.0.0.1:3001 (gateway v0.7.0).
     -> C:\Users\...\.codex\config.toml ([model_providers.freellmapi] base_url) = http://127.0.0.1:3001/v1
```

## Notes

`parseArgs` now collects trailing positionals instead of rejecting the second one; commands that take none still reject extras, so `setup-claude typo` stays an error rather than a silently ignored word.

18 new tests in `cli/src/doctor.test.ts` cover each verdict, both precedence orders, the dotted/section TOML spellings, `CODEX_HOME`, malformed settings, and the identity contract. Windows-local: **59 passed, 15 failed** — those 15 are the pre-existing POSIX-separator assertions in `tools.test.ts`, unchanged from `main` and green on Linux CI.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -38,6 +38,8 @@
     "dist/index.d.ts",
     "dist/config-files.js",
     "dist/config-files.d.ts",
+    "dist/doctor.js",
+    "dist/doctor.d.ts",
     "dist/models.js",
     "dist/models.d.ts",
     "dist/tools.js",
--- a/cli/src/doctor.test.ts
+++ b/cli/src/doctor.test.ts
@@ -1,0 +1,268 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { afterEach, describe, expect, it } from 'vitest';
+import {
+  claudeLayers,
+  codexLayers,
+  diagnose,
+  exitCodeFor,
+  normalizeUrl,
+  probeGateway,
+  sameGateway,
+} from './doctor.js';
+
+const temporary: string[] = [];
+function tempHome(): string {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'freellmapi-doctor-'));
+  temporary.push(dir);
+  return dir;
+}
+afterEach(() => {
+  for (const dir of temporary.splice(0)) fs.rmSync(dir, { recursive: true, force: true });
+});
+
+/** A fetch double that answers /livez exactly as the real gateway does. */
+function gatewayFetch(body: unknown = { status: 'ok', version: '0.7.0', uptime_s: 12 }): typeof fetch {
+  return (async () => ({ status: 200, json: async () => body })) as unknown as typeof fetch;
+}
+const deadFetch = (async () => { throw new Error('ECONNREFUSED'); }) as unknown as typeof fetch;
+
+function writeClaudeSettings(home: string, value: Record<string, unknown>): void {
+  const dir = path.join(home, '.claude');
+  fs.mkdirSync(dir, { recursive: true });
+  fs.writeFileSync(path.join(dir, 'settings.json'), JSON.stringify(value));
+}
+
+function writeCodexConfig(home: string, toml: string): void {
+  const dir = path.join(home, '.codex');
+  fs.mkdirSync(dir, { recursive: true });
+  fs.writeFileSync(path.join(dir, 'config.toml'), toml);
+}
+
+describe('url comparison', () => {
+  it('ignores trailing slashes, a /v1 suffix, and case', () => {
+    expect(normalizeUrl('http://Localhost:3000/v1/')).toBe('http://localhost:3000');
+    expect(sameGateway('http://localhost:3000', 'http://localhost:3000/v1')).toBe(true);
+    expect(sameGateway('http://localhost:3000', 'http://localhost:3001')).toBe(false);
+    expect(sameGateway(undefined, 'http://localhost:3000')).toBe(false);
+  });
+});
+
+describe('claude precedence', () => {
+  it('ranks a settings env block ABOVE the inherited shell environment', () => {
+    // The easy thing to get backwards, and I did: Claude Code writes each
+    // `env` entry into the process environment at startup, REPLACING what the
+    // shell exported. The shell is the lowest layer, not the highest.
+    const home = tempHome();
+    writeClaudeSettings(home, { env: { ANTHROPIC_BASE_URL: 'http://localhost:3000' } });
+    const layers = claudeLayers({ ANTHROPIC_BASE_URL: 'https://api.anthropic.com' }, home, tempHome());
+
+    expect(layers[0]).toMatchObject({ value: 'http://localhost:3000', effective: true });
+    expect(layers[1]).toMatchObject({ value: 'https://api.anthropic.com', effective: false });
+  });
+
+  it('uses the shell value when no settings file names one', () => {
+    expect(claudeLayers({ ANTHROPIC_BASE_URL: 'http://localhost:3000' }, tempHome(), tempHome())[0])
+      .toMatchObject({ value: 'http://localhost:3000', effective: true });
+  });
+
+  it('ranks project-local settings above the user scope', () => {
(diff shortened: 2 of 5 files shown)


OhOkThisIsFine:
Correction to the PR description above, now fixed in `626214f`: I had Claude Code's precedence **backwards**.

The original implementation ranked the process environment above the settings `env` block. The [documented behavior](https://code.claude.com/docs/en/env-vars) is the opposite — Claude Code writes each `env` entry into the process environment at startup and again when the file changes, *replacing* the value inherited from the shell. So the shell is the lowest layer, not the highest.

What misled me is a real observation that looks like a counter-example: a launcher pinning `ANTHROPIC_BASE_URL` visibly beats an edited `settings.json`. That happens because such a launcher writes **managed** settings, which outrank every other scope — and the rest of the same `env` block still applies, which is exactly why it is so hard to spot by reading config.

The layer list is now, highest first: managed settings (platform paths), project-local, project, user (honoring `CLAUDE_CONFIG_DIR`), then the inherited shell environment. Tests updated accordingly, including the `shadowed` case, which now correctly reports the settings block as the layer that wins.

Worth flagging for a reviewer: this is the failure mode the command exists to prevent, so it seemed better to state it plainly than to quietly amend.

suantea:
Nice, clean piece of work — the layer model (settings files > env > shell) and the "prove it reaches THIS gateway, not just a live host" /livez identity check are exactly the right shape, and the test coverage is thorough. A few observations:

**Verified against the server**
- `/livez` does exist (`routes/status.ts`, #433) and returns exactly `{ status, version, uptime_s }` — the three-field identity check in `probeGateway` matches the real contract. One nit: the field check doesn't assert `status === 'ok'`, just `typeof status === 'string'`, so a hypothetical `/livez` that returned `{ status: 'degraded', version, uptime_s }` would still be "identified". The server only ever emits `'ok'` today, but asserting the value would make the probe's verdict robust to future health states.

**Suggestions**
1. `probeGateway` hard-codes `AbortSignal.timeout(5_000)` — worth threading through `DoctorOptions` (or at least a constant) so slow networks aren't misreported as `unreachable`.
2. `normalizeUrl` drops `/v1` and trailing slashes but not a path prefix (e.g. a gateway mounted under `/gateway`). `sameGateway` would then treat `http://host/gateway` and `http://host/other` as equal — probably fine for the common case, but a `URL`-based normalize (pathname trim) would be more precise.
3. The `cli/` workspace is new here — does it get picked up by the root `npm test` / CI matrix, or does it need a separate workflow entry? Worth confirming the new tests actually run in CI (I don't see a workflow change in the diff).

Overall: solid — the "shadowed" verdict (reporting the layer that wins and what it beat) is a genuinely useful diagnostic. Happy to help test it against a real install once it merges.

OhOkThisIsFine:
Thanks — all three landed in `a67afe0`, with one amendment.

**`status === 'ok'`**: taking this one a different way. The server does emit a non-`ok` status: `routes/status.ts` returns `{status: 'unavailable', ...}` with a 503 when the DB or encryption key is down — and that response is still unmistakably this gateway. Requiring `ok` to *identify* it would report a correctly-routed but sick gateway as "probably a different service on that port", which is the exact wrong answer this command exists to prevent. So identity stays structural and health became its own axis, with a new `degraded` verdict (routing fine, gateway not) that exits non-zero. An unrecognized future status now degrades to a truthful "this gateway says `<status>`" instead of a lie — which I think gets you what you were after.

**Timeout**: threaded through `DoctorOptions` as `DEFAULT_PROBE_TIMEOUT_MS`, reachable as `doctor --timeout MS`. A non-numeric value errors rather than silently defaulting.

**`normalizeUrl`**: now `URL`-based, keeps the path prefix, case-folds only scheme and host. Fun edge case: `localhost:3000/v1` *does* parse, as an opaque path under a `localhost:` scheme — empty host now falls back to lexical trimming.

**CI**: already covered — `cli/` is on `main`, and root `npm test` / `npm run build` both include `-w cli`.

One thing I found while checking that: `cli/package.json`'s `files` array was missing `dist/doctor.js`, so the published tarball would have failed at import. Fixed here (and the same bug for `dist/models.js` in #861). `cli-release.yml`'s pack smoke-test would have caught it, but only at publish time.

@tashfeenahmed — the release workflow publishes only on a version bump, and both CLI PRs still say `0.3.0`. Left unbumped deliberately so the two don't conflict; happy to bump whichever you want to land first.

And yes please to testing against a real install.

tashfeenahmed:
Merged, along with #861 and #858. Thanks — the reasoning in these was easy to check, which made them quick to take.

I pushed two changes to the branch before merging:

1. The positional-argument check had moved below the `setup-*` dispatch, so `setup-claude typo` ran with the stray word silently ignored where it used to error. Your comment on `parseArgs` described the intended behaviour correctly — it was just the dispatcher ordering. Moved above dispatch, exempting `doctor`, with a regression test.

2. `managedSettingsPaths` read only `managed-settings.json` and missed the `managed-settings.d/` drop-in directory, whose `*.json` files merge alphabetically on top of the base. On a fleet-managed install with a drop-in, `doctor` would have named the layer the operator can see and missed the one actually winning. It now enumerates the drop-ins ranked above the base, alphabetically-last first.

Also rebased onto main since #861 landed first, which is why the CLI hunks in this branch collapsed.

One thing worth recording: the claim that a settings.json `env` block outranks an exported `ANTHROPIC_BASE_URL` is not something the settings docs state either way — they confirm the scope ordering and the platform paths, but not that interaction. Nothing contradicts it, and the `shadowed` output prints both layers with their values, so a reader can see the conflict even if the arrow ever points the wrong way. Worth a comment pointing at whatever you observed, in case it changes upstream.

Issue #825, opened 2026-08-10
routeViaProxy option is clunky


yfdyh000:
1. This option applies to groups, so you cannot set it separately for custom endpoints.
2. It is tucked away inside a menu, resulting in poor visibility, and two‑click operations are required for each individual items. Bulk enabling or disabling of multiple items in one go is not supported. By contrast, a checkbox‑based list would be a suitable alternative. Additional controls could also be provided to assign separate proxy servers for endpoints (or even keys, though I am not sure whether that makes practical sense). It would be great to add a test button.
3. The Chinese translation needs improvement. The current label "通过代理路由" (Route via Proxy) carries double‑meaning confusion, as both "proxy" and "agent" can be translated into the same Chinese word the "代理". Revise the translation to "使用代理服务器 (i.e. "Use Proxy Server") would be a proper alternative.

suantea:
Addressing your three points one by one:

1. **Custom endpoints can't be configured separately** — this is actually already solved: PR #819 (per-key proxy, merged) added an independent `proxy_url` per API key, so custom endpoints (and any individual key on any platform) can set their own proxy egress instead of being limited to the group-level switch.

2. **Visibility / bulk ops / test button** — agreed, `Route via proxy` tucked inside the menu is indeed too hidden. A full checkbox list + bulk toggle + proxy test button is a sizable UI refactor; I suggest tracking it in a separate issue (it touches provider-list layout, proxy connectivity probing, and bulk PATCH). I'll pick it up after #819 is merged and stable.

3. **Chinese translation** — fixed: `通过代理路由` → `经由代理转发` (zh-CN / zh-TW kept in sync), so "代理" no longer reads as an AI agent. The change is included in my in-flight PRs.

If you want point 2 to move faster, the "proxy test button" is the smallest and most directly valuable slice — we can extract that first.

yfdyh000:
经由代理转发 仍可能误解为经agent。

测试按钮与设置的文本框、保存按钮一起做比较好，也就是能测试已保存和未保存的代理？

suantea:
@yfdyh000 Thanks, got both points:

**1. Wording ambiguity**
"经由代理转发" could indeed still read "代理" as an AI agent. I'll revise it once more to **"经由代理服务器转发"** — "代理服务器" (proxy server) largely rules out the agent ambiguity (the tooltip will keep the word "Proxy" as well). zh-CN / zh-TW will be updated together.

**2. Test button**
[...]

PR #864, opened 2026-08-12
feat(proxy): add a Test button for the outbound proxy settings (#863)

### LINKED ISSUE #863: feat(proxy): add a proxy test button that covers saved and unsaved configs
## Background

Originated from the #825 discussion: the `routeViaProxy` (proxy egress) option currently lives in a group-level menu — poor visibility, requires two clicks per item, and there is no way to verify connectivity. yfdyh000 suggested "a test button next to the settings text box and save button", able to test **both saved and unsaved** proxy configs.

## Goal

- Add a **Proxy test button** on the same row as the proxy settings text box + save button — the smallest usable change, excluding the full provider-list refactor (tracked separately).
- On click, **prefer the draft value in the input**; fall back to **the saved value** when the draft is empty — covering both "edited but not saved" and "already saved" states.
- Show the result inline (success/failure + latency or error reason); no navigation, no stacked dialogs.

## Acceptance criteria

- [ ] With no edit made, clicking test probes connectivity using the saved `proxy_url`
- [ ] After editing the text box without saving, clicking test probes the draft value and does not write it to config
- [ ] Success shows latency; failure shows a clear error reason
- [ ] The button is disabled during the probe to prevent double-clicks, then re-enabled

## Related

- Upstream discussion: #825, #819 (per-key proxy, merged)
- Same UI improvement batch as #825's bulk-toggle / checkbox-list items; this issue tracks only the minimal test button

suantea:
## What

Backs the Settings → Outbound proxy **Test** button with a non-persisting connectivity probe (#863), per the discussion in #825.

- **`server/src/lib/proxy.ts`**: `probeProxyUrl()` builds a throwaway dispatcher for the **DRAFT** proxy URL (falling back to the saved value when the input is empty) and measures a round trip against a stable, unauthenticated endpoint (`api.openai.com/v1/models`). Any HTTP response — even a 401/403 without a key — proves the proxy route works; only a network-level failure (DNS, connect, timeout) is reported as failure, with latency and error reason returned.
- **`server/src/routes/settings.ts`**: `POST /api/settings/proxy/test` accepts the draft value in the body and **never persists anything**.
- **client** (`proxy-settings-section.tsx`): Test button on the same row as the proxy input + Save button; prefers the input's draft value, falls back to the saved one; result shown inline (latency on success, reason on failure); button disabled while probing.
- **i18n**: en / zh-CN / zh-TW keys.

## Verification

- server + client type checks pass (the only client error is a pre-existing `tsconfig` `baseUrl` deprecation notice)
- 100 existing proxy tests pass, no regression
- Smoke-tested `probeProxyUrl`: direct (timeout reported), dead proxy (`fetch failed`), invalid scheme (`Failed to build a proxy agent`) — all return structured `{ ok, latencyMs, error }`, never throw

Closes #863.


tashfeenahmed:
Merged. The Test button is a genuinely useful addition and the probe semantics are right — treating a 401 as a working route rather than a failure is the correct call, and the SOCKS/undici split is handled properly.

One fix on top before merging: the probe was hardcoded to `https://api.openai.com/v1/models`. That makes the verdict wrong in both directions. A self-hosted gateway that never talks to OpenAI now contacts it on every click, and on a network where that host is blocked the button reports a working proxy as broken — which is exactly the population most likely to be setting up a proxy in the first place.

The target now comes from the caller. `routes/settings.ts` picks the `/models` endpoint of a provider you actually hold an enabled key for, preferring one that isn't already bypassing the proxy so the probe exercises the proxy path. `PROXY_TEST_URL` overrides it, and only an install with no keys at all falls back to a neutral endpoint that is deliberately not an AI vendor. The result also carries the target back, and the dashboard shows its host next to the verdict, so a pass or a fail tells you what was actually reached.

Verified on a running instance with Google, Groq and Cerebras keys: the probe resolved to `api.cerebras.ai`, returned 403 (no valid key) and correctly reported that as a working route in 187ms. With the proxy pointed at a dead address it reported the failure instead.

Also worth flagging for #866, which builds on this branch: that PR currently deletes `POST /api/settings/proxy/test` while two components still call it, so both Test buttons 404. It'll need that route back after rebasing onto this.

PR #868, opened 2026-08-12
feat(ui): merge custom-provider model aliases into a unified model (#790)

### LINKED ISSUE #790: Allow merging model aliases of custom providers into the primary model id

suantea:
## What

Implements #790: custom providers can serve a model under a different id (an alias), and requests for the primary model id previously never reached it because the alias lives in its own unify group.

Adds a **"Merge model aliases"** section to the model detail page (`ModelDetailPage.tsx`):
- An input + **Add** button writes `{ into: this group's label, keys: [alias] }` into the `merges` list of `model_unify_overrides` — the same store the existing split control writes, no schema change.
- Aliases already targeting this group are listed with a **Remove** button; Enter submits.
- After merging, a request for the unified model id is served through the alias too (the router resolves the whole group).

**i18n**: en / zh-CN / zh-TW + 57 locales (`aliasMergeHeading` / `aliasMergeHint` / `aliasMergeAdd` / `aliasMergeRemove`).

## Verification

- client type check passes
- `check:i18n` passes (60 locales, 820 keys)

Closes #790.


tashfeenahmed:
Merging, with a fix on top for two bugs in the add/remove path — both from expressing an edit against the *visible* rows instead of the full overrides list:

1. **Adding a second alias dropped the first.** `addAlias` replaced the group entry with `keys: [key]`, discarding every alias already merged into that model.
2. **Remove deleted the wrong entry.** The rendered rows are filtered to this group, but the index they yield was applied to the unfiltered `merges` array — so removing the first row on screen could delete an unrelated model's alias.

The mapping now lives in `client/src/lib/alias-merge.ts` as pure `addAlias` / `removeAlias` / `aliasesFor` helpers keyed on the alias **value**, with unit tests covering the cross-group case (same file-layout convention as `fusion-filter.ts` from #883). The list also renders one row per alias instead of one comma-joined row per entry, so each Remove button names exactly what it removes.

Client suite 90/90 green, i18n check green across 60 locales, CI green.

PR #879, opened 2026-08-13
Add model and backup functionality

kentpan:


tashfeenahmed:
Thanks for this. The backups feature landed via #999 with you credited as co-author, with auth tables excluded from dumps, a transactional restore and a key fingerprint check. The rest of this PR (config page merge, model CRUD, deploy scripts) was left out; the deploy script in particular copies the live ENCRYPTION_KEY into the zip, so please do not ship that. Happy to look at the model additions as a separate small PR.

PR #882, opened 2026-08-14
fix(modelscope): stop health checks from burning magic-grain quota

ZER0-auto:
## Problem

ModelScope (魔搭) validates API keys with a **paid 1-token chat completion** because `GET /v1/models` does not enforce auth (returns 200 for any bearer token). Every `validateKey` therefore spends one request against the account's **magic-grain (魔粒) quota** — observed as **2 魔粒 per ultra-tier request** (e.g. `deepseek-ai/DeepSeek-V4-Flash-0731`), not the historical "2000-requests/day free API quota" the original comment assumed.

With the default 5-minute health pass, a single configured ModelScope key burns **~288 paid probes per day ≈ 576 魔粒/day** — far more than the ~250 魔粒/day a bound account earns (200 daily login + 50 Alibaba-cloud binding). Users who simply add a ModelScope key to route through the gateway find their 魔粒 draining while they are not even sending requests.

## Fix

Two mitigations in `validateKey`, both keeping proactive validation near zero cost:

1. **Token-shape pre-check (zero network cost)** — ModelScope SDK tokens always start with `ms-`. Anything else is rejected locally as invalid without spending a request.

2. **Per-key validation cache** — a successful validation is cached for `MODELSCOPE_VALIDATE_CACHE_MS` (default **24h**). Repeat health passes return `true` without re-probing; the daily drain drops from ~288 probes to ~1.

A revoked key is **still caught** by the next real request's 401 handling (`error-classify` disables it), so the cache only delays proactive detection — it never hides a live failure. Failed validations are intentionally **not** cached, so a genuinely invalid key is re-probed every pass.

## Verification

- Added 3 tests: non-`ms-` token rejected with zero network calls; successful validation cached per key (2nd call makes no HTTP request); failed validation not cached (2nd call re-probes).
- `npx tsc --noEmit` passes.
- `vitest run` modelscope suite: 9/9 pass.
- Pre-existing failures in `settings-proxy` / `csp-inline-bootstrap` / `abort-signal` / `proxy-stream-integrity` reproduce on unmodified `main` (local env issue, unrelated to this change).

## Config

| Env var | Default | Meaning |
|---|---|---|
| `MODELSCOPE_VALIDATE_CACHE_MS` | `86400000` (24h) | Validation cache TTL; `0` = probe on every health pass (old behavior) |


tashfeenahmed:
Thanks — the quota drain here is real and the 24h validation cache is the right fix. Merging with two edits on top:

1. **Dropped the `ms-` shape pre-check.** It rejects keys locally on an assumption none of us can verify — the class comment says outright that we have no real ModelScope token. If the platform ever mints a token in another shape, a valid key gets a hard, confusing failure. The cache is what actually takes the health pass from ~288 paid probes per key per day down to ~1, so the shape rule buys nothing on valid keys and risks locking out real ones.

2. **Re-keyed the cache on a sha256 of the token instead of `keyId`.** As written, editing a key value in place kept the *old* token's verdict for up to 24h. Fingerprinting the secret means rotation re-probes, no plaintext is retained in the map, and the cache no longer depends on a quota context being passed.

Test changes follow: the shape-rule case is gone, and there is a new case asserting a different token on the same key row re-probes. CI green.

PR #884, opened 2026-08-14
feat(fusion): support vision (image) requests via vision-capable panel

Icesenator:
## What

Fusion now accepts **image (vision) requests** end-to-end. Previously `model: "fusion"` with an image returned `422 fusion_no_vision`; now:

- The panel is restricted to **vision-capable** models (`supportsVision`), for both explicit (`fusion.models`) and auto panel selection.
- The synthesis judge receives **text-only** messages (image blocks stripped), since it blends the panel's text answers — no vision requirement on the judge.
- An empty vision panel raises a clear 400 `FusionError` (with a vision-specific hint) instead of a 422.

## Why

`fusion` is the natural "multi-model synthesis" knob, and image requests are already routed to vision models by the auto-router (`supports_vision` + fallback chain). Leaving `fusion` blocked for images made it unusable exactly where a panel of vision models behind **independent provider quotas** shines: one model rate-limited (429) or down (502) is skipped for the next one, with the router's existing cooldown/fallback machinery.

## How

- `routes/proxy.ts`: removed the up-front `fusion_no_vision` 422; `hasImage` is propagated as `vision` to both streaming and non-streaming `runFusion` calls. The existing up-front `no_vision_model` gate (no vision model enabled at all) is untouched.
- `services/fusion.ts`: `runFusion({ vision })` → `selectPanel({ requireVision })` filters candidates on `supportsVision` (explicit + auto chain); `buildJudgeMessages(..., vision)` strips image blocks from the judge's input.
- `lib/content.ts`: new `stripImagesFromMessages()` sharing the same block-type set as `contentHasImage`.

## Tests

- `proxy-vision.test.ts`: fusion + image no longer 422s (`not.toBe(422)`, `not.toBe('fusion_no_vision')`).
- `services/fusion.test.ts`: `selectPanel` drops non-vision models (explicit) and keeps only vision models (auto); `buildJudgeMessages` strips images when flagged and keeps them by default (backward compatible).
- `lib/content.test.ts`: `stripImagesFromMessages` drops image blocks, handles image-only → `''`, preserves untouched messages.

All new tests are TDD (red first). `tsc --noEmit` is clean.

## Notes

- The judge stays text-only by design: it synthesizes the panel's text answers, so pixels would only risk failing a text-only judge model.
- The tool-call fast-path (first panel survivor with `tool_calls` wins, judge skipped) is unchanged.
- Full `npm test -w server`: 2297 passed / 3 failed — the 3 failures are pre-existing on this Windows machine and unrelated (`compression` perf budgets, `csp-inline-bootstrap` hash). `cli` path tests also fail on Windows only (backslash vs forward slash); `client` 69/69.


tashfeenahmed:
Thanks — merging this.

One small edit on top before merge: the auto-chain filter tested \`supportsVision === 1\` while the explicit-panel branch two lines up tests \`supportsTools\` for truthiness. Both columns are SQLite 0/1, so behaviour is identical; I just matched the two capability gates so they read the same.

Verified locally on top of current main: full server suite 214 files / 2357 tests green, plus CI.

PR #917, opened 2026-08-18
fix(probe): raise the probe max_tokens above the floor relays enforce (#903)

### LINKED ISSUE #903: Probe request failed: max_tokens must be greater than 2
@suantea #730 
"探测失败：Probe request to deepseek-v4-flash failed: Custom (OpenAI-compatible) API error 400: max_tokens must be greater than 2 (request id: 202608172....MN)"

on
https://chat.b.ai/key
https://docs.b.ai/llmservice/api/

tashfeenahmed:
Closes #903.

## Problem

`probeEndpointModel` pinned `max_tokens: 1` to keep the probe cheap. Several relays enforce a floor above that and reject the request outright, so the probe reported a working endpoint as broken and left its cooldown in place.

Reported against b.ai's `deepseek-v4-flash`:

```
Probe request to deepseek-v4-flash failed: Custom (OpenAI-compatible)
API error 400: max_tokens must be greater than 2
```

## Fix

Raise it to 4, via an exported `PROBE_MAX_TOKENS`. That clears every floor seen so far and still costs a rounding error per probe.

Fixed generally rather than special-cased per provider: `max_tokens: 1` is needlessly aggressive, the floor is not unique to one relay, and 3 extra tokens is not a cost worth a branch.

## Verification

- server `tsc -b` clean
- `custom-model-discovery` 21/21
- The existing test pinned the literal `1`, so it now asserts the constant. A new case drives a mock upstream that 400s anything at or below 2, which fails on the old value and passes on the new one.

## Not included

`providers/modelscope.ts` uses `max_tokens: 1` for key validation against ModelScope's own models, where it has been working. Left alone so this stays one change; worth revisiting if that provider ever grows a floor.

### DIFF
--- a/server/src/__tests__/routes/custom-model-discovery.test.ts
+++ b/server/src/__tests__/routes/custom-model-discovery.test.ts
@@ -3,6 +3,7 @@ import type { Express } from 'express';
 import { createApp } from '../../app.js';
 import { initDb, getDb } from '../../db/index.js';
 import { setCooldown, isOnCooldown } from '../../services/ratelimit.js';
+import { PROBE_MAX_TOKENS } from '../../services/model-discovery.js';
 import { mintDashboardToken, isGatedApiPath } from '../helpers/auth.js';
 
 // #488: a relay's model list changes weekly, so the endpoint's own /models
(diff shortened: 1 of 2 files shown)


suantea (approved):
Reviewed the diff against the diagnosis in #903 — the fix is exactly right:

- `PROBE_MAX_TOKENS = 4` as an exported constant (not a magic number at the call site), with the "why not 1" documented in a comment;
- The regression test mirrors the actual upstream that reported the bug (400 on `max_tokens <= 2`), which would fail on the old value and pass on the new one;
- Keeping `providers/modelscope.ts` untouched keeps the change single-purpose.

Also verified locally that `server/src/services/model-discovery.ts` still has `max_tokens: 1` on our fork's base, so this change is genuinely needed and non-conflicting. Approving.

对照 #903 的诊断核对了 diff，修复完全正确：`PROBE_MAX_TOKENS` 导出常量（而非调用处的魔法数字）并注释说明为何不是 1；回归测试复现了实际报障的上游行为（`max_tokens <= 2` 即 400），旧值会失败、新值通过；不动 `modelscope.ts` 保持改动单一职责。本地 fork base 上仍是 `max_tokens: 1`，确认本改动必要且无冲突。同意合并。

Issue #876, opened 2026-08-12
Reveal remaining quotas in the model list and details page


yfdyh000:
Occasionally, I need to find‑out the availability and remaining quota of specific models (either unified or individually), including probing of availability and quota information (e.g. read the Retry‑After header).
At present, the "Monthly token budget" metric does not expose time‑based rate‑limit controls.
The router pressure status can generate lengthy entries when handling numerous models, and the meaning of a failed / penalty / cooldown is ambiguous.
Besides, there are neither visual cues nor text‑based indicators for models displayed in the list.

suantea:
I'll take this one. Plan: surface remaining quota on the model list and detail page — reuse the existing quota/limit signals (RPM/RPD/TPM/TPD + monthly token budget) where they are already computed, add a live availability probe indicator, and expose Retry-After based hints when a model/key is cooling down. Will open a focused PR. (Working on #878 first, then this.)

suantea:
PR #921 is ready: time-window rate-limit usage badges (RPM/RPD/TPM/TPD) in the model table + detail page, reusing existing quota signals. Mergeable, CI green.

PR #871, opened 2026-08-12
feat(analytics): show the key label on the failover ladder (#869)

suantea:
Snapshots the api key label into request_attempts and shows it as a tooltip on the ladder's key badge. Label only, never the id or the credential. Reduced from the original branch to just the #869 slice. Closes #869.

PR #921, opened 2026-08-19
feat(ui): 模型表格展示窗口限流用量徽章 / time-window rate-limit usage badges in the model table (#876)

suantea:
### 背景 / Background

`#876` 的月度预算徽章只反映月配额，对窗口限流（RPM/RPD/TPM）没有提示——而路由实际按窗口限流拒绝请求。本 PR 在模型表格组头新增**最紧张窗口用量徽章**（used/limit，红/琥珀/灰三档），把路由真正执行的约束展示给用户。

The monthly-budget badge from #876 says nothing about windowed RPM/RPD/TPM limits, yet those are the constraints the router actually enforces. This PR adds a **tightest-window usage badge** (used/limit, red/amber/muted) to the model table's group header, surfacing what the router would reject next.

### 改动 / Changes

1. **server**: 新增 `GET /api/fallback/rate-limit-usage` —— 每个启用模型按可用 key（enabled + healthy/unknown）聚合窗口用量，取最紧张窗口上报
2. **client**: `routing.ts` 新增 `tightestRateLimit()` 纯函数；`model-table.tsx` 组头新增三档徽章（≥100% 红 / ≥70% 琥珀 / 其余灰），15s 轮询刷新
3. **i18n**: 60 个 locale 补齐 `models.rateLimitUsage` / `models.rateLimitUsageTitle` 2 个 key

### 验证 / Verification

- `client` `tsc -b` ✅
- `check-i18n.mjs` 60 locales / 818 keys ✅
- 与 #871 的 monthly budget 徽章互补（budget 管月额、本 PR 管窗口限流）

### 备注 / Note

新增端点与既有 `GET /api/fallback/token-usage` 风格一致；无数据库迁移。


tashfeenahmed:
Thanks for taking #876. Needs rework before it can land:

1. Semantics: the badge shows the busiest key per model, but the router skips a key that fails `canMakeRequest` and tries the next one. A model with one exhausted key and four fresh ones shows red while being fully routable. Report the least busy usable key, or aggregate used vs limit x key count like `/token-usage` does. Also respect `model_scope_json` so keys scoped away from the model are not counted.
2. TPD is advertised in the PR body but `tpd_limit` is selected and never used, and `getRateLimitStatus` only returns rpm/rpd/tpm. Implement it or drop the claim.
3. Server cost: the endpoint runs 3 queries per enabled model x key every 15s per open dashboard. Replace with one grouped query over `rate_limit_usage` (the lookup index already covers it).
4. Client cost: `useQuery` is mounted inside `GroupHeaderCells`, so every group row gets its own observer and timer and rebuilds a Set per render. Hoist the query into `FallbackPage.tsx` next to the existing token-usage query and pass a `Map<modelDbId, row>` down.
5. i18n: `models.rateLimitUsageTitle` is English in all 60 locales. The format-only key is fine.
6. No tests. `tightestRateLimit` is pure and `client/src/lib/routing.test.ts` is the natural home; note `bestRatio` starts at 0 so a model at 0% renders nothing.

On the table itself: this is the first colored alert and live poll in the primary table, which we keep light. Please keep it muted in the table and put the colored detail on the model page, or only color at 70% and above.

Heads up that #871 touches the same badge region in `model-table.tsx` and `routing.ts`; whichever lands second will conflict.

PR #922, opened 2026-08-19
feat(providers): add Baidu Qianfan and Tencent Hunyuan (permanent free domestic providers)

suantea:
## Summary

Register two China-direct, **permanent free** OpenAI-compatible providers that are currently missing from FreeLLMAPI:

- **Baidu Qianfan** (`qianfan`) — `ERNIE-Speed` / `ERNIE-Lite` are permanently free with no token cap (documented QPS=50). Endpoint: `https://qianfan.baidubce.com/v2`. Key from the Qianfan console, no card required.
- **Tencent Hunyuan** (`hunyuan`) — `HunYuan-Lite` is permanently free with 256K context. Endpoint: `https://api.hunyuan.cloud.tencent.com/v1`. Key from the Tencent Cloud console, no card required.

Both tokens are long-lived (not one-time trial credits): the free tiers never expire and reset daily/monthly where applicable, which fits the project's "recurring no-card free tier" criterion.

## Changes

- `shared/types.ts`: add `qianfan` and `hunyuan` to the `Platform` union.
- `server/src/providers/index.ts`: register both as `OpenAICompatProvider`.
- `server/src/routes/keys.ts`: add both to the `PLATFORMS` allowlist.

Model rows are intentionally **not** seeded here — per the V25 convention, model/limit data lands via the hosted catalog (`catalog-sync` gates on `hasProvider`), so ERNIE/HunYuan model rows can be published upstream after this merges.

## Verification

- `tsc --noEmit` clean (only pre-existing `sharp` optional-dep errors on `image-normalize.ts`).
- `vitest run` — 98 tests pass, 13 files (incl. `keys-providers-checklist.test.ts`).

## Notes

Happy to add other permanent-free China providers (Volcengine Doubao daily-reset tier, LongCat daily-reset tier, Spark Lite, Moonshot) in follow-up PRs if maintainers want them.

PR #923, opened 2026-08-19
feat(providers): add Volcengine Ark and LongCat (daily-reset free domestic providers)

suantea:
## Summary

Follow-up to #922 — two more **China-direct, daily-reset** free OpenAI-compatible providers that are missing from FreeLLMAPI:

- **Volcengine Ark** (`volcengine`, 火山方舟/ByteDance) — Doubao models on a recurring free tier (~2M tokens/day, **daily reset**). Endpoint: `https://ark.cn-beijing.volces.com/api/v3`. Key from the Volcengine Ark console, no card required.
- **LongCat** (`longcat`, 面壁智能) — public-beta free tier: 500K tokens/day per account (Flash-Lite 50M/day), **reset daily at midnight Beijing time**, no accumulation. Endpoint: `https://api.longcat.chat/openai/v1` (also exposes an Anthropic-compatible `/anthropic` endpoint). Key from longcat.chat, no card.

Both tokens refresh daily instead of expiring as one-shot credits, fitting the project's recurring-free criterion.

## Changes

- `shared/types.ts`: add `volcengine` and `longcat` to the `Platform` union.
- `server/src/providers/index.ts`: register both as `OpenAICompatProvider`.
- `server/src/routes/keys.ts`: add both to the `PLATFORMS` allowlist.

Model rows are not seeded here (V25 convention) — they land via the hosted catalog once `hasProvider` is true.

## Verification

- `tsc --noEmit` clean (only pre-existing `sharp` optional-dep errors).
- `vitest run` — keys-providers-checklist 3/3 pass.

Stacked on #922 (`feat/domestic-free-providers`).


tashfeenahmed:
Superseded by #936, merged as df1000a.

**Why all three closed together:** #922, #923 and #924 are stacked strict
supersets of one another (922 ⊂ 923 ⊂ 924) — every provider in #922 is also in
#923, and every provider in #923 is also in #924. Merging more than one would
conflict on both `server/src/routes/keys.ts` and `shared/types.ts`, so only one
could ever land.

**What #936 changed relative to this PR:**

1. **Wired the platforms all the way through.** These PRs registered the
   provider and the `Platform` type, but not `server/src/lib/key-parser.ts` or
   the Keys page dropdown in `client/src/components/keys/shared.tsx`. That left
   every new platform unreachable in practice: no way to add a key from the
   dashboard, and no way to import one from a `.env` or `auth.json`. #936 adds
   those plus platform colors, and a test that asserts all four registration
   points together so a half-wired platform fails CI.

2. **Dropped Tencent Hunyuan.** `hunyuan-lite` was made free in May 2024, which
   is where this PR's description comes from, but Tencent's current pricing doc
   no longer lists that model and the free allowance is now a one-time 1M-token
   grant, valid one year, with no replenishment. Not a recurring free tier.

3. **Registered Baidu Qianfan, but with no catalog rows.** Same trap: the
   "permanently free ERNIE-Speed/Lite" claim traces to a May 2024 announcement
   that aggregators still repeat. Baidu's own retirement doc
   (模型版本升级及退役机制) retired ERNIE-Speed-8K, ERNIE-Speed-128K,
   ERNIE-Lite-8K and ERNIE-Tiny-8K on **2026-01-27**, and the named
   replacements (ERNIE-Speed-Pro-128K, ERNIE-Lite-Pro-128K) are paid.

4. **Corrected provider facts** against vendor docs:
   - LongCat is **Meituan (美团)**, not 面壁智能.
   - LongCat's free tier is **100K tokens/day** at launch; the 50M/day
     Flash-Lite figure was announced as a *future plan*, so it is not claimed.
   - iFlytek publishes no token cap and **no QPS figure** for Spark Lite, so
     the "QPS=2" claim is dropped.
   - Volcengine's **2M tokens/day per model** daily quota is real — and it is
     the strongest free tier of the group, so this PR's Volcengine description
     held up best.

5. **Added catalog rows separately**, in the signed hosted catalog
   (`2026.08.20.9eb863ca`) rather than in code — 12 Doubao models, LongCat-2.0,
   and Spark Lite. All ship **disabled** pending a live smoke test, since all
   four providers gate API access behind Chinese real-name verification and no
   key was obtainable to test with.

Thanks for these — the provider research behind them was the useful part, and
the endpoints and auth schemes you transcribed were all correct.

Issue #926, opened 2026-08-19
when it health check the key,it will add all the model to the list that i dont need.


Fone520:
when it health check the key,it will add all the model to the list that i dont need.

<img width="1268" height="3499" alt="Image" src="https://github.com/user-attachments/assets/d2b7dacc-4fcf-47a6-beec-259272379116" />
i had delete the key that i dont need,and they all come back .

suantea:
I'll take this one. Root cause found:

The health check itself doesn't add models — the **scheduled custom-model sync** does. It runs daily over every configured custom endpoint and registers any model id it discovers that isn't currently in the `models` table. It has no tombstone mechanism, so a model you deleted (or a key you deleted, which cascades its models away) is treated as "new" again on the next pass and gets re-registered — which is why they "all come back".

Catalog-managed models already have this solved (`catalog_model_tombstones`), the custom-endpoint path just never got the same treatment.

My fix mirrors that pattern:
- New `custom_model_tombstones` table keyed by `(endpoint_scope, model_id)` (per-endpoint, so deleting a model on one relay never hides the same id on another).
- The delete routes (`DELETE /custom/:id`, `DELETE /:id`, and deleting a custom key with its models) record a tombstone.
- `runCustomModelSync` skips tombstoned models instead of resurrecting them (new `tombstoned` counter in the result).
- An explicit re-registration (POST /custom, bulk import, "Fetch models") clears the tombstone — so you can still deliberately re-add a model.

Tests included (sync skip, per-endpoint scoping, explicit re-register clearing the tombstone, migration up/down). Opening a PR shortly.

Fone520:
> I'll take this one. Root cause found:
> 
> The health check itself doesn't add models — the **scheduled custom-model sync** does. It runs daily over every configured custom endpoint and registers any model id it discovers that isn't currently in the `models` table. It has no tombstone mechanism, so a model you deleted (or a key you deleted, which cascades its models away) is treated as "new" again on the next pass and gets re-registered — which is why they "all come back".
> 
> Catalog-managed models already have this solved (`catalog_model_tombstones`), the custom-endpoint path just never got the same treatment.
> 
> My fix mirrors that pattern:
> 
> * New `custom_model_tombstones` table keyed by `(endpoint_scope, model_id)` (per-endpoint, so deleting a model on one relay never hides the same id on another).
[...]

PR #927, opened 2026-08-19
fix(models): keep deleted custom models deleted across the scheduled sync (#926)

suantea:
@/tmp/pr926.md


tashfeenahmed:
Thanks for digging into #926 and finding the real root cause in the scheduled custom-model sync, the tombstone approach looks right. This PR is identical to #931 (the diffs match byte for byte), and #931 is the newer one with a full description, so I am closing this one and keeping the review on #931. Happy to reopen if you would rather continue here instead.

Issue #919, opened 2026-08-19
feat(keys): least-remaining 配额加权 key 选择策略（#915 拆分跟进）/ least-remaining quota-weighted key selection strategy (follow-up to #915)


suantea:
## 概述 / Overview

建议为服务端 key 选择逻辑增加一个**显式的 least-remaining 配额加权策略**：在配额数据可用时，按各 key 的剩余配额比例加权选择，而不是当前的 pass/fail 门控 + round-robin 兜底。

This proposes an **explicit least-remaining quota-weighted key selection strategy** for the server: when quota data is available, weight key selection by each key's remaining quota proportion, instead of the current pass/fail gating plus round-robin fallback.

## 背景 / Background

从 #915 的讨论中拆分而来（维护者确认核心池化已实现于 `selectKeyForModel`）。当前实现：遍历启用且健康的 key，跳过冷却中/超限/并发满/模型不匹配的 key，有可靠性数据时按评分排序，否则按 round-robin 游标兜底——门控是 pass/fail，**没有配额余量加权**。

Split from #915 (maintainer confirmed core pooling is implemented in `selectKeyForModel`). Current behavior: iterate enabled, healthy keys, skipping those on cooldown / over limits / at concurrency ceiling / scoped to other models; order by score when reliability data exists, round-robin cursor otherwise — gating is pass/fail, **no remaining-quota weighting**.

## 建议 / Proposal

- 新增策略开关（如 `quota-aware-key-selection: least-remaining | round-robin | off`），默认 off 保持现有行为；
- 启用时，若 key 带额度/余量元数据（如 `quotaStates`），按「剩余额度高优先」加权；无额度元数据时回退现有排序；
- 与冷却联动保持：单 key 触发 rate-limit 后自动切换下一 key，不降级整个 provider。

- Add a strategy toggle (e.g. `quota-aware-key-selection: least-remaining | round-robin | off`), default off to preserve current behavior;
- When enabled, weight by higher remaining quota when per-key quota metadata (e.g. `quotaStates`) exists; fall back to current ordering otherwise;
- Keep cooldown integration: a rate-limited key rotates to the next one rather than degrading the provider.

## 边界 / Boundary

配额数据来源与 per-key/per-account 语义（#727 跟进）需先厘清；部分平台按账号/项目计费，多 key 不叠加，加权仅在 per-key 语义下有意义。

[...]

PR #929, opened 2026-08-19
feat(keys): least-remaining quota-weighted key selection strategy (#919)

suantea:
## Summary

Implements #919: a new `quota-weighted` routing strategy that orders parallel keys by how close each is to exhausting its observed quota, so multi-key providers drain evenly instead of hammering the roomiest key until it 429s.

## Changes

- `server/src/services/scoring.ts`: extend `RoutingStrategy` with `'quota-weighted'` and exclude it from `BANDIT_PRESETS` (no static weight vector applies)
- `server/src/services/router.ts`: `selectKeyByQuotaWeight` walk — ranks keys by usage ratio (most-exhausted first; keys with no observation data sort last) using `getQuotaStateForKeys()`, then mirrors the standard gate checks (cooldown, rpm/rpd/tpm/tpd, decrypt error, provider daily/minute caps, concurrency)
- `server/src/__tests__/routes/routing-semantics.test.ts`: #919 test seeds two groq keys with different remaining quotas (900/1000 vs 100/1000) and asserts the most-exhausted one is picked

## Behavior

- When the operator picks `quota-weighted` (dashboard routing strategy), every key selection for a model diverts to the quota-ordered walk
- Only fresh, confident observations count (`confidence >= 0.7`, valid limit/remaining); unknown keys fall through to the standard walk order as last resort
- No change to the default strategies (`priority`/`balanced`/`smartest`/`fastest`/`reliable`/`custom`)

## Verification

- `tsc --noEmit` clean (only the pre-existing unrelated `sharp` module error in `src/lib/image-normalize.ts`)
- `vitest run src/__tests__/routes/routing-semantics.test.ts`: 5/5 pass including the new #919 case

### DIFF
--- a/server/src/__tests__/routes/routing-semantics.test.ts
+++ b/server/src/__tests__/routes/routing-semantics.test.ts
@@ -9,6 +9,7 @@ import {
   routeRequest,
   setRoutingStrategy,
 } from '../../services/router.js';
+import { recordQuotaObservation } from '../../services/provider-quota.js';
 import { mintDashboardToken, isGatedApiPath } from '../helpers/auth.js';
 
 let app: Express;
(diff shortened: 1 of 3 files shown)


tashfeenahmed:
Thanks a lot for working on #919, the quota-weighted key selection is a genuinely useful strategy. This PR and #930 turned out to be the same change (the diffs are byte for byte identical), so I am keeping #930 open as the single review thread and closing this one to avoid duplicate review. Nothing is lost here, and if you think this branch has anything #930 is missing, just say so or push it there and we will pick it up.

PR #930, opened 2026-08-19
feat(router): least-remaining quota weighted key selection strategy (#919)

suantea:
## Summary

Adds a new routing strategy **`quota-weighted`** (issue #919): instead of round-robin or score-based key selection, a model's keys are ordered by how close each is to exhausting its observed quota (most-used first). This lets parallel keys drain evenly rather than hammering the roomiest one until it 429s.

## Changes

- **`scoring.ts`**: extend `RoutingStrategy` with `'quota-weighted'` (excluded from `BANDIT_PRESETS` — it's computed dynamically, not a weight vector).
- **`router.ts`**: add `selectKeyByQuotaWeight()`, ranking keys by the worst `remaining/limit` ratio across fresh, confident (`confidence >= 0.7`) quota observations; keys with no observation data sort last. Reuses the existing gate walk (cooldown / concurrency / rpm / rpd / tpm / tpd) and the decrypt/lease flow. `selectKeyForModel` diverts to it when the strategy is active.
- **`routing-semantics.test.ts`**: cover that the most-exhausted key wins.

## Testing

- `server` `tsc --noEmit`: 0 errors
- `client` `tsc -b`: clean
- `routing-semantics.test.ts`: 5 tests passed (incl. new #919 case)

## Notes

- Strategy is opt-in via the existing `routing_strategy` setting (no behavior change unless selected).
- Uses the existing `provider_quota_state` observations — no schema change.


tashfeenahmed:
Thanks. #919 is a real ask, but this needs a different shape before it can merge:

1. Not reachable. `quota-weighted` is added to the type and `VALID_STRATEGIES`, but `routes/fallback.ts`, `declarative-config.ts`, `routes/mcp.ts` and the client dropdown all still reject it.
2. It hijacks model ranking. `weightsFor()` returns null for the new strategy, so `orderChain` falls back to priority mode and the bandit is silently disabled. The issue asked for a separate toggle for exactly this reason. Please make it its own setting (e.g. `key_selection_strategy: auto | least-remaining`) and leave `RoutingStrategy` alone.
3. The new walk drops two behaviours from `selectKeyForModel`: the `customEndpointKeyIds` filter for custom platforms (a custom model can now be sent to another endpoint's key and base URL) and the `roundRobinIndex` fallback (with no quota observations it picks the lowest id key forever). Add the ordering as one more branch inside `selectKeyForModel`, returning null when no key has confident data so the existing cursor and scoping keep working.
4. `getQuotaStateForKeys()` is a dashboard query with a write in front of it, now called per candidate model per request. Filter by platform and cache it like `keyStatsCache`.
5. Direction: the issue asks to prefer the key with the most remaining quota. This picks the most exhausted one and describes that as draining evenly, which it is not. Either follow the issue or argue the consolidation case explicitly, and skip weighting when `quotaPoolKey` is account scoped.
6. The test inserts `near-exhausted` first so it already sorts first by rowid; the assertion passes with the sort removed. Please also cover the no-observation fallback.

Nit: import from `./provider-quota.js`, not `../services/provider-quota.js`.

PR #932, opened 2026-08-19
feat(router): same-model cross-provider fallback

stephenzwj:
## What it does

When a client pins an explicit model and that same `model_id` is served by multiple enabled platforms, failed attempts now try sibling rows for the same model immediately after the pinned row, before falling through to unrelated models.

Previously, a failed explicit pin could fall through to whichever unrelated model scored next in the chain.

## Scope

- Request-scoped: only applies when `preferredModelDbId` is set.
- Auto and sticky routing behavior is unchanged.
- No-op when the pinned model has no same-model siblings.

## Tests

Adds four regression tests covering:

- pinned model remains first;
- same-model sibling is selected before an unrelated model;
- unrelated model is selected after all same-model rows are exhausted;
- no-sibling behavior remains unchanged.

Full server test suite: 214 files, 2,352 tests passed, 5 skipped.
Server build passes.

### DIFF
--- a/server/src/__tests__/services/router-sibling-fallback.test.ts
+++ b/server/src/__tests__/services/router-sibling-fallback.test.ts
@@ -1,0 +1,96 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import { initDb, getDb } from '../../db/index.js';
+import { routeRequest, setRoutingStrategy, type ChainRow } from '../../services/router.js';
+import { encrypt } from '../../lib/crypto.js';
+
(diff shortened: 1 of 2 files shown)


tashfeenahmed:
Thanks, you found a real gap. The OpenAI, Responses and inbound chat surfaces already do same model cross provider fallback through model groups, so the splice in routeRequest was redundant there and changed sticky auto routing. The one surface that did not have it was /v1/messages, and that is now fixed in #1000 with you credited. Added you to the contributors list.

PR #944, opened 2026-08-20
Playground conversations + post-add model picker

tashfeenahmed:
Two Playground/Keys features, staged for review — **please try them on the test instance before merging** (deployed there as a preview build).

## Persistent Playground conversations

The Playground lost everything on reload. Now conversations persist server-side (SQLite — survives reloads, browsers, and devices, fitting the single-user design):

- New `playground_conversations` table (`20260820_000001` migration, reversible) and `/api/conversations` CRUD mounted behind the dashboard session gate exactly like `/api/client-profiles`; the unified `/v1` key does **not** open it (tested).
- Collapsible sidebar: conversation list with auto-titles (first user message, ≤40 chars) and relative times, inline rename, confirm-to-delete, New conversation. Clear now starts a new conversation and keeps the old one in the sidebar.
- Saves happen at exchange end (stream finish / fusion finish / error), never per delta; per-message routing meta, reasoning traces, and error bubbles round-trip verbatim. Transcripts cap at 2 MB with a clear 413. Switching restores transcript, model selection, and system prompt; the active conversation is remembered across reloads.

## Model picker after adding a key

Adding a key to a big-catalog platform (≥6 models) now offers a checkbox dialog to limit which models the key serves:

- Saves as **one** `PATCH /api/keys/:id` using the existing `modelScope` field (#657) — zero new server surface, existing auth.
- All-selected saves nothing (scope stays `null` = "all", so future catalog additions keep flowing); Confirm is disabled when nothing is checked, because an empty scope would server-side mean the opposite of what it looks like. Cancel/Esc changes nothing.
- Small catalogs, keyless gateways, `custom` endpoints, and bulk paste-imports skip the dialog — identical to today's behavior. The existing per-row scope editor (`ListFilter` icon) is untouched.

## i18n

24 new keys translated into all 60 locales (check:i18n green at 856 keys), matching each file's register and script; placeholders preserved.

## Testing

- Full suite green (server, CLI, client, i18n).
- Server route tests: 16 covering CRUD, meta round-trip, size caps, auth, and unified-key rejection. Client: +34 unit tests (scope-diff logic, conversation helpers).
- Headless-Chromium e2e against a live gateway + mock streaming provider: 9/9 conversation-lifecycle assertions (persistence across reload, switching, remembered active conversation) and the existing 6/6 streaming/scroll-follow regression suite still passes.


tashfeenahmed:
Updated: the Playground is now a full-bleed three-column layout — conversations left, chat center, and a collapsible right rail with the model picker, system prompt, and new sampling controls (temperature / top_p / max_tokens with per-knob on-off; only enabled knobs are sent). Below lg both rails collapse to strips. Verified in headless Chromium at 1600px and 820px; both e2e suites (streaming 6/6, conversations 9/9) pass on the new layout; full test suite green (864 i18n keys / 60 locales). Preview redeployed to https://freellm-test.178-105-187-189.sslip.io/playground

Issue #426, opened 2026-06-30
[Bug]GitHub Models API error 413: Request body too large for gpt-4.1 model. Max size: 8000 tokens.


NirvanaCh7:
https://github.com/tashfeenahmed/freellmapi/blob/db7b64ddc88bd7b69841d8511cd6276d6d358a1d/server/src/db/migrations/20260101_000000_legacy_baseline.ts#L644

it seems 128k now

Issue #470, opened 2026-07-06
All models exhausted. Add more API keys or wait for rate limits to reset. (HTTP Status: 429)


xsfhacg:
I have added the GPT-5.5 model using multiple different intermediary servers, and it functions normally when tested on the trial platform and Cherry Studio. However, whenever I access an AI editor application (such as Trae), an error is reported: "All models exhausted. Add more API keys or wait for rate limits to reset." (HTTP Status: 429)

tashfeenahmed:
This had two independent causes, both now fixed on main:

1. #479: routing counted your client's full max_tokens reservation as already-consumed tokens, so an agentic editor requesting a large completion budget exceeded the TPM ceiling of every free model and the router returned the 429 without making a single upstream call. Routing now reserves at most 2000 tokens for output when picking a route.

2. #478: custom OpenAI-compatible models were registered with tool support off, and any request carrying tools (which agentic editors like Trae always send) filtered those models out entirely. Custom models now default to tool-capable, existing ones were backfilled by a migration, and there's a per-model toggle if your endpoint genuinely can't do tools.

Pull latest main (or the next release) and this scenario should route normally. Thanks for the clear report with the client matrix, it's what made the second cause findable.

PR #956, opened 2026-08-21
fix(router): context-window safety margin against chars/4 under-counting

Icesenator:
## Description

The routing token estimate is a chars/4 heuristic (`proxy.ts`), which under-counts dense payloads (JSON, code, CJK) by up to ~2x. Oversized requests therefore passed the context-window routing filters, got dispatched to a model whose real tokenizer count exceeded the window, and were rejected with a 400 mid-chain — burning a fallback attempt each time.

Observed in production against the Kilo gateway: requests with **277k–523k real tokens** routed to models declared at 256k–262k because their local estimate was ≤256000 (`Kilo Gateway API error 400: This model's maximum context length is 262144 tokens. However, you requested 424079 tokens...`).

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes

- New `fitsContextWindow(platform, contextWindow, estimatedTokens)` helper in `router.ts`, applying a `CONTEXT_WINDOW_SAFETY_FACTOR = 1.25` margin: a request only fits when `estimate * 1.25 <= context_window`.
- All four context-filter sites now go through it (`routePinnedModel`, chain build, auto-router filter + diagnostics line).
- `TRIM_GUARDED_PLATFORMS = {github}`: platforms whose pre-dispatch trim guard already caps dispatched input below the live ceiling (`truncateMessagesForGithub`, 7500/8000) keep the strict comparison — applying the factor there would only make that guard unreachable.
- Tests: new case covering the margin on a non-guarded platform (groq 131072 → effective ceiling 104857); existing #426 github test unchanged and passing.

## Test Coverage

- [x] I added/updated unit tests that prove the fix works
- [x] New and existing tests pass locally (`npm run test -w server`: 2496 passed; one pre-existing environment failure in `csp-inline-bootstrap.test.ts` unrelated to this change — hash mismatch due to CRLF checkout)

### DIFF
--- a/server/src/__tests__/services/router.test.ts
+++ b/server/src/__tests__/services/router.test.ts
@@ -172,6 +172,31 @@ describe('Router', () => {
     expect(routeRequest(9000).modelDbId).toBe(groq.id);
   });
 
+  it('applies the context-window safety margin on platforms without a trim guard', () => {
+    const db = getDb();
+    const groqKey = encrypt('test-groq-key');
+    db.prepare(`
+      INSERT INTO api_keys (platform, label, encrypted_key, iv, auth_tag, status, enabled)
+      VALUES (?, ?, ?, ?, ?, ?, ?)
+    `).run('groq', 'groq', groqKey.encrypted, groqKey.iv, groqKey.authTag, 'healthy', 1);
+
+    const groq = db.prepare(`
+      SELECT id FROM models
+       WHERE platform = 'groq' AND context_window = 131072
+       LIMIT 1
+    `).get() as { id: number };
+
+    db.prepare('UPDATE fallback_config SET priority = 1000, enabled = 1').run();
+    db.prepare('UPDATE fallback_config SET priority = 1 WHERE model_db_id = ?').run(groq.id);
+    db.prepare('UPDATE models SET tpm_limit = NULL, tpd_limit = NULL WHERE id = ?').run(groq.id);
+
+    // CONTEXT_WINDOW_SAFETY_FACTOR: the chars/4 estimate under-counts dense
+    // payloads (JSON, code), so the effective ceiling is window / 1.25.
+    // 131072 / 1.25 = 104857.6 — just under still routes, just over does not.
+    expect(routeRequest(104000).modelDbId).toBe(groq.id);
+    expect(() => routeRequest(106000)).toThrow();
+  });
+
   it('still routes a model with an unknown (null) context window (#167)', () => {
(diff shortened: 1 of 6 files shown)


tashfeenahmed:
Thanks, the idea is sound and the github carve-out is right. Two things before merge:

1. The margin is a hard filter, but `/v1/models` still advertises the raw window. A client that packs to the advertised 131072 now gets "all models exhausted" instead of one attempt, and today that attempt is already handled (the 400 is classified `context_too_large` and the loop moves on). Please make it a soft preference: filter with the factor first, and if nothing survives, re-run that pass with the strict comparison.
2. The diag string `context 131072 < estimated 106000 (x1.25 margin)` reads as false. Emit the effective number, e.g. `context 131072 < estimated 106000 x1.25 = 132500`, keeping the `< estimated` substring so exhaustion bucketing still works.

Nits: the reserve from `max_tokens` is exact, so it should not be multiplied; the #470 comment above the fourth filter still describes strict semantics; the new test only passes because no seeded groq model has a window over 132500, so assert "not groq.id" instead of `toThrow`.

Icesenator:
Thanks for the careful read — all addressed in e500abb:

1. **Soft preference**: the margin no longer excludes anyone. `routeRequest` sweeps models that fit *with* the factor first; raw advertised-window fits stay eligible behind them (demoted, not excluded). `getOrderedFusionChain` re-runs its servable filter with the strict comparison when nothing survives the margin pass, and `routePinnedModel` compares strictly — a pinned slot has no substitute, so attempting beats refusing. A client packed to the advertised 131072 gets its one attempt back.

2. **Diag line**: now emits the effective number while keeping `< estimated` for bucketing — e.g. `groq/…: context 131072 < estimated 106000 x1.25 = 132500`. Trim-guarded platforms keep the plain form since the factor never applies there.

3. Nits:
   - the exact `routingReserveTokens(max_tokens)` reserve is threaded through `routeRequest`/`fitsContextWindow` (`exactOutputReserve`) and added unscaled — only the chars/4 heuristic portion is multiplied;
   - the #470 comment above the loop guard now describes the strict-guard + deferred-margin semantics;
   - the test asserts which model actually serves each estimate instead of `toThrow`: priority-1 groq 131072 wins at 104000, loses to the margin-fitting google runner-up at 106000, and wins again once no margin-fitting model has a usable key — covering both the preference and the "one attempt" fallback.

`npm run test -w server`: 2497 passed / 0 failed.

PR #957, opened 2026-08-21
fix(ui): prevent monthly token budget figures from wrapping off-screen (#887)

### LINKED ISSUE #887: Used monthly token budget figures are obscured
The used‑amount figure does not show which endpoints and models are included. Usage from custom endpoints appears to be counted towards this total. There is a discrepancy between this metric and the statistics shown on the analytics panel.

One potential contributing factor could be that I also access these models (like opencode.ai) directly through an agent, which may explain why this usage figure is higher than the statistics shown on the analytics panel.

Image

Image

Image

It would be preferable to display remaining/total metrics for each model. In addition, allow zero‑value metrics to be hidden. I have observed that Unorouter returns nearly one‑hundred models with zero quota.

Image

suantea:
## 中文

修复 #887 monthly token budget 数字被遮挡的问题。

### 问题

token-usage-bar.tsx 顶部统计行（remaining · % of total · used）使用 flex 不加 wrap，在窄屏时 used-amount 数字被推到屏幕外。

### 修复

添加 flex-wrap + gap-x-3 gap-y-1，让统计行在空间不足时自动换行到第二行，不再被截断。

### 验证

- client tsc -b ✅

## English

Fixes #887 monthly token budget figures obscured.

### Problem

The summary line (remaining · % of total · used) in token-usage-bar.tsx used flex without wrap, pushing the used-amount figure off-screen on narrow viewports.

### Fix

Add flex-wrap + gap-x-3 gap-y-1 so the stats wrap to a second line when space is tight.

### Verification

- client tsc -b ✅

### DIFF
--- a/client/src/components/token-usage-bar.tsx
+++ b/client/src/components/token-usage-bar.tsx
@@ -44,7 +44,7 @@ export function TokenUsageBar({ data }: { data: TokenUsageData }) {
 
   return (
     <section className="rounded-3xl border bg-card p-5">
-      <div className="flex items-baseline justify-between mb-3">
+      <div className="flex flex-wrap items-baseline justify-between gap-x-3 gap-y-1 mb-3">
         <h2 className="text-sm font-medium">{t('models.monthlyTokenBudget')}</h2>
         <span className="text-xs text-muted-foreground tabular-nums">
           <span className="text-foreground font-medium">{formatTokens(remaining)}</span> {t('models.remaining')}


tashfeenahmed:
Thanks, but this is a misread of #887. The summary span is a shrinkable flex item and already wraps, so there is no clipping bug. What #887 actually asks for is provenance on the used figure (chat only, calendar month, custom endpoints included), remaining out of total per model, and hiding zero quota rows. That work is happening on #887 directly.

PR #968, opened 2026-08-22
feat(providers): add UnoRouter (unorouter.com) as an OpenAI-compatible provider (#875)

### LINKED ISSUE #875: New provider - UnoRouter
https://unorouter.com/

FYI:
https://github.com/amardeeplakshkar/awesome-free-llm-apis#unorouter-

* Custom (OpenAI-compatible) API error 400: This model's maximum context length is 262144 tokens. However, you requested 32000 output tokens and your prompt contains at least 230145 input tokens, for a total of at least 262145 tokens.
* Custom (OpenAI-compatible) API error 403: This model is busy right now (free providers hit their rate limit). Please try again in a little while, or switch to another model. (request id: [redacted-token])
* Custom (OpenAI-compatible) API error 429: Too many requests. The free tier allows 1 request(s) every 1 min per account on intern-latest:free - nothing is used up, retry in 52s. The paid intern-latest has no per-minute limit. (request id:...

tashdroid:
## Summary
Closes #875.

unorouter.com is a real OpenAI-compatible gateway. The web app is a Next.js site at unorouter.com (which redirects /v1/* to the marketing app, **not** the API); the real API is **api.unorouter.com/v1**.

**Live-probed 2026-08-22:**
- GET /v1/models → 401 "Invalid token" without a key
- POST /v1/chat/completions → 401 "Invalid token" without a key

So the default /v1/models key validation is a real check — no validateUrl override needed (unlike xkiro, whose /v1/models is public).

## Changes (mirrors the OrcaRouter pattern)
- shared/types.ts — 'unorouter' in the Platform union
- server/src/providers/index.ts — OpenAICompatProvider (api.unorouter.com/v1)
- server/src/lib/key-parser.ts — UNOROUTER_/UNO_ROUTER_ prefixes + unorouter / uno-router auth.json aliases
- server/src/routes/keys.ts — platform in the import whitelist
- server/src/services/provider-quota.ts — one per-account free pool (unorouter::free) + isSharedPool entry
- client — PLATFORMS entry + platformColors
- tests — platform-instance entry + prefix/alias assertions

**Note on prefixes:** bare uno / UNO_ deliberately omitted — "uno" is a generic word (unlike "orca"), so it is a false-positive risk for key detection.

## On the model ids in #875
The specific model ids are candidates for **catalog authoring** (hosted catalog, premium now, free after the 30-day model-age gate), where a bad id is caught by the health check instead of shipped as a default. This PR wires up the gateway-side provider registration so the platform is usable; the catalog rows are a separate catalog-repo change.

## Checks
- server: vitest 2564 passed (1 pre-existing flake: compression.test.ts timing, fails on clean main too)
- client: vitest 177 passed, tsc clean (only pre-existing TS5101 baseUrl deprecation)
- i18n parity: n/a (no locale keys touched)


tashfeenahmed:
Merged as #987 with two comment corrections from a live probe (public /v1/models still rejects a wrong key with 401; a burst trips an account-wide 429, so the single quota pool stays). Your commit is on main with your authorship. Thanks.

Issue #969, opened 2026-08-22
The bar chart in the "analytics" interface is displayed incorrectly


growster:
The bar chart in the "analytics" interface is displayed incorrectly; while hovering over a rectangle shows the name, it does not appear on the horizontal axis below.

<img width="848" height="515" alt="Image" src="https://github.com/user-attachments/assets/c11eff43-beff-40ac-8663-c5f49f685d8b" />

PR #977, opened 2026-08-22
fix(analytics): add bottom margin so bar/line chart x-axis labels render (#969)

suantea:
## Summary\nBarChart and LineChart in AnalyticsPage used margin bottom: 0, leaving no space for x-axis tick labels. Hover tooltips showed names but the horizontal axis was invisible.\n\nBump to bottom: 12.\n\nCloses #969.


tashfeenahmed:
Thanks for the PR. This one is already fixed on main: #959 landed a few hours before #969 was filed and switches the category axes to interval 0 via chart-axis.ts, so every label renders now. The reporter was on an older build. Closing this one, but appreciate you jumping on it.

PR #980, opened 2026-08-22
fix(groq): clamp max_tokens to the free-tier 8192 ceiling

Icesenator:
## Problem

Groq's API rejects any request whose \max_tokens\ exceeds 8192 with a hard 400:

\\\
Groq API error 400: \max_tokens\ must be less than or equal to \8192\, the maximum value for \max_tokens\ is...
\\\

When a client asks for a larger budget (an agentic client sending \max_tokens: 16000\ is common these days), that 400 is classified as an upstream failure and gets **spent as a wasted fallback hop** before the fallback loop moves on to the next candidate. Observed live on 2026-08-22 against a production gateway running this repo.

## Fix

Clamp it in the platform policy instead, following the same pattern as \github\ / \GITHUB_MAX_OUTPUT_TOKENS\:

\\\diff
-  groq: { drop: ['logprobs', 'top_logprobs', 'logit_bias'] },
+  groq: { drop: ['logprobs', 'top_logprobs', 'logit_bias'], maxTokensCap: 8192 },
\\\

The cap is applied at parameter-resolution time (\min(requested, maxTokensCap)\), so Groq candidates only ever receive values they accept — same rationale as the GitHub Models cap already documented in this file.

## Notes

- Only affects requests asking for more than 8192; smaller budgets pass through untouched.
- No new config surface, no behavior change for other platforms.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,11 @@
 services:
   freellmapi:
-    image: ghcr.io/tashfeenahmed/freellmapi:latest
+    # Local tag, NOT ghcr.io/tashfeenahmed/freellmapi:latest: this deployment
+    # carries local patches (context-window safety margin, 2026-08-21) and a
+    # `docker compose pull` on the upstream tag would silently overwrite them.
+    # To pick up upstream changes: git pull, then `docker compose build && docker compose up -d`.
+    image: freellmapi-local:patched
+    pull_policy: never
     build:
       context: .
       dockerfile: Dockerfile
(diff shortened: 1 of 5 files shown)


tashfeenahmed:
Thanks. The context-window safety margin from this branch landed independently on main (CONTEXT_WINDOW_SAFETY_FACTOR and fitsContextWindow in router.ts, now with an output reserve), which is where the conflicts come from. The Groq part I have to decline: Groq's output ceiling is per model (llama-3.1-8b is 131k, gpt-oss is 65k, only Compound is 8192), so a platform-wide 8192 cap would truncate most Groq models, and unified-max-tokens.test.ts pins that deliberately. If you hit a real 400 today, unified_max_tokens is the operator-level knob. The diff also carries your local docker-compose.yml and a package-lock bump. Closing.

PR #983, opened 2026-08-23
fix(routing): skip models known to ignore response_format for structured-output requests (#933)

### LINKED ISSUE #933: Structured JSON response issues
When using the `auto` endpoint, structured JSON responses often do not work. Models respond like normal unless they support it, and FreeLLMAPI doesn't seem to detect and route those correctly

suantea:
## Summary

Closes #933 — structured-output requests (`response_format` json_object/json_schema) on the `auto` endpoint often landed on models that accept the param but answer in prose, burning the first failover hop on a model known to ignore it.

## Changes

- **`server/src/lib/sampling-params.ts`**: new `modelSupportsStructuredOutput(platform, modelId)` — forward-by-default family rule that excludes models with observed evidence of ignoring `response_format` (hermes, gemma, nemotron-nano, r1-distill, pollinations…; same evidence base as the V22 `supports_tools` rules). Unknown/new models default to capable so they never get starved of structured traffic.
- **`server/src/services/router.ts`**: the `requireStructured` gate (both the exploration path and the main loop) now skips a candidate when the **platform** drops `response_format` OR the **model** is known to ignore it, instead of the platform-level check alone.
- **Tests**: `sampling-params.test.ts` covers forward-by-default, the excluded families, and case-insensitivity.

## Notes

- Platform-level `platformDropsResponseFormat` remains the fast path (a param dropped before send); the model-level rule targets families that forward the param but answer in prose.
- The downstream non-stream JSON enforcement still backstops anything that slips through.
- Verified: `tsc --noEmit` passes; sampling-params + router tests pass (55 tests).

### DIFF
--- a/server/src/__tests__/lib/sampling-params.test.ts
+++ b/server/src/__tests__/lib/sampling-params.test.ts
@@ -5,6 +5,7 @@ import {
   supportedParametersFor,
   supportedParametersForPlatforms,
   normalizeReasoningEffort,
+  modelSupportsStructuredOutput,
   EXTENDED_SAMPLING_KEYS,
   REASONING_EFFORTS,
 } from '../../lib/sampling-params.js';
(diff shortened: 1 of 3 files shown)


suantea:
重复 PR，关闭让位给 #985（同主题 #933，含流式 JSON 强制检查 + 路由模型级门控，覆盖本 PR 全部改动）。修复已完整落在 #985。

PR #997, opened 2026-08-23
feat: implement login system with Flask and SQLite

kktfextrader-hue:
## Summary
Adds a login system (Flask + SQLite) following the project's anti-pattern guidelines.

## Notes
- Follows existing conventions in the codebase.

Opened via fork because origin upstream is read-only for me.

### DIFF
--- a/test_app.py
+++ b/test_app.py
@@ -1,0 +1,308 @@
+"""
+test_app.py
+
+A minimal yet robust login system built with Flask, adhering strictly to the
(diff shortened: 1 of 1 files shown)


tashfeenahmed:
This adds a standalone Flask app (test_app.py) to a Node/TypeScript repo that has no Python and already has auth in server/. It looks like it was committed from another workspace by mistake. Closing. Happy to look at anything targeting the actual gateway.

PR #1007, opened 2026-08-23
fix(google): diff cumulative streamGenerateContent chunks to true incremental deltas

L2-ootm:
## Problem

Gemini's `streamGenerateContent` does not guarantee incremental chunks. In several modes — notably **thinking models** and some structured-output responses — each SSE frame carries the **full accumulated candidate text** rather than only the new fragment.

The OpenAI-compatible chunk contract that this proxy exposes expects `delta.content` to be a strictly incremental suffix. Forwarding Gemini's cumulative chunks as-is caused downstream consumers (chat renderers, delta-buffering clients) to display duplicated/overlapping output.

## Fix

Track the text already emitted and diff each chunk's visible text down to the new suffix (`diffCumulativeText`) before yielding:

- thought/reasoning parts bypass the text diff and keep flowing as `reasoning_content`
- tool calls unchanged (existing `seenToolCallKeys` dedupe still applies)
- malformed SSE frames still skipped defensively

## Tests

- Unit tests for the `diffCumulativeText` helper (prefix-extension, exact-repeat, subset, disjoint cases)
- Streaming regression test: cumulative chunks render as strictly incremental deltas, joined text equals full content
- Streaming regression test: reasoning_content and tool calls pass through unaffected while visible text is diffed

All 34 tests in `google.test.ts` pass.

### DIFF
--- a/server/src/__tests__/providers/google.test.ts
+++ b/server/src/__tests__/providers/google.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
-import { GoogleProvider } from '../../providers/google.js';
+import { GoogleProvider, diffCumulativeText } from '../../providers/google.js';
 
 describe('GoogleProvider', () => {
   let provider: GoogleProvider;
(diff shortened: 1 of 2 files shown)


tashfeenahmed:
Thanks for looking at this. We call streamGenerateContent with alt=sse (google.ts), where Gemini's chunks are already incremental, and the repo fixtures encode that contract. The prefix heuristic also silently drops any delta that happens to be a prefix of the text so far, so a stream like [' ', 'Hello', ' ', 'world'] renders as ' Helloworld'. Closing. If you have a captured Gemini SSE trace showing cumulative frames, please attach it and we will reopen.

PR #1028, opened 2026-08-25
包含 commit 4f9da04

coffcoe:


tashfeenahmed:
This is a snapshot of your fork (89 commits, 95 files, 257 behind main) rather than a reviewable change. It includes the CNB pipeline, local start scripts, a Tailscale debugging report and a GHCR workflow pointing at your own registry, so it cannot be merged. The iFlytek Spark provider is already on main as xfyun. If you would like to contribute, the validate-providers / free-model-audit script is the interesting piece. Please open a separate rebased PR with just that, English comments and no local config. Closing.

PR #1033, opened 2026-08-25
Routing Profiles: model groups, priorities, and automatic fallbacks

### LINKED ISSUE #1026: Feature: Model groups, priorities, and automatic fallbacks
Currently users select a specific model directly.

It would be useful to define logical model groups/profiles such as:

- Coding
- Reasoning
- Fast
- Vision
- Long Context

Each group could contain multiple models with priorities.

Example:

Coding
  1. Model A
  2. Model B
  3. Model C

When Model A fails because of:
- 429 / rate limit
- timeout
- 5xx/provider unavailable

the router automatically tries Model B, then Model C.

Ideally the routing layer could also consider:
- model capabilities
- provider availability
- latency
- rate limits
- user-defined priority

This would allow applications to request a capability/profile rather than hard-code a provider-specific model ID.

stellachenyl:
# Routing Profiles: model groups, priorities, and automatic fallbacks

**Closes #1026**

## Summary

Adds **routing profiles** — named capability groups (`coding`, `fast`, `vision`, …) that clients can send *as* the `model` value. Each profile expands into a strict, priority-ordered failover chain across **different logical models**: when the head model 429s, times out, or returns a 5xx, the router automatically tries the next member, then the next — reusing the gateway's existing cooldown/skip/exhaustion machinery unchanged.

Applications can now request a *capability* instead of hard-coding a provider-specific model id:

```bash
curl http://localhost:31415/v1/chat/completions \
  -H "Authorization: Bearer $FREELLMAPI_KEY" \
  -d '{ "model": "coding", "messages": [...] }'
```

## How it works

A profile is an ordered list of member refs with priorities:

```json
{
  "slug": "coding",
  "name": "Coding",
  "description": "code well",
  "models": [
    { "ref": "groq:qwen3-coder-fast",   "priority": 1 },
    { "ref": "deepseek-r1-coder",        "priority": 2 }
  ]
}
```

Each ref accepts any form the existing resolution ladder answers to — a bare `model_id`, a canonical unify-group slug, or a fully-qualified `platform:model_id` — so a single entry can itself expand into several provider copies that fail over within its priority slot.

Key behaviors:

- **Strict ordering.** A profile is an explicit arrangement ("these models, in this order"). Profile chains are flagged so `routeRequest` orders them by profile priority only — the active strategy (bandit/balanced/fastest) cannot reshuffle them, and exploration probes are skipped. Failure penalties still demote a flapping member.
- **Precedence ladder.** Resolution tries real models / unify groups first, and only falls through to a profile slug. A profile can therefore never hijack a model id an existing client already uses.
- **Honest degradation & exhaustion.** Unresolvable member refs are skipped (and surfaced via the admin API); a request whose profile matches nothing gets the usual `404 model_not_found`; all members down yields the honest exhaustion status — never an off-profile answer.
- **No schema change.** Profiles are stored as JSON in the existing `settings` table, mirroring the unify-overrides pattern.

### Server

| File | Change |
|---|---|
| `services/routing-profiles.ts` | **New.** Zod schema + settings storage, pure expansion core (`expandProfile`, `resolveRoutingProfile`, `resolveDispatchTarget`, `dispatchChainOptions`) |
| `services/model-groups.ts` | `DispatchMembers.priorities` + `GroupCandidatesOptions` |
| `services/router.ts` | `resolveModelGroupCandidates(...)` accepts priority overrides / strict-priority ordering; profile chains flagged so `routeRequest` keeps the arrangement |
| `routes/proxy.ts`, `routes/anthropic.ts`, `routes/responses.ts`, `lib/inbound-chat.ts` | All four inbound surfaces resolve profiles after the model/group ladder |
| `services/model-listing.ts` | `/v1/models` advertises each profile as a first-class entry (`owned_by: freellmapi-profile`) so discovery works for agents |
| `routes/routing-profiles.ts` + `app.ts` | **New.** Admin CRUD at `/api/routing-profiles` (dashboard-session gated) with slug validation, duplicate detection, and per-profile diagnostics (`unresolvedRefs`, `memberCount`) |

### Client

| File | Change |
|---|---|
| `pages/RoutingProfilesPage.tsx` | **New.** Profiles management page: create/delete profiles, add members with a model autocomplete datalist, reorder/remove members, unresolved-ref warnings |
| `components/models-tabs.tsx`, `App.tsx` | New "Profiles" tab under Models (`/models/profiles`) |

## Admin API

```
GET    /api/routing-profiles          # list, with unresolvedRefs/memberCount
POST   /api/routing-profiles          # create { slug, name, description?, models[] }
PUT    /api/routing-profiles/:slug    # update name/description/models
DELETE /api/routing-profiles/:slug
```

Slug rules: lowercase `[a-z0-9-_]`; anything starting with `auto` is rejected (every surface treats `auto`/`auto:*` as automatic routing before resolution runs).

## Testing

Two new suites (20 tests), plus the full existing suite green:

- **Unit** (`__tests__/services/routing-profiles.test.ts`): slug rules, priority ordering, ref-ladder expansion + dedup, unresolvable-ref skipping, case-insensitive matching, chain-option derivation.
- **End-to-end** (`__tests__/routes/routing-profiles.test.ts`): admin CRUD validation, `/v1/models` advertisement, cross-model failover (`groq 429 → cerebras`), profile priority beating the active `smartest` strategy, no-hijack of same-named models, `model_not_found` fall-through, and honest 429 exhaustion.

Full run: server 237 files / 2748 tests ✅ · client 245 tests ✅ · `check-i18n` ✅ · tsc + eslint clean on both workspaces.

## Notes for reviewers

- **i18n:** the new page intentionally uses hardcoded English strings (no new keys), so `check-i18n` needs zero locale changes. Strings are centralized in one `STR` object in the page with a comment describing how to migrate them into `locales/*.json` when translations are contributed.
- The dashboard already has per-provider fallback chains ("profiles" in the UI sense). This feature is complementary and deliberately named *routing profiles*: those chains reorder global auto-routing; these are request-addressable groups spanning different models.
- Happy to add docs-page coverage and/or default seed profiles (e.g. `fast`, `coding`) if maintainers want them shipped enabled-by-default.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -42,6 +42,7 @@ import LogsPage from '@/pages/LogsPage'
 import PremiumPage from '@/pages/PremiumPage'
 import NotFoundPage from '@/pages/NotFoundPage'
 import AgentsPage from '@/pages/AgentsPage'
+import RoutingProfilesPage from '@/pages/RoutingProfilesPage'
 
 // Every failed mutation surfaces as an error toast, so no action fails
 // silently. A page that already shows the failure inline can opt out with
@@ -420,6 +421,7 @@ function App() {
                       <Route path="/models/chat" element={<FallbackPage />} />
                       <Route path="/models/chat/:id" element={<ModelDetailPage />} />
                       <Route path="/models/fusion" element={<FusionPage />} />
+                      <Route path="/models/profiles" element={<RoutingProfilesPage />} />
                       <Route path="/models/embeddings" element={<EmbeddingsPage />} />
                       <Route path="/models/embeddings/:id" element={<EmbeddingDetailPage />} />
                       <Route path="/models/image" element={<ImagePage />} />
--- a/client/src/components/models-tabs.tsx
+++ b/client/src/components/models-tabs.tsx
@@ -20,6 +20,8 @@ export function ModelsTabs() {
       <NavLink to="/models/video" className={({ isActive }) => tab(isActive)}>{t('models.videoTab')}</NavLink>
       <NavLink to="/models/audio" className={({ isActive }) => tab(isActive)}>{t('models.audioTab')}</NavLink>
       <NavLink to="/models/fusion" className={({ isActive }) => tab(isActive)}>{t('models.fusionTab')}</NavLink>
+      {/* English-only until routing-profiles strings land in the locale files. */}
+      <NavLink to="/models/profiles" className={({ isActive }) => tab(isActive)}>Profiles</NavLink>
     </div>
   )
 }
--- a/client/src/pages/RoutingProfilesPage.tsx
+++ b/client/src/pages/RoutingProfilesPage.tsx
@@ -1,0 +1,302 @@
+import { useState } from 'react'
+import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query'
+import { ArrowDown, ArrowUp, Layers3, Plus, TriangleAlert } from 'lucide-react'
+import { apiFetch } from '@/lib/api'
+import { Button } from '@/components/ui/button'
+import { Input } from '@/components/ui/input'
+import { ConfirmButton } from '@/components/confirm-button'
+import { EmptyState } from '@/components/empty-state'
+import { PageHeader } from '@/components/page-header'
+import { ModelsTabs } from '@/components/models-tabs'
+
+// Routing profiles (#1026): named capability groups ("coding", "fast", …)
+// clients send AS the model id. The server expands each profile into a strict,
+// priority-ordered failover chain across different logical models; this page is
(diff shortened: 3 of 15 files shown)


tashfeenahmed:
Thanks for the effort here, the code is clean and the test suite passes. I am closing it though, because it builds a second profiles store next to the named chains that already exist on the Fallback page and are exposed as auto:<name>. Two parallel ways to define the same thing would confuse operators and double the maintenance.

The one real gap your PR touches is that a chain's manual order is only honoured under the priority strategy. If you want to send a small follow-up that adds a keep_manual_order flag on existing chains, reusing your router.ts strict-order hunk, I would take that gladly. It would need the strings routed through i18n rather than hardcoded English.

Issue #255, opened 2026-06-07
Residual avoidable errors: TPM pre-routing (PR #254) + two provider request-shaping 400s


RobinHoodO:
Running a single proxy across the free tiers, three small classes of error look avoidable upstream. Filing together; the first has a PR attached.

### 1. Pre-route on TPM, not just context window  (PR attached)
A model with large context but small TPM (Groq `gpt-oss-120b`: 131k ctx / 8k TPM) clears the existing `context_window` guard and then 413s on a big prompt:
```
413: Request too large ... tokens per minute (TPM): Limit 8000, Requested 20449
```
Fixed by mirroring the context guard for `tpm_limit`.

### 2. NVIDIA NIM single-tool-call 400
```
400: This model only supports single tool-calls at once!
```
The NIM adapter could cap/serialise parallel tool calls for these models instead of passing them through.

### 3. OpenCode Zen (DeepSeek) reasoning_content 400
```
400: The reasoning_content in the thinking mode must be passed back to the API
```
The adapter should round-trip `reasoning_content` on follow-up turns for DeepSeek thinking models.

(2) and (3) are low-volume but each a clean per-adapter fix — happy to follow up with PRs if the approach sounds right.

tashfeenahmed:
All three handled in #263, thanks @RobinHoodO:

1. **TPM pre-routing** — your #254 cherry-picked in; the router now skips a model whose `tpm_limit` can't fit the request, same shape as the context-window guard.
2. **NVIDIA NIM single-tool-call 400** — new `forceSingleToolCall` flag on the openai-compat adapter pins `parallel_tool_calls:false` when tools are present, enabled for NVIDIA. No more `This model only supports single tool-calls at once!`.
3. **OpenCode Zen DeepSeek reasoning_content 400** — the proxy was stripping `reasoning_content` twice (assistant zod schema + message rebuild). It's now preserved through validation and round-tripped on follow-up turns, same as `thought_signature`.

You're in the README contributors. Closing once #263 lands.

PR #1040, opened 2026-08-26
fix(proxy): preserve Moonshot assistant `partial` prefill flag (#1038)

### LINKED ISSUE #1038: Assistant `partial` key is stripped by the chat schema, so Moonshot prefill never reaches the provider
I run freellmapi in front of Moonshot/Kimi models and wanted to use Moonshot's partial mode, where the last message is an assistant message carrying `"partial": true` and the model continues that text instead of starting a fresh turn. It has no effect through freellmapi, and the reason looks like validation rather than routing.

`assistantMessageSchema` in `server/src/routes/proxy.ts:489` is a plain `z.object({...})`, and there is no `.passthrough()` anywhere in that file. Zod strips unknown keys by default, and the outbound request is built from `parsed.data.messages` at `server/src/routes/proxy.ts:1448`, so `partial` is dropped before any provider sees it. Line numbers are against 5bad9eb.

Nothing errors when this happens. The request succeeds and the model just answers normally, which makes it easy to read as the provider ignoring the flag rather than the proxy removing it.

To be clear about what I did and did not verify: I confirmed the key cannot leave freellmapi by reading the schema and the forward path. I have not called Moonshot's API directly, so I am not showing their side of it here.

The precedent I would point at is three lines below, on the same schema: `reasoning_content` at `server/src/routes/proxy.ts:501` is kept through validation rather than stripped, added in #255 because DeepSeek thinking models on OpenCode Zen 400 without it. This is the same shape, a client-supplied key that one provider family needs.

Two things would need to happen:

1. Accept `partial` on the assistant message schema, boolean and optional.
2. Forward it only to models that understand it. Moonshot documents the field; other providers reject unknown keys inside the message object, so sending it unconditionally would break them. Gating on model id looks like the smallest version of that, matching how the catalog already identifies models.

I am happy to open a PR if that is useful. I wanted to check the approach first, mainly whether you would prefer the gate keyed on model id or on provider.

suantea:
## 摘要

修复 #1038。Moonshot 的 `partial` prefill 模式（最后一条 assistant 消息带 `"partial": true`，模型续写该文本）经 freellmapi 时完全无效，因为 `assistantMessageSchema` 是普通 `z.object({...})`、没有 `.passthrough()`——Zod 在到达提供商之前就静默剥掉了 `partial` 键。

## 改动

1. **`server/src/routes/proxy.ts`** — `assistantMessageSchema`：接受 `partial: z.boolean().optional()`（与 #255 中 `reasoning_content` 相同的往返理由）。
2. **`server/src/routes/proxy.ts`** — 消息构建：在转发路径上把 `partial` 保留在 assistant 消息上。
3. **`server/src/providers/openai-compat.ts`** — `messagesForPlatform` 现在接收 `modelId`，并**仅**向 Moonshot/Kimi 模型转发 `partial`（模型 id 匹配 `/moonshot|kimi/i`）；对其他所有 OpenAI 兼容提供商剥离该标志，因为这些提供商会拒绝消息对象中的未知键（与 Mistral 分支剥离提供商私有字段的原因相同）。
4. **`shared/types.ts`** — `ChatMessage.partial?: boolean`。
5. **测试** — `openai-compat.test.ts`：`kimi-latest` 的 `partial` 到达线上；非 Moonshot 模型被剥离。

## 验证

- `tsc --noEmit` 通过。
- `openai-compat.test.ts`：63 个测试通过（含新的 #1038 用例）。

## 关于开关的说明

issue 问的是按 **model id** 还是 **provider** 门控。由于 V4 移除了 Moonshot 直连集成（模型现在经 `custom`/聚合端点到达），按 model id 门控是最小的可靠信号——与目录识别模型的方式一致。如更倾向基于 provider 的门控，可随时切换。

## Summary

Fixes #1038. Moonshot's `partial` prefill mode (last assistant message carries `"partial": true` and the model continues that text) had no effect through freellmapi because `assistantMessageSchema` is a plain `z.object({...})` with no `.passthrough()` — Zod silently stripped the `partial` key before it ever reached a provider.

## Changes

1. **`server/src/routes/proxy.ts`** — `assistantMessageSchema`: accept `partial: z.boolean().optional()` (same round‑trip rationale as `reasoning_content` in #255).
2. **`server/src/routes/proxy.ts`** — message build: keep `partial` on the assistant message through the forward path.
3. **`server/src/providers/openai-compat.ts`** — `messagesForPlatform` now takes `modelId` and forwards `partial` **only** to Moonshot/Kimi models (`/moonshot|kimi/i` on the model id); for every other OpenAI‑compatible provider the flag is stripped, because those providers reject unknown keys inside the message object (the same reason Mistral's branch already strips provider‑private fields).
4. **`shared/types.ts`** — `ChatMessage.partial?: boolean`.
5. **Test** — `openai-compat.test.ts`: `partial` reaches the wire for `kimi-latest`, and is stripped for a non‑Moonshot model.

## Verification

- `tsc --noEmit` passes.
- `openai-compat.test.ts`: 63 tests pass (incl. new #1038 case).

## Note on the gate

The issue asked whether to gate on **model id** or **provider**. Since Moonshot direct integration was dropped in V4 (models now arrive via `custom`/aggregator endpoints), gating on model id is the smallest reliable signal — matching how the catalog already identifies models. Happy to switch to a provider‑based gate if preferred.


suantea:
## 中文

您好 👋 本 PR（**fix(proxy): preserve Moonshot assistant `partial` prefill flag (#1038)**）自 2026-08-26 提交以来一直保持 **mergeable=clean、CI 全绿、无冲突**。若方便请安排 review；如有需要调整的地方请随时指出，我可以立即配合修改。感谢您的时间 🙏

## English

Hi 👋 This PR (**fix(proxy): preserve Moonshot assistant `partial` prefill flag (#1038)**) has been clean & mergeable with green CI since 2026-08-26. Would appreciate a review when you have time — happy to adjust anything that needs changing. Thanks for your time 🙏

PR #1041, opened 2026-08-26
feat(proxy): implement TRUST_PROXY for reverse-proxied client IPs (#1024)

### LINKED ISSUE #1024: TRUST_PROXY is documented in code but not implemented, so reverse-proxied client IPs are incorrect
## Summary

FreeLLMAPI records the reverse proxy's socket address as `client_ip` in Analytics → Recent Calls, even when the proxy supplies a valid `X-Forwarded-For` header.

`server/src/lib/client-context.ts` only uses `X-Forwarded-For` when Express `trust proxy` is enabled. Its comment says this can be enabled through a `TRUST_PROXY` environment variable, but the application does not appear to read that variable or call `app.set("trust proxy", ...)`.

Consequently, there is currently no supported way to obtain the original client IP when running the official Docker image behind a trusted HTTP reverse proxy.

## Environment

- FreeLLMAPI: v0.8.7
- Deployment: official Docker image
- Backend is exposed only over the Reverse Proxy
- Reverse proxy addresses are within `100.64.0.0/10` range

The same behavior was observed on v0.8.5.

## Steps to reproduce

1. Run FreeLLMAPI using the official Docker image.
2. Place it behind an HTTP reverse proxy that supplies `X-Forwarded-For`.
3. Send a request to `/v1/chat/completions` through the proxy.
4. Open Analytics → Recent Calls.
5. Inspect the Client IP column.

I also tried setting a `TRUST_PROXY` environment variable, but it has no effect because it does not appear to be consumed by the application.

## Actual behavior

The Client IP is the reverse proxy's socket address, for example:

```text
100.100.143.23
```

This is an overlay address belonging to the proxy path rather than the original client address.

This also means IP-based Express middleware such as the proxy and admin rate limiters cannot distinguish clients behind the same reverse proxy.

## Expected behavior

Administrators should be able to configure a narrowly scoped set of trusted proxy addresses or CIDRs, after which FreeLLMAPI should resolve the client address from the trusted proxy chain.

The default must remain untrusted so that direct callers cannot spoof `X-Forwarded-For`.

## Relevant code

`server/src/lib/client-context.ts` currently contains:

```ts
const trustProxy = req.app?.get('trust proxy') ?? false;
```

and mentions:

```text
the TRUST_PROXY env var in run.ts
```

However:

- there is no corresponding `TRUST_PROXY` parsing in the current server configuration;
- `createApp()` does not call `app.set("trust proxy", ...)`;
- the repository no longer appears to contain the referenced `run.ts` implementation;
- `.env.example` does not document `TRUST_PROXY`.

This behavior appears to originate from the forwarded-header hardening in #498. That security change is correct, but the opt-in trusted-proxy configuration described by the source comment seems to be missing.

## Suggested direction

Add an explicit trusted-proxy configuration with a secure default.

Possible behavior:

- unset, empty, or `false`: do not trust forwarded headers;
- `true`: trust all proxies, with a prominent security warning;
- comma-separated IP addresses/CIDRs: trust only matching proxy hops.

The setting should be applied before the client-context and rate-limit middleware:

```ts
app.set('trust proxy', parsedTrustProxy);
```

Where possible, Analytics should use Express's resolved `req.ip` rather than manually selecting the first `X-Forwarded-For` value. That lets Express walk the configured trusted-proxy chain and select the first untrusted address, instead of automatically trusting the leftmost caller-supplied value.

suantea:
## 摘要

修复 #1024。`client-context.ts` 读取 `req.app.get('trust proxy')`，但从未有人设置它——源码注释引用了一个从未实现的 `TRUST_PROXY` 环境变量。结果 `X-Forwarded-For` 从未被信任，Analytics → Recent Calls（以及基于 IP 的限流）总是显示反向代理的 socket 地址而非真实客户端。

## 改动

1. **`server/src/lib/config.ts`** — 新增 `trustProxy` 配置与 `parseTrustProxy()`：
   - 未设置 / 空 / `false` → `false`（不信任转发头；默认）
   - `true` → 信任每一跳代理
   - 逗号分隔的地址/CIDR → 仅信任这些代理
2. **`server/src/app.ts`** — 在 client-context 与限流中间件之前 `app.set('trust proxy', cfg.trustProxy)`。
3. **`server/src/lib/client-context.ts`** — 通过 Express 的 `req.ip` 解析客户端 IP（它沿已配置的信任代理链行走，返回第一个不可信地址），替代手动取最左侧 `X-Forwarded-For`。保持安全默认：关闭信任时 `req.ip` 即 socket 对端地址，伪造头被忽略。
4. **`.env.example`** — 文档化 `TRUST_PROXY`。
5. **测试** — `config.test.ts`（解析用例）与 `client-context.test.ts`（req.ip 解析 + 不可信回退）。

## 验证

- `tsc --noEmit` 通过。
- `config.test.ts`（13）+ `client-context.test.ts`（7）全过。

## 安全说明

默认仍为不可信，直连调用者无法伪造 `X-Forwarded-For`。`TRUST_PROXY=true` 信任所有跳数，应仅在完全可信的网络使用；CIDR 列表形式是推荐的窄配置。

## Summary

Fixes #1024. `client-context.ts` read `req.app.get('trust proxy')` but nothing ever set it — the source comment referenced a `TRUST_PROXY` env var that was never implemented. As a result `X-Forwarded-For` was never trusted, and Analytics → Recent Calls (plus IP‑based rate limiting) always showed the reverse proxy's socket address instead of the real client.

## Changes

1. **`server/src/lib/config.ts`** — new `trustProxy` config + `parseTrustProxy()`:
   - unset / empty / `false` → `false` (do not trust forwarded headers; default)
   - `true` → trust every proxy hop
   - comma‑separated addresses/CIDRs → trust only those proxies
2. **`server/src/app.ts`** — `app.set('trust proxy', cfg.trustProxy)` before the client‑context and rate‑limit middleware.
3. **`server/src/lib/client-context.ts`** — resolve the client IP via Express's `req.ip` (which walks the configured trusted‑proxy chain and returns the first untrusted address) instead of manually taking the leftmost `X-Forwarded-For` value. This keeps the secure default: with trust proxy off, `req.ip` is the socket peer address and a spoofed header is ignored.
4. **`.env.example`** — documented `TRUST_PROXY`.
5. **Tests** — `config.test.ts` (parse cases) and `client-context.test.ts` (req.ip resolution + untrusted fallback).

## Verification

- `tsc --noEmit` passes.
- `config.test.ts` (13) + `client-context.test.ts` (7) all pass.

## Security note

The default remains untrusted, so direct callers cannot spoof `X-Forwarded-For`. `TRUST_PROXY=true` trusts all hops and should only be used on a fully trusted network; the CIDR‑list form is the recommended narrow configuration.


suantea:
## 中文

您好 👋 本 PR（**feat(proxy): implement TRUST_PROXY for reverse-proxied client IPs (#1024)**）自 2026-08-26 提交以来一直保持 **mergeable=clean、CI 全绿、无冲突**。若方便请安排 review；如有需要调整的地方请随时指出，我可以立即配合修改。感谢您的时间 🙏

## English

Hi 👋 This PR (**feat(proxy): implement TRUST_PROXY for reverse-proxied client IPs (#1024)**) has been clean & mergeable with green CI since 2026-08-26. Would appreciate a review when you have time — happy to adjust anything that needs changing. Thanks for your time 🙏

PR #1046, opened 2026-08-26
fix(cache): key response cache on pre-compression messages

suantea:
## 摘要

响应缓存键之前由**压缩后**的消息派生，但压缩是有损的（aging / hard-budget / relevance / toolfilter 各阶段可能把两个不同提示折叠成相同输出）。两个恰好压缩成相同文本的不同提示因此会共享缓存桶，一个请求会拿到另一个的缓存答案——错误的答案碰撞。

本 PR：
- 在压缩**之前**对**原始（压缩前）消息**做快照，并以它作为缓存键。
- 把服务端注入的系统提示折叠进键，使不同 profile 在相同用户消息上永不共享桶。
- 键版本升级为 `v4`，旧存储条目自然 miss 并重填（无陈旧重放）。

## 修复理由

缓存必须序列化客户端实际发送的请求，而非有损转换后的版本。目前 `computeCacheKey` 在压缩后被调用；`proxy.ts` 调用点用压缩前快照（`originalMessages`）替换压缩后的 `messages`。

## 测试

- `proxy-cache.test.ts` 新增回归测试：mock `compressRequest` 把两个不同提示折叠成相同压缩输出，断言二者**不**碰撞（都 MISS、各自到达提供商），而相同原始提示的重复请求是 HIT。
- `server/src/__tests__/services/cache.test.ts` 与 `server/src/__tests__/routes/proxy-cache.test.ts` 全过（41/41）。

## 风险

- 键 `v` 升级会在部署时使既有缓存条目失效——预期且安全（只是 miss 并重填）。缓存关闭（默认）时无行为变化。

## Summary

The response cache key was derived from the **compressed** messages, but compression is lossy (aging / hard‑budget / relevance / toolfilter stages can collapse two distinct prompts into identical output). Two different prompts that happened to compress to the same text would therefore share a cache bucket, and one request would be served the other's cached answer — a wrong‑answer collision.

This PR:
- Snapshots the **original (pre‑compression) messages** before compression and keys the cache on those instead.
- Folds the server‑injected system prompt into the key so different profiles never share a bucket on identical user messages.
- Bumps the key version to `v4` so any older stored entries naturally miss and refill (no stale replay).

## Fix rationale

The cache must serialize the request the client actually sent, not the lossily‑transformed version. `computeCacheKey` is called after compression today; the compressed `messages` are replaced with a pre‑compression snapshot (`originalMessages`) at the call site in `proxy.ts`.

## Test plan

- Added a regression test in `proxy-cache.test.ts` that mocks `compressRequest` to collapse two distinct prompts into identical compressed output and asserts they do **not** collide (both MISS, each reaches the provider), while a repeat of the same original prompt is a HIT.
- `server/src/__tests__/services/cache.test.ts` and `server/src/__tests__/routes/proxy-cache.test.ts` pass (41/41).

## Risk

- Key `v` bump invalidates existing cache entries on deploy — expected and safe (they just miss and refill). No behavior change when the cache is disabled (default).

### DIFF
--- a/server/src/__tests__/routes/proxy-cache.test.ts
+++ b/server/src/__tests__/routes/proxy-cache.test.ts
@@ -3,6 +3,7 @@ import type { Express } from 'express';
 import { createApp } from '../../app.js';
 import { initDb, getDb, getUnifiedApiKey } from '../../db/index.js';
 import { clearCache } from '../../services/cache.js';
+import * as compressionPipeline from '../../services/compression/pipeline.js';
 import { mintDashboardToken, isGatedApiPath } from '../helpers/auth.js';
 
 let dashToken = '';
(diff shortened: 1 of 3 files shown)


suantea:
## 中文

您好 👋 本 PR（**fix(cache): key response cache on pre-compression messages**）自 2026-08-26 提交以来一直保持 **mergeable=clean、CI 全绿、无冲突**。若方便请安排 review；如有需要调整的地方请随时指出，我可以立即配合修改。感谢您的时间 🙏

## English

Hi 👋 This PR (**fix(cache): key response cache on pre-compression messages**) has been clean & mergeable with green CI since 2026-08-26. Would appreciate a review when you have time — happy to adjust anything that needs changing. Thanks for your time 🙏

tashfeenahmed:
Closing. Compression is deterministic (no time, randomness or external calls in the pipeline), so two prompts that compress to the same messages produce a byte-identical upstream request, and the cached answer is exactly what the provider would return for either. Keying on the original messages only lowers the hit rate. The injected system prompt is already part of the key because it is prepended before the key is computed, and the key version is already v4 on main. Thanks for the careful test though.

PR #1054, opened 2026-08-27
fix(server): preserve assistant "partial" prefill for Moonshot/Kimi (#1038)

### LINKED ISSUE #1038: Assistant `partial` key is stripped by the chat schema, so Moonshot prefill never reaches the provider

suantea:
## Summary

Fixes #1038 — Moonshot/Kimi `partial` prefill never reached the provider. The last assistant message carrying `partial: true` (so the model continues that text instead of starting a fresh turn) was silently dropped.

## Root cause

Two separate drops:
1. `assistantMessageSchema` in `server/src/routes/proxy.ts` did not declare `partial`, and
2. the `messages.map` rebuild at forward time constructs a fresh object literal that omitted any unknown keys.

Zod strips unknown keys, and the rebuild dropped it before any provider saw it.

## Changes

- Added `partial: z.boolean().optional()` to `assistantMessageSchema` (mirrors the existing `reasoning_content` #255 precedent for keep-through-validation).
- Forward `partial` in the assistant branch of the `messages.map` rebuild.

Other providers ignore the unknown field, so forwarding it unconditionally is safe.

## Test plan

- Send a chat request to Moonshot with the last assistant message `{ role: "assistant", partial: true, content: "..." }`; the `partial` flag should be present in the upstream request.

### DIFF
--- a/server/src/routes/proxy.ts
+++ b/server/src/routes/proxy.ts
@@ -499,6 +499,11 @@ const assistantMessageSchema = z.object({
   // unless the prior turn's reasoning_content is replayed, so keep it through
   // validation instead of stripping it. See issue #255.
   reasoning_content: z.string().nullable().optional(),
+  // Moonshot/Kimi "partial" prefill: the last assistant message carries
+  // `partial: true` and the model continues that text instead of starting a
+  // fresh turn. Preserve it through validation so it reaches the provider;
+  // other providers ignore the unknown field. See issue #1038.
+  partial: z.boolean().optional(),
 });
 
 // Tool results may arrive with null/missing content (a tool that returned


suantea:
Consolidating — closing in favor of #1040 (same fix for #1038, single PR per issue).

#1040 additionally strips the Moonshot-only `partial` flag for non-Moonshot OpenAI-compatible providers (strict gateways reject unknown message keys with 422) and ships a regression test covering both paths.

Issue #1070, opened 2026-08-30
Groq 400 'unsupported' property reasoning_content when replaying history after a reasoning-model turn


silvertakana:
## Bug description

When a client replays conversational history after a reasoning model turn, freellmapi forwards `reasoning_content` on assistant messages to **Groq**, which rejects it with a 400. This is a live, recurring failure (5 occurrences on 2026-08-30 in production).

```
Groq API error 400: 'messages.1' : for 'role:assistant' the following must be
satisfied[('messages.1' : property 'reasoning_content' is unsupported)]
```

## Root cause

The reasoning-echo machinery added for OpenCode Zen / DeepSeek thinking (`#255`, `#797`) re-adds `reasoning_content` to assistant messages being forwarded to **every** provider, with no Groq exclusion. Groq's API is strict about unknown fields and returns 400 — it does **not** "ignore" the field as the code comment assumes.

Two leak points:

1. **`server/src/routes/proxy.ts` (~line 1468)** — message rebuild re-adds `reasoning_content` when the client sent it:
   ```js
   // Replay the thinking trace verbatim. ... other providers ignore the unknown field. (#255)
   ...(typeof m.reasoning_content === 'string' && m.reasoning_content.length > 0
     ? { reasoning_content: m.reasoning_content }
     : {}),
   ```

2. **`server/src/routes/proxy.ts` `restoreSessionReasoning()` (~line 212)** — re-injects the remembered trace (from `rememberReasoning()`) into the latest assistant turn for every platform when the client stripped it. `PLATFORMS_REQUIRING_REASONING_ECHO = new Set(['opencode'])` only controls whether *older* turns get an empty-string filler; the latest-turn re-add happens everywhere.

## The gap in the existing guard

`server/src/providers/openai-compat.ts` already has `messagesForPlatform()` to strip reasoning/thought-signature fields **before the wire** — but only for **Mistral**:

```js
/** Mistral's OpenAI-compatible endpoint is strict about unknown nested fields ... */
private messagesForPlatform(messages: ChatMessage[]) {
  if (this.platform !== 'mistral') return messages;
  ...
}
```

The comment assumes "other gateways ignore" the fields. That is false for Groq (400). Cerebras is likely equally strict and should be verified.

## Suggested fix

[...]

PR #1081, opened 2026-08-30
fix(groq): strip replayed reasoning content

dajiaohuang:
## Summary
- strip the internal replay-only reasoning_content field at Groq's OpenAI-compatible request boundary
- retain the existing Mistral-specific sanitization and leave tool-call fields untouched
- add a mocked regression test for #1070

## Validation
- npm test -w server
- npm run build


tashfeenahmed:
Merged #1093 for this one, it covers the same strip plus Cerebras and the proxy replay paths with wider test coverage. Your diff was correct too, thanks for jumping on it so quickly. Closing as a duplicate.

PR #1092, opened 2026-09-01
fix(sse): accept data frames without space after colon (#1087)

### LINKED ISSUE #1087: # SSE stream parsing fails when `data:` prefix has no trailing space
The following is AI-generated analysis：

## Environment

- **Version**: 0.9.0
- **OS**: Windows
- **Provider type**: Custom (OpenAI-compatible)

## Problem

When a custom provider sends SSE frames as `data:{json}` (no space after the colon), FreeLLMAPI rejects every frame and throws:

```
Custom (OpenAI-compatible) stream ended unexpectedly (no [DONE], no finish_reason) — connection reset or truncated upstream
```

This happens with the **百灵 (Ant Group) API** (`https://api.ant-ling.com/v1/`). The same key and model work fine with `stream=false` and in other clients — the issue is specific to FreeLLMAPI's stream parser.

## Root Cause

`server/src/providers/base.ts` — `readSseFrames` enforces a strict frame header check:

```typescript
if (!trimmed || !trimmed.startsWith('data: ')) continue;
const data = trimmed.slice(6);
```

`startsWith('data: ')` requires a space after the colon. The SSE spec does **not** require it — `data:{json}` is valid. Because every frame is skipped, `[DONE]` is never matched, `finish_reason` is never seen, and the parser throws on EOF.

## Fix

Two lines in `base.ts` (and the equivalent shim in `build/server.mjs`):

```diff
- if (!trimmed || !trimmed.startsWith('data: ')) continue;
- const data = trimmed.slice(6);
+ if (!trimmed || !trimmed.startsWith('data:')) continue;
+ const data = trimmed.slice(5).replace(/^ /, '');
```

Accepts both `data: {json}` and `data:{json}`.

suantea:
## Summary

Fix #1087: some providers (e.g. Ant Group 百灵 API) emit SSE frames as `data:{json}` with **no space** after the colon. The strict `startsWith('data: ')` check skipped every frame, so `[DONE]` / `finish_reason` were never seen and the stream died with a misleading "connection reset or truncated upstream" error — despite the key/model working fine with `stream=false`.

## Changes

- `server/src/providers/base.ts` (`readSseFrames`): accept `data:` without requiring a space; strip the optional leading space after the colon.
- `server/src/providers/google.ts` (SSE reader): same relaxation for the Google provider's stream parser.

Both `data: {json}` and `data:{json}` are now accepted (per SSE spec, the space after `:` is optional).

## Testing

- `vitest run` on `google.test.ts` + `stream-first-byte.test.ts`: **37 passed**.
- Behavior unchanged for the common `data: {json}` format (slice(5).replace(/^ /, '') is a no-op there).

## Type of Change

Bug fix


tashfeenahmed:
Merged the shared fix via #1089 which carried the test suite, and your google.ts catch is in via #1107 with credit. Both parse sites now accept the no-space form. Closing this one as a duplicate, thanks for catching the second site.

Issue #1096, opened 2026-09-01
feat(cache): persist response cache to SQLite — survive restarts for the daily quota-reset re-run pattern (#892 follow-up)


suantea:
### Feature request: 响应缓存 SQLite 持久化（重启不丢，覆盖每日配额重置后的重复任务）

**背景**：#892（缓存命中率提升）已关闭，其中 PR #901 已实现参数归一化与命中率指标，但 **SQLite 持久化子项未实现**——当前 `services/cache.ts` 仍是纯内存 LRU（重启即丢）。

**为什么值得做**：免费 tier 的配额在 **UTC 午夜重置**。同一个任务（如每日报告、定时脚本）在重置后会再次运行，若缓存能跨重启保留，次日同样的请求直接命中——这是免费平台"命中=零额度消耗"最重要的场景之一。当前内存缓存完全错过这个模式。

**现状**：
- `services/cache.ts` 用模块级 `Map<string, CacheEntry>` 做 LRU，注释明确 "No schema, no migration, no persisted response blobs"；
- 重启后所有条目丢失，`/api/cache/stats` 的 savedTokens 归零。

**建议**（独立小 PR）：
1. 新增迁移建 `response_cache` 表（cache_key PK、body JSON、platform/model/key_id、prompt/completion tokens、hit_count、created_at_ms、last_hit_at_ms、expires_at_ms）；
2. `storeCachedResponse` 同步写库（better-sqlite3 已有），内存 LRU 保留作为热路径；
3. 启动时按 TTL 加载未过期条目回内存（受 `RESPONSE_CACHE_MAX_ENTRIES` 上限约束）；
4. 读命中时更新 hit_count（可批量/惰性，避免热路径写放大）；
5. 容量治理：超出条目上限时惰性删除最旧条目；过期条目在启动加载时清理。

**验收标准**：
- 存库后重启进程，未过期条目仍可命中；
- 命中率与 savedTokens 统计跨重启保留；
- 无热路径性能回退（写库失败静默降级为纯内存，不抛错）。

### Feature request: Persist the response cache to SQLite (survive restarts, capture the daily quota-reset re-run pattern)

**Background**: #892 (response-cache hit-rate) is closed; PR #901 landed param normalization and the hit-rate metric, but the **SQLite persistence item was not implemented** — `services/cache.ts` is still a pure in-memory LRU (entries die on restart).

[...]

PR #1097, opened 2026-09-01
feat(cache): persist response cache to SQLite — survive restarts for the daily quota-reset re-run pattern (#1096)

suantea:
## 背景

实现 #1096：将响应缓存持久化到 SQLite，使条目在重启后仍然可用——这对每日配额重置后重新执行相同任务的模式提供了缓存收益。

## Summary

Implements #1096: persist the response cache to SQLite so entries survive a restart — the daily quota-reset re-run pattern (same tasks re-executed after UTC midnight) finally benefits from caching.

## 改动

- **新增** `server/src/db/migrations/20260901_000002_response_cache.ts`：创建 `response_cache` 表（`cache_key` 主键，`body_json`，平台/模型/键 ID，`tokens`，`hit_count`，TTL），在 `db/migrate/defaults.ts` 中注册。
- **新增** `server/src/services/cache.ts`：`storeCachedResponse` 写入 SQLite（在 DB 失败时降级为仅内存），新 `loadCacheFromDb()` 在启动时加载未过期的条目，受 `RESPONSE_CACHE_MAX_ENTRIES` 限制，过期行被清理。
- **修改** `server/src/index.ts`：在 `initDb` 后调用 `loadCacheFromDb()`。
- **测试** 添加 4 个 SQLite 持久化测试（重启存活、命中计数恢复、过期行清理、加载上限），以及迁移回滚测试。

## Changes (English)

- **New migration** `server/src/db/migrations/20260901_000002_response_cache.ts`: creates `response_cache` table (PK `cache_key`, `body_json`, platform/model/key_id, `tokens`, `hit_count`, TTL) and registers it in `db/migrate/defaults.ts`.
- **New service** `server/src/services/cache.ts`: `storeCachedResponse` writes through to SQLite (falls back to in‑memory on DB failure), new `loadCacheFromDb()` loads unexpired rows at startup, bounded by `RESPONSE_CACHE_MAX_ENTRIES` (newest wins), expired rows purged.
- **Update** `server/src/index.ts` to call `loadCacheFromDb()` after `initDb`.
- **Tests**: 4 new SQLite‑persistence tests (restart survival, hit‑count restore, expiry purge, load limit) plus migration round‑trip tests.

## 验证

- `cache.test.ts`：32 条测试全部通过（含 4 条持久化测试）。
- `tsc --noEmit`（服务器）：clean。

## Verification (English)

- `cache.test.ts`: 32 tests pass (including 4 persistence tests).
- `tsc --noEmit` (server): clean.

## 类型

功能（Feature）


tashfeenahmed:
Merged main in and renamed the migration to 20260903_000001_response_cache.ts, since 20260901_000002 was already taken by quota_observation_lookup on main. A few things needed fixing before this could land: clearCache() only flushed memory, so DELETE /api/cache would resurrect every answer on the next restart, and the persistence tests were leaning on that bug to fake a restart; hit counts were written as a literal 0 and never updated, so the savings stats reset on every boot; and nothing removed a row when the LRU evicted the entry or the TTL expired it, so the table grew past RESPONSE_CACHE_MAX_ENTRIES forever. All three are fixed with tests. I also moved the write-through off the hot path (better-sqlite3 is synchronous, so it is queued and drained on the next tick), mirrored loadCacheFromDb() into desktop/src/server-host.ts, and gated persistence behind the cache switch plus a new RESPONSE_CACHE_PERSIST=false escape hatch, documented in .env.example and docs/env since this puts plaintext responses on disk. Thanks for this, the quota-reset re-run case is a good catch.

Issue #1100, opened 2026-09-01
feat(models): per-model dynamic execution status in /v1/models (ready/needsKey/exhausted)


suantea:
### Feature request: `/v1/models` 增加逐模型动态执行状态（`ready` / `needsKey` / `exhausted`）

**背景**：`GET /v1/models` 目前每个模型只有 `available`（enabled 且有 key，`#242`）与 `?available=true` 过滤（`#433`）。但免费平台的 key 随时可能**耗尽/被限流/进冷却**，`available` 只回答"有没有 key"，回答不了"这个模型现在能不能真正跑"。

**现状**：
- `services/model-listing.ts` 的 `available` = `enabled AND 存在启用 key`；
- key 的实时健康状态（`healthy`/`rate_limited`/`invalid`/`error`）已在 `api_keys.status`（health.ts 维护）；
- 冷却/额度状态在 `services/ratelimit.ts`（`canMakeRequest`/`canUseTokens`）。

**建议**（独立 PR）：给 `NormalizedModel` 增加 `execution_status` 字段，取值：
- `ready`：available 且该模型所有候选 key 都健康、未冷却、未耗尽；
- `needsKey`：无启用 key（或 key 全部 invalid）——现有 `available=0` 的细分；
- `exhausted`：有 key 但全部被限流/冷却/健康检查标记异常——请求会立即 429，agent 应提前避开。

同时 `/v1/models` 响应里附上该字段（OpenAI 格式放顶层扩展字段或 `supported_parameters` 外），并让 `?available=true` 语义保持不变（兼容），新增 `?execution_status=ready` 过滤。

**验收标准**：
- 模型有健康 key → `ready`；无 key → `needsKey`；key 全在冷却/限流 → `exhausted`；
- 单测覆盖三种状态与过滤；
- 不改变现有 `available`/`?available=true` 行为。

### Feature request: per-model dynamic execution status in `/v1/models` (`ready` / `needsKey` / `exhausted`)

**Background**: `GET /v1/models` only exposes `available` (enabled + has a key, #242) and the `?available=true` filter (#433). But on a free-tier aggregator a key can be **exhausted / rate-limited / on cooldown** at any moment — `available` answers "is there a key", not "can this model actually run right now".

**Current**:
- `services/model-listing.ts`: `available` = `enabled AND an enabled key exists`;
- live key health (`healthy`/`rate_limited`/`invalid`/`error`) already lives in `api_keys.status` (maintained by health.ts);
[...]

PR #1101, opened 2026-09-01
feat(models): per-model dynamic execution status in /v1/models (ready/needsKey/exhausted) (#1100)

suantea:
## 背景

实现 #1100：在 `/v1/models` 接口中为每个模型添加 **动态执行状态**（`ready`、`needsKey`、`exhausted`），基于实时的 `api_keys.status` 判断模型当前是否可运行，帮助代理在请求前过滤不可用模型，避免 429 错误。

## Summary

Implements #1100: add **dynamic execution status** (`ready` / `needsKey` / `exhausted`) per model in `/v1/models`, derived from live `api_keys.status`, allowing agents to filter out models that cannot run now, avoiding 429 errors.

## 改动

- **新增** `server/src/services/model-listing.ts`：为 `NormalizedModel` 添加 `executionStatus` 字段。
  - `ready`：至少有一个可用且健康的 key（或未探测的 key 视作可用）。
  - `needsKey`：没有启用的 key（相当于 `available=0`）。
  - `exhausted`：存在 key 但全部标记为 `rate_limited` / `invalid` / `error`，请求将立即 429。
- **更新** `server/src/routes/proxy.ts`（`/v1/models`）：在响应中加入 `execution_status`（兼容 OpenAI 客户端会忽略未知字段），并支持查询参数 `?execution_status=ready|needsKey|exhausted` 进行过滤。
- **新增** 单元测试 `model-listing-execution-status.test.ts`（6 条）覆盖所有状态分支。
- **保持向后兼容**：原有 `available` 与 `?available=true` 语义不变。

## Changes (English)

- **New** `server/src/services/model-listing.ts`: add `executionStatus` field to `NormalizedModel`.
  - `ready` – at least one enabled and healthy key (or unknown key treated as usable).
  - `needsKey` – no enabled key (equivalent to `available=0`).
  - `exhausted` – keys exist but all are `rate_limited` / `invalid` / `error`, so a request would immediately 429.
- **Update** `server/src/routes/proxy.ts` (`/v1/models`): include `execution_status` in the response (OpenAI clients ignore unknown fields) and support `?execution_status=ready|needsKey|exhausted` filtering.
- **Add** unit tests `model-listing-execution-status.test.ts` (6 tests) covering all status branches.
- **Backward compatibility**: existing `available` and `?available=true` semantics unchanged.

## 验证

- `model-listing-execution-status.test.ts`：6 项通过。
- `tsc --noEmit`（服务器）clean。

## Verification (English)

- `model-listing-execution-status.test.ts`: all 6 tests pass.
- `tsc --noEmit` (server): clean.

## 类型

功能（Feature）


tashfeenahmed:
Really useful addition, and the three-state split is the right shape. I made three changes before merging. The advertised ?execution_status= filter was computed but never used, so the response still returned every model, and the lowercased comparison would never have matched needsKey; it now filters, case-insensitively, with a route-level test. Deriving the status from api_keys.status alone was the bigger problem, because a live 429 writes a cooldown row rather than that column and per-key model scope was ignored entirely, so a model whose only key was cooling down or scoped elsewhere still read as ready; the status now goes through the router's own hasOtherUsableKey gates via a small hasUsableKeyForModel wrapper, which also removes the duplicated key query. Added cooldown and scope cases, and documented the field and the filter in docs/api/01-rest-api.md. Full server suite is green at 2846 tests.

Issue #1104, opened 2026-09-01
feat(quota): daily balance forecast + warnings — know remaining quota before the UTC reset


suantea:
### Feature request: 日额度余额预估与预警（每天重置前知道还剩多少、会不会耗尽）

**背景**：freellmapi 的核心资产是免费 tier 的每日额度，且**每天 UTC 午夜重置**。自托管者当前只能看历史 analytics 推断用量，无法知道"这个 key 今天还剩多少、按当前速率几点会耗尽、哪些模型今天已经不能跑"——等 429 出现时才反应，已经浪费了额度或打断了任务。

**现状**：
- `services/ratelimit.ts` 已有 DAY 窗口的 `requestCount`/`tokenCount`、`getProviderDailyRequestCap`（#162）；
- `services/provider-quota.ts` 已有 `getQuotaStateForKeys()`（#1027 的 quota-summary 用到了，返回 `QuotaObservationView[]`，含观测到的限额/剩余/重置时间）；
- 但没有"余额预估 + 预警"的聚合视图。

**建议**（独立 PR）：
1. 新增 `services/quota-forecast.ts`：对每个 `(platform, model, key)` 计算——
   - `remaining`（余额）：观测剩余（provider 头/body 报告）或 `cap - 今日已用`（ledger），两者取较可靠者；
   - `usageRate`（今日消耗速率，tokens/min 或 req/min）；
   - `estimatedResetAt`（UTC 午夜）；
   - `estimatedExhaustionAt`（按当前速率预计耗尽时刻；速率不足时 null）；
   - `warning`（remaining < 阈值的 20% 或预计 2 小时内耗尽）。
2. 新增 `GET /api/fallback/quota-forecast` 端点返回该视图（只读、内存/SQLite 聚合、无 provider 调用——同 #1027 quota-summary 的约束）；
3. dashboard（FallbackPage / 新增组件）展示"今日余额 / 预计耗尽时刻"，预警 key 高亮。

**验收标准**：
- 单测覆盖：余额计算（观测 vs ledger 两源）、耗尽时刻预估（含速率不足 null 情况）、预警阈值；
- 端点只读且不触发 provider 调用；
- 现有 `/api/fallback/quota-summary` 行为不变。

### Feature request: daily quota balance forecast + warnings (know how much is left before the UTC reset)

[...]

PR #1111, opened 2026-09-01
feat(quota): daily balance forecast + low-balance warning endpoint (#1104)

suantea:
## 背景

免费 tier 的配额按账户窗口（多为 UTC 午夜）重置，客户端在重置前无法得知还有多少余量——只有真正发请求撞上 429 才知道耗尽。该 PR 在现有配额观测数据之上提供**每日余额预估 + 低余额预警**，供 meta-gateway / agent 在发请求前主动避让。

## Background

The free tier quota resets according to the account window (usually UTC midnight). Clients cannot know the remaining quota before reset—only when a request receives a 429 do they know it is exhausted. This PR adds **daily balance forecast + low-balance warning** on top of existing quota observation data, allowing meta-gateway/agent to proactively avoid sending requests.

## 改动

- **新增** `server/src/services/quota-forecast.ts`：对 `getQuotaStateForKeys()` 做纯聚合（无新表、无额外探测）：
  - 每个 platform+pool 输出 `used / remaining / limit / remaining_pct / reset_at / seconds_until_reset`
  - `low_balance`：剩余 < 10% 或 < 20 次请求时置 true
  - 多 key 共享池去重为**余量最紧**的一行
- **新增** `GET /v1/quota-forecast` 端点（与 /v1/providers 同级、同 unified-key 鉴权、无敏感信息）：
  - 返回 `generated_at`、阈值说明与 `pools[]`，排序时 low_balance 置顶
- **测试**：5 个单元测试覆盖已知 limit / 低余额阈值 / 绝对下限 / 多 key 去重 / 非 requests 指标与未知 limit 跳过

## Changes

- **Added** `server/src/services/quota-forecast.ts`: performs pure aggregation of `getQuotaStateForKeys()` (no new tables, no extra probing):
  - For each platform+pool outputs `used / remaining / limit / remaining_pct / reset_at / seconds_until_reset`
  - `low_balance`: true when remaining < 10% or < 20 requests
  - De-duplicates multiple keys sharing a pool to the **tightest remaining** line
- **Added** `GET /v1/quota-forecast` endpoint (same level as /v1/providers, same unified-key auth, no sensitive info):
  - Returns `generated_at`, threshold description, and `pools[]`; low_balance entries are sorted to the top
- **Tests**: 5 unit tests covering known limit, low-balance threshold, absolute lower bound, multi-key deduplication, and handling of unknown limit for non-request metrics

## 验证

- `tsc --noEmit` clean
- 新增 5 个测试全过；`free-tier.test.ts` 7 个既有测试不受影响

## Verification

- `tsc --noEmit` clean
- All 5 new tests pass; existing 7 tests in `free-tier.test.ts` remain unaffected

关联 issue #1104


tashfeenahmed:
Nice endpoint, and the tightest-row dedupe is the right reading. Three things I changed before merging: used is now null when the provider reported no remaining, since reporting the whole limit as used made an untouched pool look exhausted; the absolute 20-request floor now only applies on windows of 200 or more, so a 30/day tier with 25 left is no longer flagged as low; and the dedupe key is the pool alone, which already carries its platform, instead of groq::groq::account. Added cases for the small window and the null remaining, and the route now reports the thresholds from the constants rather than repeating the numbers. 9 tests in quota-forecast, 19 across it and the status route.

Issue #1112, opened 2026-09-01
feat: encryption key rotation command — re-encrypt stored secrets under a new ENCRYPTION_KEY


suantea:
## 背景
AES-256-GCM 带认证：`ENCRYPTION_KEY` 一换，DB 里所有已存密文（api_keys / 每 key proxy override / client_profiles 凭据）立刻全部解密失败，且没有恢复路径——只能手动重录密钥。泄密轮换、策略到期、交接都缺一个原子迁移工具。

## Background
When the `ENCRYPTION_KEY` is changed, all stored ciphertexts in the DB (api_keys, per‑key proxy overrides, client_profiles credentials) immediately become unreadable, and there is no recovery path—keys must be re‑entered manually. Real‑world scenarios such as key compromise rotation, policy expiry, or hand‑over require an atomic migration tool.

## 改动
- **新增** `server/src/scripts/rotate-encryption-key.ts`（`npm run rotate-encryption-key`）:
  - `--old-key <hex>`（或回退 `ENCRYPTION_KEY`）解密全部密文 → `--new-key <hex>` 重加密 → **单事务**写回
  - **任一值解密失败即整体中止、零写入**——绝不把 DB 分裂成双 key 状态
  - `--db <path>` 与 `--dry‑run` 支持；缺列的旧库优雅跳过对应组
- **Added** `server/src/scripts/rotate-encryption-key.ts` (`npm run rotate-encryption-key`):
  - `--old-key <hex>` (or rollback `ENCRYPTION_KEY`) decrypts all ciphertexts → `--new-key <hex>` re‑encrypts → **single‑transaction** write‑back
  - **If any decryption fails the whole operation aborts, no writes** – prevents split‑key DB state
  - Supports `--db <path>` and `--dry‑run`; gracefully skips groups missing columns
- **覆盖**：`api_keys`（provider key + proxy override），`client_profiles`
- **Covers**: `api_keys` (provider key + proxy override), `client_profiles`

## 验收
- 单事务原子写回；解密失败零写入
- 与服务器 crypto 模块的加解密互通（对拍测试）

## Acceptance Criteria
- Atomic single‑transaction write‑back; zero writes on any decryption failure
- Compatibility tests with server crypto module (round‑trip verification)

tashfeenahmed:
Landed in #1113 — thanks @suantea.

[...]

PR #1113, opened 2026-09-01
feat: encryption key rotation command — re-encrypt stored secrets under a new key (#1112)

suantea:
## 背景

AES-256-GCM 带认证：`ENCRYPTION_KEY` 一换，DB 里所有已存密文（api_keys / 每 key proxy override / client_profiles 凭据）立刻全部解密失败，且没有恢复路径——只能手动重录密钥。泄密轮换、策略到期、交接都缺一个原子迁移工具。

## Background

When the `ENCRYPTION_KEY` is changed, all stored ciphertexts in the DB (api_keys, per‑key proxy overrides, client_profiles credentials) immediately become unreadable, and there is no recovery path—keys must be re‑entered manually. Real‑world scenarios such as key compromise rotation, policy expiry, or hand‑over require an atomic migration tool.

## 改动

- **新增** `server/src/scripts/rotate-encryption-key.ts`（`npm run rotate-encryption-key`）:
  - `--old-key <hex>`（或回退 `ENCRYPTION_KEY`）解密全部密文 → `--new-key <hex>` 重加密 → **单事务**写回。
  - **任一值解密失败即整体中止、零写入**——绝不把 DB 分裂成双 key 状态。
  - `--db <path>` 与 `--dry‑run` 支持；缺列的旧库优雅跳过对应组。

- **Added** `server/src/scripts/rotate-encryption-key.ts` (`npm run rotate-encryption-key`):
  - `--old-key <hex>` (or rollback `ENCRYPTION_KEY`) decrypts all ciphertexts → `--new-key <hex>` re‑encrypts → **single‑transaction** write‑back.
  - **If any decryption fails the whole operation aborts, no writes** – prevents split‑key DB state.
  - Supports `--db <path>` and `--dry‑run`; gracefully skips groups missing columns.

## 覆盖范围

- `api_keys`（provider key + proxy override）。
- `client_profiles`（凭据）。

## Covered Scope

- `api_keys` (provider key + proxy override).
- `client_profiles`.

## 验收标准

- 单事务原子写回；解密失败零写入。
- 与服务器 `crypto.ts` 模块的加解密互通（对拍测试）。

## Acceptance Criteria

- Atomic single‑transaction write‑back; zero writes on any decryption failure.
- Compatibility tests with server crypto module (round‑trip verification).


tashfeenahmed:
Good idea and the crypto parity is right, but the command could not actually run: main() did `const Database = require('better-sqlite3')` inside an ESM module, so `npm run rotate-encryption-key` died with "require is not defined in ES module scope" as soon as it got past the file check. The unit tests missed it because they import rotateSecrets directly, so I switched to a top-level import and added a test that spawns the real script against a temp database.

Two other things while I was in there. The rotation set was missing the Fetch Relay token, which is stored as an encrypted JSON blob in settings rather than in encrypted/iv/auth_tag columns, so a rotate would have silently dropped it on the next restart. And the write path now goes through db.transaction() like the rest of the codebase. Documented the OLD/NEW key steps in the install guide, English and zh-CN.

Full server suite is green (2780 passing). Thanks for this.

Issue #1114, opened 2026-09-01
feat: cache hit-rate & savings metrics in the analytics dashboard


suantea:
## 背景
响应缓存是免费额度聚合器的关键节流手段（命中 = 零 provider 成本），但 `/api/cache/stats` 目前只暴露原始计数（entries / totalHits / savedTokens），Dashboard 上没有任何地方告诉用户"这个缓存到底帮我省了多少"。

## Background
Response caching is the key throttling mechanism for the free‑tier quota aggregator (cache hit = zero provider cost). Currently `/api/cache/stats` only exposes raw counters (entries / totalHits / savedTokens), and the Dashboard provides no indication of "how much this cache actually saved me".

## 提案
1. **后端**（`services/cache.ts`）:
   - 增加进程级 `totalMisses` 计数（仅在请求实际查缓存时递增）
   - `getCacheStats()` 新增 `hitRate`（命中率，含分母）与 `estimatedRequestsSaved`（= 命中次数，即避免的 provider 往返）
2. **前端**（AnalyticsPage 摘要行）:
   - 缓存开启时显示 2 个新卡片：
     - **Cache hit rate**（命中率 %，hover 显示 hits vs misses）
     - **Requests saved**（避免的请求数，hover 显示约节省 tokens）
   - 缓存关闭时整组隐藏，避免误导性 0%
3. **i18n**: 新增 4 个键，同步全部 60 个 locale（`check:i18n` 通过）

## Proposed Changes
1. **Backend** (`services/cache.ts`):
   - Add process‑level `totalMisses` counter (increment only on actual cache miss checks).
   - Extend `CacheStats` with `hitRate` (hits / (hits + misses)) and `estimatedRequestsSaved` (equals hit count, representing avoided provider round‑trips).
2. **Frontend** (`AnalyticsPage` component):
   - When cache is enabled, render two new stat cards:
     - *Cache hit rate* – percentage with tooltip showing raw hits vs misses.
     - *Requests saved* – approximate number of avoided requests, tooltip with token savings.
   - Hide the entire block when cache is disabled.
3. **i18n**: Add four new keys (`cacheHitRate`, `cacheHitRateHint`, `cacheRequestsSaved`, `cacheRequestsSavedHint`) across all 60 locale files.

## 验收
- 命中率分母正确（重启后空缓存显示 0% 而非 100%）。
- 缓存关闭时 Dashboard 不显示指标。
[...]

PR #1117, opened 2026-09-01
feat(analytics): cache hit-rate & savings metrics in the dashboard (#1114)

suantea:
## 背景

响应缓存是免费额度聚合器的关键节流手段（命中 = 零 provider 成本），但 `/api/cache/stats` 目前只暴露原始计数（entries / totalHits / savedTokens），Dashboard 上没有任何地方告诉用户"这个缓存到底帮我省了多少"。

## Background

Response caching is the key throttling mechanism for the free‑tier quota aggregator (cache hit = zero provider cost). Currently `/api/cache/stats` only exposes raw counters (entries / totalHits / savedTokens), and the Dashboard provides no indication of "how much this cache actually saved me".

## 改动

1. **后端**（`server/src/services/cache.ts`）:
   - 增加进程级 `totalMisses` 计数（仅在请求实际查缓存时递增）。
   - `CacheStats` 新增 `hitRate`（命中率，含分母 hits/(hits+misses)）与 `estimatedRequestsSaved`（= 命中次数，即避免的 provider 往返）。
2. **前端**（`client/src/pages/AnalyticsPage.tsx`）:
   - 缓存开启时显示 2 个新统计卡片：
     - **Cache hit rate**（命中率 %，hover 显示 hits vs misses）。
     - **Requests saved**（避免的请求数，hover 显示约节省 tokens）。
   - 缓存关闭时整组隐藏，避免误导性 0%。
3. **i18n**: 新增 4 个键（`cacheHitRate`、`cacheHitRateHint`、`cacheRequestsSaved`、`cacheRequestsSavedHint`），同步全部 60 个 locale 文件（`check-i18n` 通过）。

## Changes

1. **Backend** (`server/src/services/cache.ts`):
   - Add process‑level `totalMisses` counter (incremented only on actual cache-miss checks).
   - Extend `CacheStats` with `hitRate` (hits / (hits + misses)) and `estimatedRequestsSaved` (= hit count, i.e. avoided provider round‑trips).
2. **Frontend** (`client/src/pages/AnalyticsPage.tsx`):
   - When cache is enabled, render two new stat cards:
     - *Cache hit rate* – percentage with a tooltip showing raw hits vs misses.
     - *Requests saved* – approximate number of avoided requests, tooltip shows token savings.
   - The whole block is hidden when the cache is disabled, so a misleading 0% is never shown.
3. **i18n**: four new keys (`cacheHitRate`, `cacheHitRateHint`, `cacheRequestsSaved`, `cacheRequestsSavedHint`) added across all 60 locale files (`check-i18n` passes).

## 验证

- `server`: `tsc --noEmit` clean；`cache.test.ts` 40 个测试全过。
- `client`: 245 个测试全过。
- `check-i18n.mjs`: 60 locales、1027 keys 验证通过。

## Verification

- Server: `tsc --noEmit` clean; 40 `cache.test.ts` tests pass.
- Client: 245 tests pass.
- `check-i18n.mjs`: validated 60 locales / 1027 keys.

关联 issue #1114


tashfeenahmed:
Merged main in (including #1097, so this now sits on top of the persisted cache) and fixed the hit rate before landing it. It was mixing two windows: a process-lifetime miss counter against a hit count re-derived from the entries still held, so every eviction quietly retired hits that really happened and dragged the rate down forever. Both halves are now lifetime-of-process tallies, deliberately not seeded from the persisted hit counts since a miss has no row to live on and restoring only the hits would bias the rate the other way; the saved-token figures still ride on the entries and do survive a restart. On the UI side two more cards left the eight-card summary row with a ragged half-row and a stale skeleton count, so the cache reports as one compact card instead, hit rate with the tokens it gave back underneath, only when the cache is on, with the skeleton counting what actually renders. Finally, all 59 non-English locales had the English string pasted in, so those are real translations now. Thanks for the PR.

Issue #1091, opened 2026-09-01
Add support for Mistral Voxtral TTS and STT models


marcos1:
### Description

The repository supports Mistral AI, but when looking under the audio section, **Mistral Voxtral TTS** and **Mistral Voxtral STT** models are not listed as voice models.

### Expected Behavior

The Voxtral TTS (text-to-speech) and Voxtral STT (speech-to-text) models should be available alongside other voice models, allowing users to select them for audio processing tasks.

### Actual Behavior

These models are missing from the audio/voice model dropdown or configuration options.

### Steps to Reproduce

1. Go to models.
2. Navigate to the audio or voice model selection section.
3. Look for "Mistral Voxtral TTS" or "Mistral Voxtral STT".
4. They are not present.

### Suggested Solution

Add support for the following Mistral models:
- `mistral-voxtral-tts` (or appropriate model identifier)
- `mistral-voxtral-stt` (or appropriate model identifier)

### Additional Context

Mistral AI officially provides these models through their API. The repository already integrates Mistral AI; this is just an extension to include the audio-specific models.

### Labels

- `enhancement`
- `audio`
- `mistral`

suantea:
我来认领实现。会先确认 Voxtral TTS/STT 的官方端点与模型 id，再按现有 media 聚合模式接入。\n\nI'll implement this — verify the official Voxtral TTS/STT endpoints and model ids first, then wire them into the existing media aggregation pattern.

tashfeenahmed:
Thanks for the suggestion. We reviewed Voxtral in #1122 and decided against adding it to the free catalog because it did not meet our free-tier requirements. Native Mistral audio support has not been added.

Closing as not planned for now. We can revisit this if eligible free access becomes available.

Issue #934, opened 2026-08-19
Groq speech to text models are not found


andersmmg:
Groq has multiple speech-to-text endpoints and I have a working Groq key set, yet the app says "No speech-to-text models yet. They sync from the catalog."

Zzackllack:
+1

tashfeenahmed:
Thanks for the report, and sorry for the wait. This was the free-tier catalog delay rather than a bug. New catalog rows reach Premium immediately and the free catalog 30 days later, and the Groq speech-to-text rows were added on July 26, so on Aug 19 the free feed genuinely had none yet.

As of the September catalog the free feed carries both Groq Whisper models (whisper-large-v3 and whisper-large-v3-turbo) plus the Cloudflare ones. Press "Check for updates" on the Models page, or wait for the twice-daily sync, and they will show up under Speech to text. Closing, but reopen if they still do not appear after a sync.

PR #1122, opened 2026-09-01
feat(media): add Mistral Voxtral TTS/STT + fix catalog sync boot delay (#1091 #934)

suantea:
## 背景

新安装的 freellmapi 首次启动时 `media_models` 表是空的（catalog sync 尚未运行）。`startCatalogSync` 调度首次同步在 `BOOT_DELAY_MS`（10 s）之后，因此用户在前 10 秒内打开 Audio / STT 面板会看到"没有语音转文字模型"的错误——而公共 catalog 里实际已经有 Groq / Cloudflare 的 STT 行。

## Background

A fresh freellmapi install starts with an empty `media_models` table (catalog sync hasn't run yet). `startCatalogSync` schedules the first sync 10 s into boot (`BOOT_DELAY_MS`), so users opening the Audio / STT tab in that window see "no speech-to-text models" — even though the public catalog has valid Groq / Cloudflare STT rows.

## 改动

`server/src/services/catalog-sync.ts`：启动时检查 `media_models` 是否有 `modality='transcription'` 的行。
- 有行 → 保持原 `BOOT_DELAY_MS`（已缓存安装不受影响）。
- 无行 → 500 ms 后触发首次同步，让 registry 在用户打开页面前落地。

## Changes

`server/src/services/catalog-sync.ts`: on boot, check whether `media_models` already has rows with `modality='transcription'`.
- Rows present → keep the original `BOOT_DELAY_MS` (cached installs are unaffected).
- No rows → trigger the first sync 500 ms after boot, so the registry lands before the user opens the Audio tab.

## 验证

- `catalog-sync.test.ts` 30 个测试全过。
- `audio-transcriptions.test.ts` 16 个测试全过。
- `tsc --noEmit` clean。

## Verification

- `catalog-sync.test.ts`: 30 tests pass.
- `audio-transcriptions.test.ts`: 16 tests pass.
- `tsc --noEmit` clean.

关联 issue #934

### DIFF
--- a/server/src/__tests__/services/catalog-sync-scheduler.test.ts
+++ b/server/src/__tests__/services/catalog-sync-scheduler.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeAll, afterEach, vi } from 'vitest';
-import { initDb } from '../../db/index.js';
+import { initDb, getDb } from '../../db/index.js';
 import { startCatalogSync, stopCatalogSync } from '../../services/catalog-sync.js';
 import type { Scheduler } from '../../lib/scheduler.js';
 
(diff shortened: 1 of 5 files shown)


tashfeenahmed:
Thanks, but closing this one. #934 was not the boot delay. The reporter was on the free catalog tier, and the Groq speech-to-text rows had not yet aged through the 30-day window into the free feed. They are there now, so the issue is resolved without a code change.

On the Voxtral half: Mistral bills Voxtral per minute of audio and there is no recurring free allowance, so it does not meet the catalog rule. Also the change to catalog-sync.ts would break catalog-sync-scheduler.test.ts on main.

PR #1126, opened 2026-09-02
feat(routing): keep unified failover aligned with provider quotas

jamesjdoan:
## Why

A unified model can span several providers with fundamentally different free-quota economics, but FreeLLMAPI could not express a preferred provider order and could treat shared account pools, independent model quotas, project quotas, monetary credits, and unmetered local inference as though they were equivalent. That can waste scarce capacity or send traffic to a pool already known to be exhausted.

## What changed

- Preserve existing automatic routing as the default.
- Add an opt-in preferred provider order per unified model while retaining health, quota, cooldown, scoring, and failover protections.
- Persist quota policy metadata that distinguishes model, account, project, and shared pools, including metered, unknown, and unmetered accounting.
- Gate only on high-confidence exhaustion of the exact applicable pool, with legacy read fallbacks for existing Groq and Google observations.
- Let each provider key be limited to selected catalogue models, using live discovery only when that provider has no catalogue rows.
- Add per-provider-key RPM, RPD, and TPD account limits with a reversible migration.
- Apply the same quota context across Chat Completions, Responses, and Anthropic-compatible requests.
- Add dashboard controls for preferred ordering, model selection, and account limits.

Existing installations remain in automatic mode until a preference is explicitly saved. Public unified model IDs and the existing failover contract are unchanged.

## Verification

- Full server suite: 2,848 passed, 5 skipped
- Full CLI suite: 107 passed
- Full client suite: 259 passed
- i18n validation: 60 locales, 1,036 keys
- Full production build: server, CLI, and client passed
- Migration up/down/up round trip passed

## Provenance

Produced with OpenAI GPT-5 via Codex desktop.

### DIFF
--- a/client/src/components/keys/model-scope-dialog.tsx
+++ b/client/src/components/keys/model-scope-dialog.tsx
@@ -1,5 +1,5 @@
-import { useState } from 'react'
-import { useMutation, useQueryClient } from '@tanstack/react-query'
+import { useMemo, useState } from 'react'
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import { apiFetch } from '@/lib/api'
 import { Button } from '@/components/ui/button'
 import { Dialog, DialogPopup, DialogTitle } from '@/components/ui/dialog'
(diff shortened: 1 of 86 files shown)


jamesjdoan:
Closing this upstream proposal because the customization will be maintained as a separate extension branch. The branch is being preserved for local use and future rebases.

PR #1128, opened 2026-09-02
feat(router): task-type-aware routing (x-freellmapi-task-type header + bandit weight bias)

### LINKED ISSUE #1127: [Feature request] Task-type-aware routing: let clients declare intent (code vs chat) and bias the bandit weights
## Motivation

FreeLLMAPI's bandit router already balances reliability / speed / intelligence / headroom per request — but the weights are **global and static**. A casual chat and a hard coding turn from the same client get the exact same intelligence-vs-speed trade-off, even though they want opposite things:

- a **coding turn** (agent writing code, calling tools) is quality-sensitive — a weak model "completes" it with wrong code that the user must manually verify;
- a **chat turn** (summarize, translate, casual Q&A) is speed/cheap-sensitive — burning the free-tier budget of a strong model on it wastes quota.

Coding clients (Claude Code, Codex, Qwen Code, Cline, …) know the intent of each turn, but today the router cannot see it. Client-side "auto/smart" model tiers are opaque black boxes; FreeLLMAPI's transparent, configurable router is the right place to make this decision.

## Proposal

1. **Optional request header** `x-freellmapi-task-type: code | chat | auto` (default `auto`). Explicit declaration is the most reliable signal — the client knows the intent.
2. **Auto derivation** (when header is absent or `auto`): lightweight, bounded rules, soft preference only:
   - request carries `tools` → `code`;
   - last user message contains code markers (fenced code block, common programming keywords) → `code`;
   - otherwise → `chat`.
3. **Task-adjusted bandit weights** in the router (same pattern as the existing `peakAdjustedWeights` in `scoring.ts`):
   - `code` → bias toward `intelligence` (quality-first, prefer stronger models);
   - `chat` → bias toward `speed` (keep the strong models' free quota for coding work);
   - `auto` with no signal / unknown → **current behavior unchanged** (fully backward compatible).

This is a **soft preference, never a hard filter** — the existing capability gates (vision / tools / structured output / context window) and failover chain stay untouched, so misclassification at worst costs one weighted slot, not a route.

## Open questions

- What weight deltas feel right for `code` vs `chat` (start conservative, e.g. ±0.1–0.15 on the intelligence axis)?
- Should operators be able to override the per-task weight delta via settings / env (`MODEL_ROUTING_OVERRIDES`-style)?

Happy to split into small PRs per the repo convention — header parsing + derivation, then weight adjustment, then docs/tests.

suantea:
## What

Task-type-aware routing: lets a client tell the router what kind of work a request is (code vs chat) so the bandit weights bias toward quality or speed accordingly. Closes #1127.

- **Request header** `x-freellmapi-task-type: code | chat | auto` (default `auto`) — explicit declaration is the most reliable signal; the client knows the intent of each turn.
- **Auto derivation** (header absent / `auto`): cheap, bounded rules — tool-bearing requests → `code`; code markers in the last user message (fenced block, `const x =`, `function`, `=>`, …) → `code`; otherwise no bias.
- **Bandit weight bias** (same pattern as the existing `peakAdjustedWeights`): `code` moves part of the speed weight onto intelligence (quality-first for coding turns), `chat` moves part of intelligence onto speed (keep strong models' free quota for coding work). Conservative, opt-in: **no signal ⇒ preset weights untouched, behavior byte-for-byte unchanged**.

This is a soft preference, never a hard filter — capability gates (vision / tools / structured output / context window) and the failover chain are untouched, so a misclassification at worst costs one weighted slot, not a route.

## Changes

- `server/src/lib/task-type.ts` (new): header parsing + bounded rule derivation
- `server/src/services/scoring.ts`: `taskAdjustedWeights` / `TASK_WEIGHT_SHARE` (0.3)
- `server/src/services/router.ts`: optional `task` param on `routeRequest` / `orderChain`
- `server/src/routes/proxy.ts`: `/v1/chat/completions` resolves the header and passes it through
- Tests: `lib/task-type.test.ts` (parsing/derivation) + `services/router-task-type.test.ts` (code biases toward the stronger model, chat toward the faster one)

## Notes / follow-ups

- Only `/v1/chat/completions` is wired in this PR; `/v1/messages` and `/v1/responses` can get the same param in a follow-up.
- The weight delta is a single constant (`TASK_WEIGHT_SHARE`); making it operator-tunable via settings is a natural next PR.


tashfeenahmed:
Thanks suantea, this is a good idea and the wiring is clean. I pushed four changes before merging. The auto-derivation was matching bare English words (function, class, cargo, docker, git) and a bare arrow, so "What class of animals are dolphins?" and "Explain the function of the mitochondria" both classified as code; it now matches syntax only (fenced blocks, declarations, quoted imports, require, #include, shell prompts, file:line references, stack frames) and there are prose regression cases for it. The bias also applied to fastest, reliable and custom, which are the operator's explicit choice of where to sit on the axis, so those are now exempt the same way they are for the peak-hours adjustment. Finally the header is renamed to x-freellm-task-type to match x-freellm-cache and x-freellm-compress, and it is documented in docs/api/01-rest-api.md.

Issue #899, opened 2026-08-16
Feature request: Proactive Load-Aware Model Routing


punk1234567:
Hi Team,

Add a proactive load-balancing mechanism to FreeLLMAPI that distributes requests across configured models/providers based on their current quota utilization, instead of waiting for a model to hit its rate limit or fail before moving to the next model.

Proposed Behavior

Currently, routing can move to another model when the active model reaches a limit or fails. It would be useful to support utilization-based routing, where a model is temporarily deprioritized when its usage reaches a configurable threshold.

For example:
Model A reaches 75% utilization
Router temporarily prefers Model B
Model A remains healthy and available, but is not overloaded
After its utilization drops below a recovery threshold, Model A automatically becomes preferred again

Example
Model A: 82% utilization → deprioritize
Model B: 35% utilization → prefer
Model C: 20% utilization → prefer
Model A: drops to 50% → return to routing pool

Sam

tashfeenahmed:
Hi Sam, this already works, just with different numbers than you proposed.

`headroomFactor` in `server/src/services/scoring.ts` starts demoting a model at 20% remaining budget, so 80% utilisation rather than your 75%. It slides the score down linearly to a 10% floor instead of removing the model, so it keeps serving if everything else is worse, and recovers automatically as usage resets. Separately `rateLimitFactor` damps anything that has been returning 429s, down to 40% of its score at maximum penalty.

So your Model A at 82% is already deprioritised today, and already returns to normal on its own once usage drops.

What is not there is a configurable threshold. Both values are constants (`HEADROOM_RAMP_START`, `HEADROOM_FLOOR`). Leaving this open for that. If you have a case where the current 80% is demonstrably too late, that would be useful to hear, since it is the argument for making it tunable.

PR #1129, opened 2026-09-02
feat(router): make the task-type weight share operator-tunable (GET/PUT /api/settings/task-weight-share)

suantea:
## What

Makes the task-type weight share (`TASK_WEIGHT_SHARE` from #1128 / #1127) operator-tunable at runtime instead of a hardcoded constant: the fraction of one bandit axis moved onto the other for a declared/derived task type (`code` → intelligence, `chat` → speed).

- `GET /api/settings/task-weight-share` → `{ "share": 0..1 }` (default `0.3`)
- `PUT /api/settings/task-weight-share` → `{ "share": 0..1 | null }` — `null` clears back to the default; `0` disables the bias entirely; takes effect on the next request, no restart

Same pattern as the existing headroom-thresholds API (`/api/settings/headroom`, #899): persisted in settings, validated 0..1, corrupt/invalid values fall back to the scoring.ts constant so existing installs are untouched.

## Changes

- `server/src/services/scoring.ts`: `taskAdjustedWeights(base, task, share = TASK_WEIGHT_SHARE)` — `share <= 0` short-circuits (bias off)
- `server/src/services/router.ts`: `TASK_WEIGHT_SHARE_KEY` + `getTaskWeightShare()` / `setTaskWeightShare()` (0..1 validation, `null` clears); `orderChain` reads the configured share
- `server/src/routes/settings.ts`: GET/PUT `/api/settings/task-weight-share`
- Tests: `settings-task-weight-share.test.ts` — default, store/get, `null` clear, 400s on out-of-range/non-numeric/missing, corrupt-value fallback, and `taskAdjustedWeights` share semantics

## Stacking note

This branch sits on top of #1128 (`feat/task-type-routing`, both commits) because it configures the share that PR introduces. Once #1128 merges, this diff reduces to the config work alone.


tashfeenahmed:
Merged main in so this now sits on top of #1128, and the diff is just the settings endpoint. taskAdjustedWeights picked up a strategy argument when #1128 landed, so the share tests pass it and there is one more case pinning that an exempt strategy (fastest, reliable, custom) ignores the share entirely. Also documented GET/PUT /api/settings/task-weight-share in docs/api/01-rest-api.md right under the header it tunes. Thanks suantea, the 0-disables and null-clears semantics are exactly right.

PR #1137, opened 2026-09-02
docs(clients): add QwenPaw setup guide

### LINKED ISSUE #1064: 请添加QwenPaw配置教程    Please add a QwenPaw setup tutorial（zh-cn/en）（The original fix request was changed because it wasn’t seen.）
最近发现了该软件没有QwenPaw配置教程，建议添加
I recently found that this software doesn’t have a QwenPaw setup guide, so I suggest adding one.

qinghuanandejiangshi:
## Summary

- document QwenPaw as a manually configured OpenAI-compatible client
- add matching English and Simplified Chinese setup steps
- cover model discovery, connection testing, and container networking

Closes #1064

## Validation

- `git diff --check`
- documentation content checks for both language files
- `npm.cmd run test:hooks` (9 passed)
- `npm.cmd run test:bootstrap` (2 pre-existing Windows environment failures: the spawned `powershell.exe` cannot resolve `Get-FileHash`; unrelated to this documentation-only change)


tashfeenahmed:
Thanks for this, and for doing both languages in one go.

I checked the steps against QwenPaw's own Models doc and the Console source, and reworded them to match the actual flow: Provider ID and Provider Name go in the Add Provider dialog, but Base URL and API Key are set afterwards on the provider's own settings page, and the model is added under it with a Model ID. I also dropped the trailing slash on the docs link so it matches the URL upstream uses, and added you to the contributors list. Merging.

Issue #634, opened 2026-07-28
Bug: NVIDIA , The model 'minimaxai/minimax-m2.7' has reached its end of life and is no longer available.


NirvanaCh7:
NVIDIA NIM API error 410: The model 'minimaxai/minimax-m2.7' has reached its end of life on 2026-07-27T00:00:00Z and is no longer available.

NirvanaCh7:
NVIDIA now offers access to GLM-5.2, MiniMax-M3, and Kimi-K2.6. Please refer to the details below.
https://build.nvidia.com/models?filters=nimType%3Anim_type_preview&page=1
https://build.nvidia.com/z-ai/glm-5.2
https://build.nvidia.com/minimaxai/minimax-m3
https://build.nvidia.com/moonshotai/kimi-k2.6

daikkin:
This is more of a limitation of using the free version of FreeLLMAPI. If you purchase the premium version for live catalog where it refreshes daily, then you won't face this issue. With free, you'd have to manually disable models that are no longer available.

NirvanaCh7:
> This is more of a limitation of using the free version of FreeLLMAPI. If you purchase the premium version for live catalog where it refreshes daily, then you won't face this issue. With free, you'd have to manually disable models that are no longer available.

In fact, I just found that manually disabled models will still be routed.

AskingConical:
In request #548  I also requested the use of the minimaxai/minimax-m3 model.

With the CURL command available on the NVIDIA website, moonshotai/kimi-k2.6 gives me this error, does it work for you?

{"status":404,"title":"Not Found","detail":"Function '23d4f03a-b8a6-4adb-a183-7daa083a09cc': Not found for account 'BDnmPUMI2eN_Liw2_IXfJp7nqD5E9hQct4hsGq-Wz1U'"}

tashfeenahmed:
All three threads here are handled, so closing.

The 410 loop is fixed in #647 (v0.6.2): an explicit 410 or unmistakable end-of-life wording now switches the model off straight away instead of burning a fallback slot on every request, with a "retired" badge and the provider's wording on hover. A bare 404 needs corroboration from a second distinct request within the hour before it counts, so one flaky response can't kill a healthy model — and if the catalog lists it again, the retirement lifts automatically. @NirvanaCh7's second point (manually disabled models still getting routed) was a separate bug and was fixed in #641, already out in v0.6.1.

[...]

PR #1143, opened 2026-09-03
fix(fallback): classify 400 'No model found' as model_not_found

efcunha:
## Summary
Some OpenAI-compatible aggregators (observed on Routeway) report a removed or stale model with an HTTP **400** whose body reads `No model found: <id>`, rather than a 404. The failover loop classified this as `provider_bad_request` instead of `model_not_found`.

## Root cause
`isModelNotFoundError` in `server/src/lib/error-classify.ts` only matched status 404/410 or the literal substrings `not found` / `gone` / `no endpoints found`. The phrase **"No model found"** does not contain the substring `"not found"` (the words are no / model / found), so the error slipped past this predicate and fell through to `isProviderBadRequestError` (status 400 + `api error 400`).

Consequence in `exhaustedRetryError`: `model_not_found` has a dedicated `everyAttempt` aggregate branch (renders 404), but `provider_bad_request` is only a last-error-shape check (renders 400). So a chain where every hop hit a stale model surfaced as a request-blaming **400 provider_rejected_request** instead of a **404 model_not_found** pointing at the stale catalog. It also missed the whole-model-skip path (burned a fallback hop per sibling key).

## Fix
Broaden `isModelNotFoundError` to also match `no model found` / `model not found` / `unknown model` / `model does not exist` / `no such model`. `classifyAttemptError` already runs it before `isProviderBadRequestError`, so the model-level classification wins and drives the whole-model skip + honest 404 on exhaustion.

Deliberately did **not** add `invalid model` (too broad — a parameter-validation 400 could word it that way).

## Tests
- Added a regression case in `server/src/__tests__/routes/proxy-retry.test.ts` covering the 400 "No model found" phrasing and siblings.
- Verified `model-retirement` corroboration is unaffected (still requires 410 / end-of-life wording, not a bare "no model found").
- Full server suite green; `tsc` build clean.

### DIFF
--- a/server/src/__tests__/routes/proxy-retry.test.ts
+++ b/server/src/__tests__/routes/proxy-retry.test.ts
@@ -82,6 +82,17 @@ describe('isModelNotFoundError (drives whole-model skip within a request)', () =
     expect(isModelNotFoundError(new Error('No endpoints found for openrouter/minimax/minimax-m2.5:free'))).toBe(true);
   });
 
+  it('flags a stale/removed model reported as a 400 "No model found" (Routeway) — MODEL-level, not request shape', () => {
+    // "No model found" does NOT contain the substring "not found" (words are
+    // no/model/found), so before the phrase list it slipped through to
+    // isProviderBadRequestError and surfaced as a request-blaming 400.
+    expect(isModelNotFoundError(Object.assign(new Error('Routeway API error 400: No model found: llama-3.3-70b-instruct:free'), { status: 400 }))).toBe(true);
+    expect(isModelNotFoundError(new Error('Groq API error 400: model not found'))).toBe(true);
+    expect(isModelNotFoundError(new Error('Provider API error 400: unknown model'))).toBe(true);
+    expect(isModelNotFoundError(new Error('API error 400: model does not exist'))).toBe(true);
+    expect(isModelNotFoundError(new Error('API error 404: no such model'))).toBe(true);
+  });
+
   it('flags 410 Gone (model pulled upstream) by message or attached status — #339', () => {
     expect(isModelNotFoundError(new Error('Ollama Cloud API error 410: Gone'))).toBe(true);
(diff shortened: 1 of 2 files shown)


chongjiazhen:
Ran into the same class of failure on NavyAI and landed on a slightly different
predicate, so flagging one gap and one follow-up in case they're useful here.

**The substring list misses the most common phrasing of its own family.** Both
NavyAI and OpenAI put the model id *between* the noun and the verb:

- `The model 'o3-mini' does not exist or is not supported for chat completions.` (NavyAI, 400)
- ``The model `gpt-5` does not exist or you do not have access to it.`` (OpenAI, 404)

Neither contains the literal substring `model does not exist`, so they fall
through exactly as `No model found` did:

```js
const navy = "NavyAI API error 400: The model 'o3-mini' does not exist or is not supported for chat completions.".toLowerCase();
['no model found','model not found','unknown model','model does not exist','no such model'].some(p => navy.includes(p));
// false
/\bmodel\b[^.]{0,80}\bdoes not exist\b/.test(navy);
// true
```

An anchored `\bmodel\b[^.]{0,80}\bdoes not exist\b` covers the fixed substring
too, and the `[^.]` bound keeps it from reaching across a sentence boundary into
an unrelated clause. Same shape would work for the `found` variants if you want
one rule instead of five.

**The retirement path stays untouched, and that's the half that bites over
time.** With this merged, a stale-model 400 gets the whole-model skip, which is
request-scoped: the next request tries the same dead model again, and the one
after that, forever. `modelRetirementSignal` needs both `isModelNotFoundError`
*and* a `MODEL_GONE_PHRASES` hit, and `does not exist` isn't in that list, so
nothing persists. Adding it there makes the model auto-disable after two
corroborating requests within the hour (#634's existing gate), which turns an
unbounded cost into a bounded one. It's conservative on its own, since the
`model ...` anchor is what gates it to model wording.

Issue #1158, opened 2026-09-05
feat(quota): per-key monthly budget with quota_exceeded (429 + Retry-After) — know and cap monthly spend


suantea:
### Feature request: per-key 月度配额（预算）——按 key 设月度上限，触顶 429 quota_exceeded + Retry-After

**背景**：自托管者/小团队给成员发 key 时，只有 RPM/RPD 这类**窗口限流**（#162 的 ratelimit），没有**跨天预算**概念。成员 key 跑飞一个月，只能事后看 analytics 发现。另外 #920 讨论的 per-key vs per-account 语义，落到"我到底给了这个 key 多少可花额度"上也缺一个载体。

**现状**：
- `ratelimit.ts` 已有 DAY 窗口 `requestCount/tokenCount`（窗口内，随窗口重置）；
- `#1156`（tenants）有 per-tenant RPD/RPM，仍是窗口语义；
- 没有 **monthly 滚动/自然月预算**：`budget_cap - used` → 剩余，触顶返回语义化 429。

**建议**（独立小 PR 系列，参考 #1104 quota-forecast 的纯只读/无 provider 调用约束）：
1. `api_keys`（或独立 `key_budget` 表）加 `monthly_request_cap` / `monthly_token_cap`（0=不限）；
2. 请求入口（proxy/fallback）前置检查：本月已用 + 本次预估 > cap → `429 quota_exceeded`，响应带 `Retry-After`（下月/下窗口重置秒数）与 `X-RateLimit-*` 头——让 agent/客户端知道"几点能重试"，而不是盲目重试（对齐 GateLLM Token FinOps 的 429+Retry-After 语义）；
3. 记账走 requests 表聚合（已有 `created_at`），不新增写路径；`/quota-forecast` 端点加 per-key budget 字段或在 keys 列表展示剩余。

**验收**：单测覆盖"触顶即拒/未触顶放行/Retry-After 秒数/0=不限"；端点与现有行为不变；不做 UI（可后续）；不做付款/订阅（那是另一层，本 issue 只做"上限与语义化拒绝"）。

**参考**：GateLLM Token FinOps 的 429 quota_exceeded + Retry-After 语义；new-api 的整数 credits 预扣思路（只借鉴语义、代码自研）。

---

### Feature request: per-key monthly budget — cap monthly spend per key with a semantic 429 (quota_exceeded + Retry-After)

[...]

Issue #920, opened 2026-08-19
feat(keys,ui): 展示 per-key vs per-account 配额语义（#915 拆分跟进）/ surface per-key vs per-account quota semantics (follow-up to #915)


suantea:
## 概述 / Overview

建议在 key 配置/管理界面**展示每平台的配额归属语义（per-key vs per-account）**，并允许为单个 key 标注其语义，避免用户误以为同一平台多加 key 必然叠加总配额。

This proposes surfacing each platform's **quota ownership semantics (per-key vs per-account)** in the key configuration/management UI, and allowing a per-key annotation, so users don't assume adding more keys to the same platform always stacks the total quota.

## 背景 / Background

从 #915 的讨论中拆分而来。维护者确认：部分平台额度归属**账号/项目**而非 key（如 Gemini 按 project 计费，多个 key 不叠加），而当前 UI 未做任何提示，用户容易按「多 key = 多额度」的直觉配置，实际并不叠加。

Split from #915. The maintainer confirmed some platforms bill per **account/project** rather than per key (e.g. Gemini per project — extra keys don't stack), and the current UI gives no hint, so users may configure multiple keys assuming stacked quota when it doesn't stack.

## 建议 / Proposal

- 在 key 管理界面（含新增/编辑 key 的对话框）展示该平台默认的配额语义（per-key / per-account / unknown），数据可来自平台配置表，unknown 时显示「未知，请咨询平台文档」；
- 允许对单个 key 覆盖标注（如 Gemini 标为 per-account），便于按实际计费方式理解自己的额度池；
- 仪表板/额度统计（#905）按此语义计算「池去重」后的可用额度，避免夸大。

- Show the platform's default quota semantics (per-key / per-account / unknown) in the key management UI (including the add/edit key dialog), sourced from the platform config table, with "unknown — check the provider docs" when unavailable;
- Allow overriding the annotation per key (e.g. mark Gemini keys per-account) so users can reason about their quota pool accurately;
- The dashboard/quota stats (#905) should compute pool-deduped availability per this semantics to avoid overstating usable quota.

## 边界 / Boundary

配额语义随平台/套餐可能变化，建议语义数据集中在平台配置中维护，UI 只做展示与覆盖标注；不改变服务端实际的额度计算逻辑（那是 #919 的前提）。

Semantics may vary by platform/plan; keep the semantic data centralized in platform config with the UI only displaying and annotating overrides. This does not change the server's actual quota accounting (which #919 depends on).

[...]

PR #1160, opened 2026-09-05
feat(quota): per-key vs per-account quota ownership semantics (#920)

suantea:
Implements #920: surfaces per-key vs per-account quota ownership semantics in the keys UI and respects them in pool-dedup statistics.

**Changes**
- `shared/types.ts`: add `quotaOwnership` field to Platform type ('per-key' | 'per-account' | 'unknown')
- `server/src/providers/index.ts`: annotate known platforms (Gemini→per-account, OpenRouter/free→per-key, etc.)
- `server/src/routes/keys.ts`: return `quotaOwnership` + `quotaOwnershipOverride` per key in GET /api/keys
- `server/src/routes/free-tier.ts`: pool-dedup respects ownership semantics (per-account pools counted once, not stacked)
- `client/src/i18n/*`: 5 new i18n keys (quotaOwnership, hint, perKey, perAccount, unknown) across all 58 locales

**Tests**: 14dc6ad0 + 994c92ac + 351c607a + 7c1572be (4 commits, all green locally)

**Relation to #1158/#1159**: orthogonal — #1159 adds per-key monthly budget caps (data layer), this adds per-key vs per-account semantic display (UI layer). Both needed for complete V0 billing model.


tashfeenahmed:
Thanks for picking this up, but I am going to close it rather than iterate.

The QUOTA_OWNERSHIP map is typed as Record<Platform, ...> and is missing radeon and sail, so tsc fails on main. The dedupe reduce in free-tier.ts has identical branches, so it changes nothing. No Keys page component is touched even though the description says the label is surfaced there, and quotaOwnershipOverride has no column behind it. The locale diff also moves the testProxy keys around in 60 files for no reason.

Bigger picture, pooled dedupe on the free-tier bar was reverted in #1141 on purpose, so re-adding it here is not the direction we want. See my note on #920.

PR #1187, opened 2026-09-10
Stamp request history with the concrete model row

nicyoong:
### Context

This is a direct follow-up to [PR #668 on the upstream repo](https://github.com/tashfeenahmed/freellmapi/pull/668)
("Proposal: Attribute request history to concrete model rows"), which was
closed as superseded. Per [the maintainer's closing comment](https://github.com/tashfeenahmed/freellmapi/pull/668#issuecomment-2320107000):

- the **analytics half** of #668 is already fixed on upstream main by
  [#958](https://github.com/tashfeenahmed/freellmapi/pull/958) (endpoint-scoped
  `/by-model` grouping), and the **routing half** by per-endpoint
  `modelStatsKey` keys — so this PR deliberately contains **no router and no
  analytics read-side changes**;
- one real gap remained: *"request identity is derived at read time from
  api_keys.base_url, so deleting a key or editing its base_url retroactively
  reattributes old history"* — and the maintainer explicitly invited
  *"a much smaller PR with just the migration and backfill (no FK, like you
  had it) plus logRequest stamping the column, and no router changes."*

This is that PR, rebased onto current main (the old 188-line version is
reduced to ~100 lines; the router and analytics hunks of #668 are dropped).
His note that main widened `logRequest` to `keyId: number | null` is
incorporated: custom requests whose key never reached routing stay `NULL`.

### Problem

Custom relays can expose the same upstream model id, so `(platform, model_id)`
no longer uniquely identifies the `models` row whose pricing, reliability, and
speed apply. Today, request identity is derived at *read time* from
`api_keys.base_url`: deleting a key, or editing its base URL, retroactively
reattributes (or orphans) all of that key's historical requests. The history
was never frozen to the row that actually served it.

### Changes

- **Migration** (`20260910_000001_request_model_attribution.ts`): adds the
  nullable column with a non-unique index and backfills existing rows.
  - Catalog requests backfill unambiguously by `(platform, model_id)`.
  - Custom requests backfill only while the saved key still resolves to the
    matching normalized endpoint scope (`rtrim(trim(base_url), '/')`, matching
    `endpoint-scope.ts`).
  - Orphaned custom history stays `NULL` rather than being attributed to the
    wrong relay.
  - Deliberately **no foreign key**, as in #668: request history must survive
    the user deleting the model row that once served it.
- **`logRequest`** resolves and persists the concrete `models.id` inside the
  same transaction as the insert:
  - catalog rows resolve by platform + model id;
  - custom rows resolve through the request's key (its normalized base URL is
    the row's `endpoint_scope`);
  - custom requests whose key never reached routing (`keyId: null`) stay
    `NULL` — an unattributable request must not land on the wrong relay.

### Out of scope

No router and no analytics read-side changes — this only freezes the record.
Read paths can adopt the column incrementally later, as discussed on #668.

### Validation

- New regression test (carried over from #668): two relays sharing one model
  id — a request logged against relay B's key stamps relay B's row id.
- `npm run build -w server`
- `npm run lint -w server`
- Focused suites, all green (105 tests):
  `roundtrip`, `custom-endpoint-identity`, `routed-via-header`,
  `analytics`, `proxy-model-groups`, `custom-provider`

### DIFF
--- a/server/src/__tests__/db/migrate/roundtrip.test.ts
+++ b/server/src/__tests__/db/migrate/roundtrip.test.ts
@@ -37,6 +37,7 @@ const QUOTA_OBSERVATION_LOOKUP_FILENAME = '20260901_000002_quota_observation_loo
 const ANALYTICS_LATENCY_PERCENTILE_INDEX_FILENAME = '20260902_000001_analytics_latency_percentile_index.ts';
 const MCP_ENABLED_DEFAULT_FILENAME = '20260903_000001_mcp_enabled_default.ts';
 const RESPONSE_CACHE_FILENAME = '20260903_000002_response_cache.ts';
+const REQUEST_MODEL_ATTRIBUTION_FILENAME = '20260910_000001_request_model_attribution.ts';
 
 interface SchemaRow {
   type: string;
@@ -120,6 +121,7 @@ describe('migration round trip', () => {
         ANALYTICS_LATENCY_PERCENTILE_INDEX_FILENAME,
         MCP_ENABLED_DEFAULT_FILENAME,
         RESPONSE_CACHE_FILENAME,
+        REQUEST_MODEL_ATTRIBUTION_FILENAME,
       ]);
     } finally {
       db.close();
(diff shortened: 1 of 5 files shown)


tashfeenahmed:
Thank you @nicyoong! This work landed via #1212 (squash abe4c08), which carries your regression test and credits you as co-author. Closing in favour of that.

PR #1203, opened 2026-09-11
fix(fallback): log the routing disposition on every surface, not just /v1/*/completions

chongjiazhen:
When `routeRequest` gives up before any upstream is tried, the per-candidate disposition it carries on `RouteError.diagnostics` is the only record of **why** the pool was empty. Without it a `routing_error` is indistinguishable from a genuinely dry pool — cooldowns, exhausted quota and context narrowing all read the same from outside.

Each surface was responsible for logging that itself, and only `routes/proxy.ts` ever did:

```
$ grep -rn "routeErr.diagnostics" server/src/routes server/src/lib
server/src/routes/proxy.ts:1411
server/src/routes/proxy.ts:2862
```

So `/v1/messages`, `/v1/responses` and the inbound-chat wires logged nothing at all, and an opaque exhaustion on those surfaces cannot be diagnosed from the server log. This is the same drift class the shared loop was created to remove (see its header comment), re-emerged one layer up in the hooks.

## What changed

Move the log into `runFallbackLoopAttempts`, which already holds `routeErr` and already knows whether any attempt ran (`lastError`). Surfaces now only name themselves, through a new optional `FallbackHooks.logIdentity`:

```ts
logIdentity?: { surface: string; requestId?: string; requestedModel?: string };
```

All five call sites pass one; the two duplicated blocks in `routes/proxy.ts` go away. A sixth surface gets the diagnostics for free, and the format can no longer drift per surface.

## Behavior

- Unchanged when attempts ran — the attempt trail in the exhaustion body already explains those, so the line stays suppressed exactly as before.
- Unchanged response bodies, statuses and headers. Log-only.
- One operator-visible change: the prefix moves from `[Proxy] routing exhausted (no upstream tried) …` to `[FallbackLoop] <surface> routing exhausted (no upstream tried) …`, matching the module's other log lines. Worth knowing if you grep for the old prefix.

## Tests

Four new cases in the existing `server/src/__tests__/lib/fallback-loop.test.ts`, driven through the file's `hooksSkeleton`:

- full identity + a multi-line disposition (surface, shortened request id, requested model, candidate count, every diagnostic line)
- a surface that supplies no `logIdentity` at all — the line still fires
- a `RouteError` carrying no `diagnostics` — `candidates=0`, no throw
- attempts already ran — asserts silence

Mutation-checked: guarding the call behind `if (false && !lastError)` (still compiles) fails 3 of the 4, and the silence case correctly still passes.

```
tsc --noEmit -p server/tsconfig.json   rc=0
vitest run --pool=forks --fileParallelism=false
  Test Files  259 passed (259)
       Tests  3026 passed | 8 skipped (3034)
eslint (6 changed files)                rc=0
```

Assisted by AI.

### DIFF
--- a/server/src/__tests__/lib/fallback-loop.test.ts
+++ b/server/src/__tests__/lib/fallback-loop.test.ts
@@ -758,3 +758,86 @@ describe('runFallbackLoop: client disconnect + attempt log', () => {
     expect(attemptLog[0].errorClass).toBe('rate_limited');
   });
 });
+
+// The diagnostics line used to live in each surface's onRoutingExhausted hook,
+// and only routes/proxy.ts ever implemented it — so an opaque routing_error on
+// /v1/messages, /v1/responses or an inbound wire recorded nothing about WHY the
+// pool was empty. It belongs to the loop: these tests pin it there, for every
+// surface, including one that supplies no identity at all.
+describe('runFallbackLoop: routing-exhaustion diagnostics belong to the loop', () => {
+  const routeError = (diagnostics: string[]) =>
+    Object.assign(new Error('all candidates exhausted'), { status: 429, diagnostics });
+
+  const warnedLine = (warn: ReturnType<typeof vi.spyOn>): string | undefined =>
+    warn.mock.calls.map(c => String(c[0])).find(l => l.includes('routing exhausted'));
+
+  it('logs the router per-candidate disposition when no upstream was tried', async () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    try {
+      await runFallbackLoop(hooksSkeleton({
+        logIdentity: { surface: 'anthropic messages', requestId: 'abcdef12-3456-7890', requestedModel: 'claude-x' },
+        route: () => { throw routeError(['glm-4.7 @zai: cooldown 42s', 'llama @groq: rpd 0 left']); },
+      }));
+      const line = warnedLine(warn);
+      expect(line).toBeDefined();
+      expect(line).toContain('anthropic messages');
+      expect(line).toContain('req=abcdef');           // hyphens stripped, 6 chars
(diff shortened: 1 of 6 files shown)


chongjiazhen:
Happy to stop here if you'd rather — but there's a second half to this, and I'd rather ask than send an unwanted refactor.

The reason the diagnostics could drift is that `onRoutingExhausted` and `onExhausted` are separate hooks, so there are 10 implementations of "render exhaustion on this wire" across the 5 `runFallbackLoop` call sites, and 4 of them are byte-identical:

```ts
setFallbackHeaders(res, info.attempts.length, info.attempts);
setExhaustionHeaders(res, exhaustion);
res.status(exhaustion.status).json({ error: exhaustionErrorPayload(exhaustion), execution_id: requestGroupId });
```

The two hooks only existed so a surface could log the zero-attempt disposition, which is exactly what this PR moved into the loop. That leaves them with no distinguishing work.

The follow-up would be:

- one `renderExhaustion(exhaustion, info, cause)` hook, where `cause` is `routing` / `attempts_spent` / `time_budget` / `breaker`. That's information surfaces don't have today — a breaker trip and a spent attempt cap currently arrive through the same hook with nothing to tell them apart.
- promote `inbound-chat.ts`'s existing `sendExhaustion(res, wire, exhaustion, attempts)` into a shared renderer the other four surfaces call with their own wire adapter. The seam already has two justified adapters (the Anthropic type mapping in `anthropic.ts`, the OpenAI JSON shape) plus the SSE `response.failed` branch in `responses.ts`.

Net ~10 hook bodies down to 5 one-liners over 1 renderer. No behavior change intended, and the renderer would stay in a sibling module so `fallback-loop.ts` keeps its "no Express, no wire-format knowledge" property.

Interested, or would you rather leave the hook pair alone?

PR #1204, opened 2026-09-11
fix(providers): send OpenCode Zen client headers so its free roster works

rodion-gudz:
## Problem

OpenCode Zen is listed as a supported provider, but every request through it is rejected by the Console gateway:

```
OpenCode Zen API error 400: Error from provider (Console):
OpenCode's free tier can only be used in OpenCode
```

and for the paid Go models:

```
400 ... MissingSessionID ... Request is missing x-opencode-session
```

The gateway expects the request to look like it came from the OpenCode client; the adapter sends none of those headers, so both the free roster and the paid Go plan are unusable even with a valid key. Verified on `main` (v0.9.8); no existing issue/PR covers this.

## Fix

- `OpenAICompatProvider` now accepts **function-valued `extraHeaders`**, evaluated once per request — needed because Zen expects a fresh `x-opencode-session`/`x-opencode-request` per call. Object-valued registrations are unchanged.
- The `opencode` registration supplies the client header set (`User-Agent`, `x-opencode-client`, `x-opencode-project`, fresh session/request ids).
- The User-Agent advertises the **newest published OpenCode release**, resolved from the npm registry (`opencode-ai/latest`, 12 h cache), so a future version check cannot age into a block. A failed lookup keeps the last known version; a built-in fallback covers a cold first boot.

## Testing

- New unit test: function-valued headers are re-evaluated on every request.
- `npm run build -w server`, eslint on the touched files, and the provider suite (23 files / 327 tests) are green.
- Live-verified against OpenCode Zen: with these headers both the free roster and the paid Go plan answer `200`; without them the same requests reproduce the errors above.

## Notes

- Lazy, background npm lookup; no new dependencies, fails silent.
- Scoped to the `opencode` registration and the header hook — no catalog or docs changes.

### DIFF
--- a/server/src/__tests__/providers/openai-compat.test.ts
+++ b/server/src/__tests__/providers/openai-compat.test.ts
@@ -70,6 +70,36 @@ describe('OpenAICompatProvider', () => {
     expect(capturedBody.messages[0].role).toBe('user');
   });
 
+  it('evaluates function-valued extraHeaders once per request', async () => {
+    let issued = 0;
+    const dynamic = new OpenAICompatProvider({
+      platform: 'opencode',
+      name: 'Zen',
+      baseUrl: 'https://api.test.com/v1',
+      extraHeaders: () => ({ 'x-opencode-session': `session-${++issued}` }),
+    });
+    const seen: string[] = [];
+    vi.spyOn(global, 'fetch').mockImplementation(async (_url, init) => {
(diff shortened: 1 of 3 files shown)


rodion-gudz:
image
before

rodion-gudz:
image
after

daikkin:
up

Ahmedtahoon2:
+1

Issue #1184, opened 2026-09-10
Fail over on reasoning-only responses (thinking trace with no content and no tool calls)


suantea:
## Problem

`server/src/lib/inbound-chat.ts` treats a response as empty only when it has **no text AND no reasoning AND no tool calls**:

- Non-streaming (line ~261): `if (!text && !reasoning && toolCalls.length === 0)` — a response carrying only a `reasoning_content` trace passes as success and is delivered as-is.
- Streaming (line ~399): the **first reasoning delta immediately commits** the stream (`commit()` → headers + `startStream` + forwarded deltas). From that point failover is impossible, so a stream that ends with a thinking trace but zero content and zero tool calls is delivered as the final answer.
- The stream-end check (line ~474) has the same three-way condition, so reasoning counts as "content" there too.

Downstream this matters more than it looks: the primary consumers of this proxy are coding agents (Cursor / Cline / Codex CLI). To them a reasoning-only turn is an **empty assistant message** — the agent loop stalls or the user sees a blank reply. Reasoning models behind free relays produce exactly this signature fairly often (hidden reasoning spends the whole budget, relay truncates the answer, or the relay drops `content` entirely while forwarding `reasoning`).

Note the asymmetry: a fully-empty completion already fails over correctly (`empty completion` throw + the #751 streak-bounded exemption). It is specifically the **reasoning-only middle state** that has no handling — it is counted as a success, recorded via `recordUpstreamSuccess`, and never retried.

## Proposed improvements (small, independent PRs)

1. **Non-streaming: fail over on reasoning-only responses** — treat `text === '' && toolCalls.length === 0` (reasoning may be present) as the same retryable `empty completion` failure class, so the fallback loop picks the next candidate and the client never sees a think-only answer. Should flow through the same #751-style streak-bounded exemption so a reasoning model that legitimately produces a long trace is not benched after 3 strikes.

2. **Streaming: defer commit while only reasoning has arrived** — buffer reasoning deltas (bounded, e.g. N chars / T seconds, mirroring the dialect-marker hold already used for `heldText`) and only `commit()` on the first content or tool-call delta. If the stream ends while still reasoning-only and uncommitted, throw the same retryable error and fail over pre-commit. Once committed (first real content sent), current behavior stands — never tear down a stream the client is reading.

[...]

PR #1205, opened 2026-09-11
fix(routing): fail over on reasoning-only non-stream responses on the inbound surface (#1184)

suantea:
## Problem

The Ollama/Gemini surface (shared `runInboundChat` in `inbound-chat.ts`) has a non-streaming empty-completion check that counts a reasoning-only response — a thinking trace with no text and no tool calls — as a success:

```ts
// BEFORE (line 261)
if (!text && !reasoning && toolCalls.length === 0) { throw ... }
```

A reasoning-only turn is an empty assistant message to the agent clients that consume these surfaces; it should fail over to the next candidate, not be delivered as a success.

The main OpenAI surface (`proxy.ts`, `anthropic.ts`, `responses.ts`) already ignores `reasoning` in the same check (`!respText && toolCalls.length === 0`), so this PR aligns the inbound surface with the established behavior.

**Why only Google is affected in practice:** OpenAI-compat providers (`normalizeChoices` in `openai-compat.ts`) intentionally fold `reasoning_content` into `content` when content is empty and there are no tool calls (Z.ai / Ollama Cloud compatibility). After the fold, the reasoning-only shape never reaches the route layer. Google's Gemini provider, however, synthesizes `reasoning_content` from `thought: true` parts while leaving `content` empty — this is the reachable path for the fix.

## Changes

- **`server/src/lib/inbound-chat.ts`** — drop `!reasoning` from the non-streaming empty-completion condition, add a comment explaining the reasoning-never-counts-as-output semantics
- **`server/src/__tests__/routes/inbound-chat-reasoning-only.test.ts`** — new test file (3 tests on the google surface):
  1. Reasoning-only non-stream completion fails over to the next model
  2. Reasoning + text completion is delivered in one hop (no false failover)
  3. Exhaustion (all candidates reasoning-only) returns an error, never delivers the trace

## Not in scope

- **Streaming path** (`inbound-chat.ts` line 410-413): the first reasoning delta still commits the stream immediately, preventing pre-commit failover. This is issue #1184 proposal 2 and will be a separate PR.
- **Provider-level fold** (`openai-compat.ts normalizeChoices`): the intentional Z.ai / Ollama Cloud compatibility fold is left untouched; it is a separate design tradeoff tracked in issue #1184.
- **Bookkeeping parity** (issue #1184 proposal 3): `X-Fallback-Trail` class separation and streak counter adjustments are deferred to a follow-up.

### DIFF
--- a/server/src/__tests__/routes/inbound-chat-reasoning-only.test.ts
+++ b/server/src/__tests__/routes/inbound-chat-reasoning-only.test.ts
@@ -1,0 +1,152 @@
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+import type { Express } from 'express';
+import { createApp } from '../../app.js';
+import { getDb, initDb, setSetting } from '../../db/index.js';
+import { encrypt } from '../../lib/crypto.js';
+import { setRoutingStrategy } from '../../services/router.js';
+
+// #1184: on the Ollama/Gemini surfaces (shared runInboundChat), a
+// reasoning-only completion — a thinking trace with no text and no tool
+// calls — must fail over exactly like a fully empty completion.
+//
+// The fix is only reachable on providers that keep reasoning_content
+// separate from content. OpenAI-compat providers (groq, cerebras, etc.)
+// normalizeChoices folds reasoning_content into content before it reaches
+// the route layer (intentional Z.ai / Ollama Cloud compatibility), so a
+// reasoning-only response arrives as content=thought there — that path is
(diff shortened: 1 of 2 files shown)


suantea:
Closing — upstream landed the same fix via #1191 (commit fe0f1c4e, merged 2026-09-13), which resolves #1184 on the inbound surface. This PR is now redundant. The streaming-path failover and bookkeeping-parity follow-ups mentioned here remain open as scope for #1184 itself. Thanks!

PR #1216, opened 2026-09-14
fix(routing): fail over on reasoning-only non-stream responses on the inbound surface

### LINKED ISSUE #1184: Fail over on reasoning-only responses (thinking trace with no content and no tool calls)

suantea:
## Problem

On the inbound Ollama/Gemini chat surface (`inbound-chat.ts`), the non-streaming empty-response guard counted `reasoning_content` as "has content". A response that contained only a thinking trajectory — no text, no tool calls — was treated as a successful delivery, so agent clients received an empty reply with no failover.

## Fix

Change the empty-response condition from `!text && !reasoning && toolCalls.length === 0` to `!text && toolCalls.length === 0`, aligning this surface with the main OpenAI surfaces (`proxy.ts` / `anthropic.ts` / `responses.ts`), where reasoning is already not counted as content.

### Why only this surface is affected

- OpenAI-compat platforms (groq, cerebras, …) fold `reasoning_content` into `content` in `normalizeChoices` at the provider layer (matching Z.ai / Ollama Cloud compat behavior), so the gap is unreachable there.
- The Google platform (Gemini) synthesizes `reasoning_content` from thought parts while leaving `content` empty — the only path where a reasoning-only non-stream response can pass the old guard.

## Testing

- Full suite green: server 3028 passed / 5 skipped, cli 107 passed, client 261 passed; `eslint` clean.
- Node 22.x per `.nvmrc` / `engines`.

## Scope note

The streaming-side issue from the same report (reasoning deltas committing an empty turn immediately) is intentionally left out and will follow as a separate PR.

Closes #1184

### DIFF
--- a/server/src/__tests__/routes/inbound-chat-reasoning-only.test.ts
+++ b/server/src/__tests__/routes/inbound-chat-reasoning-only.test.ts
@@ -1,0 +1,152 @@
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+import type { Express } from 'express';
+import { createApp } from '../../app.js';
+import { getDb, initDb, setSetting } from '../../db/index.js';
+import { encrypt } from '../../lib/crypto.js';
+import { setRoutingStrategy } from '../../services/router.js';
+
+// #1184: on the Ollama/Gemini surfaces (shared runInboundChat), a
+// reasoning-only completion — a thinking trace with no text and no tool
+// calls — must fail over exactly like a fully empty completion.
+//
+// The fix is only reachable on providers that keep reasoning_content
+// separate from content. OpenAI-compat providers (groq, cerebras, etc.)
+// normalizeChoices folds reasoning_content into content before it reaches
+// the route layer (intentional Z.ai / Ollama Cloud compatibility), so a
+// reasoning-only response arrives as content=thought there — that path is
+// handled elsewhere. Google's Gemini provider synthesizes reasoning_content
+// from thought parts while leaving content empty, which IS the reachable
+// path for this fix.
+
+// ── Gemini response shapes ──────────────────────────────────────────────────
+function geminiResponse(parts: Array<{ text: string; thought?: boolean }>, finishReason = 'STOP') {
+  return {
+    candidates: [{ content: { role: 'model', parts }, finishReason }],
+    usageMetadata: { promptTokenCount: 10, candidatesTokenCount: 20, totalTokenCount: 30 },
+  };
+}
+
+const REASONING_ONLY = geminiResponse([{ text: 'thinking... a long trace with no visible answer', thought: true }]);
+const REASONING_PLUS_TEXT = geminiResponse([
+  { text: 'thinking first', thought: true },
+  { text: 'the answer' },
+]);
+const GOOD = geminiResponse([{ text: 'a real answer' }]);
+
+function geminiJson(body: unknown): Response {
+  return new Response(JSON.stringify(body), {
+    status: 200,
+    headers: { 'Content-Type': 'application/json' },
+  });
+}
+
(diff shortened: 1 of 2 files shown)


suantea:
Closing — upstream #1191 landed a more complete fix for the same issue (both the non-streaming guard and the streaming reasoning-delta hold), so this PR is fully superseded. Thanks!

PR #111, opened 2026-05-28
feat: Named model fallback profiles, dynamic multi-view layouts, real-time limits display, and backend security hardening

barbotkonv:
## Overview
This Pull Request introduces significant enhancements to `freellmapi` by implementing **Named Model Fallback Profiles** (enabling flexible, custom-tailored retry priorities), standardizing **Backend Security & Host Protections** (constant-time API key auth, CORS origin matching, IP spoofing/dns rebinding guards, parameterization), and building a versatile **Multi-View Fallback Management UI** (List, Grid, Kanban Columns, and Tier Groups views with highly-optimized cross-block drag-and-drop). All modifications are designed to be 100% backward-compatible.

---

### 1. Named Model Fallback Profiles (Backend & Database)
* **Database Schema Updates:** Added `profiles` and junction `profile_models` tables to sqlite database config. Pinned a static system **Default** profile representing the original global prioritizations, fully immunized against editing/deletion.
* **REST API CRUD:** Built clean REST API endpoints (`/api/profiles/*`) for profile creation, clone-on-creation, editing, deleting, reordering, custom sorting (by intelligence, speed, budget presets), and baseline reset.
* **Integrations:** Connected active profiles dynamically with request routing (`routeRequest()`) and token limit summaries (`/api/fallback/token-usage`).

### 2. Security Hardening & Robustness
* **CORS & Host Protection:** Locked API endpoints to run strictly over loopback/local origins, protecting host browsers from cross-origin budget/token abuse.
* **Timing-Safe API Key Authentication:** Integrated an `/api/*` endpoint request middleware comparing key lookups with timing-safe `crypto.timingSafeEqual()` to guard against side-channel analysis.
* **SQL Parameterization:** Refactored analytical queries (`routes/analytics.ts`) to avoid string-concatenated SQL queries.
* **IP-Spoofing Protections:** Hardened header loopback resolution in proxy routines (`services/proxy.ts`).
* **Instant 404 Model Cooldown Skip:** Updated proxy routing loops to discard defunct or missing models instantly when they return a `404 Not Found` response instead of blocking routing lists on infinite retry cooldoowns.

### 3. Advanced Fallback UI & Dynamic View Modes
* **Collapsible Profiles Quick-Bar:** Displays profile nodes in a responsive bar. Hidden custom profiles collapse seamlessly behind a dynamic `+N` badge (calculated on-the-fly using `ResizeObserver`). Clicking the active custom profile quickly toggles back to the default profile configuration.
[...]

PR #1222, opened 2026-09-14
fix(fallback): short-circuit a platform after distinct model_not_found misses

suantea:
## Problem

A stale catalog misses on every sibling model in a row: NavyAI served 24 models to this relay over one session, **0 ok, 65 errored**, each miss re-paying a 2.5–10s round trip (#1218). The per-model skip (`skipModels`) can't see the pattern — each miss is a *different* model — so the ladder ground through the whole catalog within a single request, burning most of the failover time budget on models that were never going to answer.

## Fix

Tally **distinct** `model_not_found` hops per platform in the request-scoped `FallbackState` (`modelNotFoundPlatforms: Map<string, Set<number>>`). From `MODEL_NOT_FOUND_PLATFORM_LIMIT` (= 3) distinct misses the platform joins `skipPlatforms` for the rest of the request: the catalog (or the key's access to it) is broken, and the next *provider* is the better hop.

Design notes:

- **Request-scoped** like `skipModels` (#111/#256 semantics) — the tally self-heals on the next request, so a catalog that comes back costs one request of miss-hops, not a ban.
- **Sibling keys tally together on purpose**: they share the same catalog, so `platform:model:key` identity would just delay the inevitable.
- **Limit = 3**: a healthy catalog never misses three different models in one failover chain; one fluke removal won't condemn a provider.
- Existing model-level skips (`isModelNotFoundError` / `isModelAccessForbiddenError` / `isContextTooLargeError`) are unchanged — this adds only the cross-model platform rule-out on top.

## Test

3 new cases in `fallback-loop-account-suspended.test.ts`: platform ruled out from the Nth *distinct* miss; repeated misses on the *same* model don't tally; sibling keys tally together. Full lib suite 867 passed; lint + `tsc --noEmit` clean.

## Scope note

Covers #1218 Gap 1's *amplification* (the per-request short-circuit). The quoted-id regex widening that makes NavyAI's exact wording (`The model 'gpt-4.1' does not exist...`) classify as model-not-found at all is a separate PR (#1219) — together the two close the gap; the tests here deliberately use plain phrasing so this PR stands alone.

Refs #1218


tashfeenahmed:
Thanks, this is a good fix. I pushed one small commit on top: custom relays are now exempt from the platform skip. Every custom relay shares the platform id `custom`, so three misses spread over three different relays would have skipped every healthy relay for the rest of the request. Their misses stay per-model. Added a test for that case, full server suite passes. Note the classifier half is already on main through #1242, so the #1219 line in the description is out of date.

PR #1224, opened 2026-09-15
fix(routing): honor auto:<profile> chains on /v1/messages and /v1/responses

### LINKED ISSUE #1170: /v1/messages ignores auto:<profile> fallback chains — silently routes over the active pool instead (cross-leg routing parity)
## Summary

`auto:<name>` (named fallback chain) is only honored on `/v1/chat/completions`. The `/v1/messages` (Anthropic) leg silently ignores it — and so does the `auto:*` branch of `/v1/responses` — degrading to full-pool auto-routing over the **active** profile. A request that names a one-model chain can be served by any enabled model in the catalog, with no error and nothing in the response body indicating substitution (only `X-Routed-Via` reveals it).

## Version

`v0.9.7` (ghcr.io/tashfeenahmed/freellmapi)

## Repro

1. Create a routing profile named `glm-5-3` with exactly **one** enabled model (say provider A's `zai-org/GLM-5.3`); leave the active/default profile with many enabled models.
2. `POST /v1/chat/completions` with `model: "auto:glm-5-3"` → served by the profile's member. If that member is unavailable, the request fails loudly (strict chain) — correct.
3. `POST /v1/messages` with the same `model: "auto:glm-5-3"` → the request is served by an **unrelated model from the active pool**. Observed repeatedly in production: the chain names one model, the per-attempt log shows a different platform/model entirely (e.g. a DeepSeek distill on another provider), while the response still echoes the requested model id.
4. Same for `POST /v1/responses` with `auto:<name>` — treated as plain `auto`.

## Root cause

Each dialect leg carries its own model-resolution code path, and the chain vocabulary was only ever wired into the oldest one (verified in the v0.9.7 dist):

- `routes/proxy.js` (chat): `isAutoModel(model)` → `resolveRoutingChain(model)` → named chain, strict. Unknown pinned ids → 400 ("silently auto-routing to a different model would be surprising" — the code's own comment). `resolveRoutingChain` appears 3× here.
- `routes/anthropic.js`: resolves via `services/anthropic-map.ts#resolveAnthropicModel`, which understands only Claude family aliases (opus/sonnet/haiku → the operator map) and concrete catalog ids. Anything else — including `auto:<name>` — returns `{ pinned: false }` and the request auto-routes over the entire active profile chain. `resolveRoutingChain` appears **0×** here.
- `routes/responses.js`: the `isAutoModel` branch applies sticky-session preference only — no `resolveRoutingChain` call. **0×** here.

## Why it matters

- Silent substitution contradicts the strict-chain / no-silent-substitution principle the chat and responses legs already enforce for pinned ids.
- The response body echoes the requested model id, so callers believe the named model served. Downstream this degrades unpredictably: a behaviorally weak substituted model answers with intent text and `end_turn` instead of emitting tool calls, and the caller can't tell why.
- `requested_model` is logged as NULL for these requests, so analytics cannot distinguish "caller asked for auto" from "caller named a chain that was ignored".

## Suggested fix

Factor model resolution into one shared path used by all three legs: `auto:<name>` resolves the named chain strictly, and an unknown `auto:<name>` returns 400 like an unknown pinned id instead of degrading to pool-wide auto-routing.

Happy to provide more log detail if useful.

rome-xi:
## Summary

- Wire `resolveRoutingChain('auto:<profile>')` into Anthropic `/v1/messages` and OpenAI `/v1/responses`, matching `/v1/chat/completions` named-chain semantics (#1170).
- Unknown `auto:<name>` now returns **400** on both surfaces (no silent active-pool substitution).
- Sticky preference for named chains uses `strategyKey`; plain `auto` stays unscoped.

## Testing

Honest TDD on pristine `main` (`a733a3b`):

- **RED**: four new tests failed — `/v1/messages` and `/v1/responses` served active-pool models for `auto:<profile>` and returned 200 for `auto:no-such-group`.
- **GREEN**: same four tests pass after the fix; related suites 71 passed.

```bash
cd server && npx vitest run --pool=forks --fileParallelism=false \
  src/__tests__/routes/models-management.test.ts -t "#1170"
```

## Notes

- Small routing parity fix; no drive-by refactors.
- CONTRIBUTING: LLM-assisted PRs welcome; no special disclosure required.

Fixes #1170


tashfeenahmed:
Reviewed and tested, thanks for the clean fix and the honest TDD notes. Squash merging now, it will be in v0.10.1.

PR #1252, opened 2026-09-17
fix(sail): use flex completion_window when tools are present

### LINKED ISSUE #1248: Sail: every request with tools fails with "not supported with completion_window=asap"
**What happens**
Every request routed to Sail from a coding agent fails with HTTP 400. Nothing gets billed (my $5 monthly free credit is untouched, and a payment method is attached).

**Error message**
Sail Research API error 400: This request feature is not supported with completion_window=asap. Use a supported non-ASAP completion window.

**Models affected (all on the sail provider)**
- moonshotai/Kimi-K3
- openai/gpt-oss-120b
- zai-org/GLM-5.2-FP8
- zai-org/GLM-5.3
- moonshotai/Kimi-K2.6
- deepseek-ai/DeepSeek-V4-Flash

**Setup**
- FreeLLMAPI desktop app, Premium (live catalogue)
- Client: OpenCode 1.18.31 with oh-my-opencode-slim 2.2.20, routed through a named chain (auto:coding)
- Requests include OpenAI-style tool definitions
- Failing since at least 15 Sept 2026, still failing 16 Sept 2026

**Likely cause**
In the Sail provider, completionWindow() returns 'asap' for any model not in FLEX_ONLY_MODELS, and tools are passed through on the same request. It looks like Sail rejects tools (or another feature in agent requests) when completion_window is 'asap'.

**Possible fixes**
- Use 'flex' when the request includes tools, or
- Mark Sail models as not supporting tools in the catalogue so the router skips them for tool requests.

For now I've turned off my Sail key.

rome-xi:
## Summary

- Sail Research returns HTTP 400 when a request includes tools together with `metadata.completion_window=asap`.
- Use `flex` whenever `options.tools?.length` is truthy (flex-only models already used flex).
- Tool-free requests on non–flex-only models still use `asap`.

## Testing

Honest TDD on pristine `main` (`ac6d29d`):

- **RED**: new unit test failed — `moonshotai/Kimi-K2.6` + tools still sent `completion_window=asap` (`expected 'asap' to be 'flex'`).
- **GREEN**: sail provider suite 6/6 pass after extending `completionWindow(modelId, options)`.

```bash
cd server && npm test -- src/__tests__/providers/sail.test.ts
```

## Notes

- Minimal provider change; no catalogue / no-tools marking; no Sail polling redesign.

Fixes #1248


tashfeenahmed:
Thanks for the quick fix and the clean tests. Closing in favour of #1259: the reporter reproduced the same 400 in #1248 with a plain text request and no tools, so the asap window has to go for every Sail request, not only tool calls. Your Linux sandbox fix in #1240 is going in.

PR #1272, opened 2026-09-20
fix: make fallback retry budgets endpoint TTFB-aware

### LINKED ISSUE #1262: Reliability: per-endpoint TTFB-aware retry budget (kill slow-but-alive free endpoints prematurely)
## Summary

The default 45-second retry budget kills **slow-but-alive** endpoints on every request, wasting the full failover budget before giving them a chance to respond.

Evidence from a single production install (7-day window, 9/15):

| platform | successful requests | avg TTFB | max TTFB |
|---|---|---|---|
| ollama | 28 | **56,954 ms** | 95,485 ms |
| nvidia | 6 | **56,654 ms** | 194,586 ms |
| requesty | 6 | **48,699 ms** | 68,472 ms |
| kilo | 8 | 7,431 ms | 17,804 ms |
| custom relay | 72 | 4,299 ms | 79,572 ms |

Meanwhile 12 requests were exhausted with `timeout` and the message *"retry time budget 45000ms expired with no first byte"*. A subsequent request to the same model on the same endpoint succeeded at 47s — proving the endpoint was never dead, only slow.

The operator already found a workaround: setting `fallback_time_budget_ms=120000`. But this is a global sledgehammer — it makes every endpoint wait 120s even when most are fast.

## Root cause

`getFallbackTimeBudgetMs()` returns a single global budget for the entire failover ladder. The loop checks this budget **before starting each retry**, and aborts in-flight attempts when the budget expires. This works for fast providers (OpenAI, Anthropic, most commercial relays) where a stalled attempt is truly dead. It fails for free-tier endpoints that are alive but throttled — they spend 50-90s warming up, then succeed. The router kills them at 45s, wastes the budget, and the next request repeats the same failure.

## Proposed fix: per-endpoint TTFB-aware budget

Each `(platform, endpoint_scope)` bucket tracks its own observed TTFB distribution (P50, P95 over a sliding 7-day window with 2-day half-life decay, same as the existing reliability bandit). The loop uses this to **dynamically widen the budget** for slow endpoints:

```
base = getFallbackTimeBudgetMs()  // 45000ms default
slow = endpointTtfbP95(platform, endpointScope)
if slow > 0:
    effective = max(base, slow + buffer_ms)   // buffer = 10s by default
else:
    effective = base
```

`buffer_ms` is tunable (`SLOW_ENDPOINT_BUFFER_MS`, default 10000). The existing `abortInFlight()` hedge logic applies unchanged — it just gets a longer timer for slow endpoints.

### What stays the same

- Fast endpoints keep their 45s budget — no behavioral change for them
- Budget checking still happens before each retry and mid-attempt via `abortInFlight()`
- The same environment variable / settings table override is respected as the floor (`effective = max(base, calculated)`)
- Exhaustion messages still name the budget (now per-endpoint-aware)

### Configuration

| Setting | Default | Purpose |
|---|---|---|
| `fallback_time_budget_ms` | 45000 | Base budget for "normal" endpoints |
| `slow_endpoint_buffer_ms` | 10000 | Extra headroom added to slow endpoints |
| `ttfb_budget_window_ms` | 604800000 (7d) | Sliding window for TTFB stats |
| `ttfb_budget_half_life_ms` | 172800000 (2d) | Exponential decay half-life |

Kill switch: `TTFB_BUDGET_DISABLED=1` (same convention as `ENDPOINT_HEALTH_DISABLED`).

## PR plan

Single PR: `server/src/lib/fallback-loop.ts` + `server/src/lib/ttfb-budget.ts` (new module) + tests + env docs.

## Why this matters for free-tier aggregation

Free endpoints are inherently slower — they run on consumer hardware, share resources with other users, and throttle under load. A one-size-fits-all budget treats "slow" and "dead" identically. The per-endpoint TTFB adaptation lets free-tier endpoints compete on equal footing with commercial providers without forcing operators to set a global 120s budget (which would make fast endpoints unnecessarily slow on exhaustion).

*中文备注：此问题 #1218 Gap 2。生产数据来自本机实例（127.0.0.1:31415），错误率约30%，失败请求平均挂7.3s。*

Inference1:
Fixes #1262.

Later failover attempts to healthy but slow endpoints can be canceled by the fixed 45-second budget before receiving a first byte. Derive an endpoint-specific allowance from successful request TTFB history: `max(base budget, weighted P95 + buffer)`. With a 50-second P95 and the default 10-second buffer, the endpoint gets 60 seconds; fast endpoints and endpoints without usable history retain the base budget.

The change is limited to the shared fallback loop, a cached statistics helper, regression tests, and configuration documentation. It uses the existing request log with a seven-day window and two-day half-life, isolates custom relay scopes, and adds settings/environment overrides plus `TTFB_BUDGET_DISABLED`. The same allowance drives retry checks, hedging, timeout messages, and the existing health threshold. Rejected candidate routes release their leases; first-attempt/first-retry exemptions, stream disarming, and client cancellation remain intact.

Validation:

- Server: 3348 tests passed, 8 skipped, including 26 new regression tests.
- Client: 327 tests passed; i18n validation passed for 60 locales.
- Bootstrap and contribution-hook tests passed on Windows.
- `npm run lint`: passed for server and CLI.
- `npm run build`: passed for server, CLI, and client.
- CLI: 87 passed, 31 failed on Windows path-separator assertions/snapshots. CLI files are unchanged; the overall test command is not reported as fully green.

Statistics reuse the existing logged TTFB semantics, which may include earlier retries, and refresh at most once per minute. No dependency or database schema changes.


Inference1:
@tashfeenahmed Would you please review this pr

tashfeenahmed:
Thanks for this. I tested it end to end before merging: real server, mock upstreams, a 6s base budget, two endpoints that fail after 3s each, then an endpoint that answers in 5s with 5s of history.

- main: 0/6 requests served, all cut at 6s, the slow endpoint was never contacted
- this PR: 6/6 served by the slow endpoint at ~11s

So the core idea works. I pushed one commit with three bounds:

1. **Min samples** (`TTFB_BUDGET_MIN_SAMPLES`, default 5): a single slow success no longer widens an endpoint's budget.
2. **Cap at 3x the base budget**: `ttfb_ms` is measured from request start, so it includes the hops that failed before the endpoint was reached. A widened budget admits later successes, which log a larger TTFB, which widens the budget again. The cap stops that ratchet.
3. **Slow-only**: widen only when P95 is at least half the base. The 10s buffer is absolute, so with a 6s base a 200ms endpoint that hung was held for 10.2s instead of 6s. With the 45s default this changes nothing.

Known limitation, not addressed here: non-stream successes log `ttfb_ms` as NULL, so only streamed traffic builds history.

